PR #11, opened 2025-02-25
keyboard glitch fix and Export logs feature

trinadhthatakula:
What's new: 
* Neko Logger (test feature)
* Export logs - now users can export logs for individual apps

Bug Fixes:
* One of the issues with search bar and keyboard glitch is now fixed.

APK Size was around 1.8MB


trinadhthatakula:
Awaiting testers confirmation

PR #241, opened 2026-07-06
feat(freezer): choose Freeze or Suspend mode in the Freezer (#239)

trinadhthatakula:
Closes #239.

## What & why
Adds a global **Freezer action mode** so the Freezer — including screen-off **auto-freeze** — can **suspend** apps (`pm suspend`, app stays visible but paused) instead of **freezing/disabling** them (`pm disable`). Default stays **Freeze**, so existing users see no change on update.

Requested in #239 ("Suspend by default"): *"choose the freezing mode for the Freezer … suspend the application instead of freezing it … when automatic freezing is enabled, the apps would be suspended."*

## Approach
`setAppSuspended` was already implemented end-to-end (all three gateways, use-case, `MainViewModel` actions, `AppInfo.isSuspended`) but never wired into the Freezer. This PR adds one `FreezerMode` preference and routes every freeze path through it:

- **Data:** `FreezerMode` enum + `freezerMode` DataStore preference (default `FREEZE`, string-key with safe `valueOf` fallback).
- **Freeze paths honor the mode:** single-app freeze, batch "Freeze all" (`performCountedFreeze` carries `useSuspend`, keeping the compact count UI **and** the unsafe/UAD skip), the multi-select toolbar, and `AutoFreezeManager`.
- **Core generalization:** the Freezer previously equated *frozen* with *disabled*; now **frozen = disabled OR suspended** (`isFrozen`/`isActive` predicates), and **unfreeze is state-aware** (`restoreApp` unsuspends and/or enables as needed).
- **UI (both, one shared preference so they stay in sync):** a `ConnectedButtonGroup` (Freeze | Suspend) in the Freezer settings sheet, and a "Suspend instead of freeze" switch in Settings.

## Robustness — adversarial review
An automated multi-agent review of the diff found **5 real bugs** around the mixed *disabled+suspended* state (single-op restore reporting false success; multi-select toolbar using the raw `enabled` flag; non-state-aware remove paths; auto-freeze stacking disable onto a suspended app). **All 5 are fixed and re-verified** in this branch:
- restore now clears **both** dimensions;
- freeze only ever acts on **active** apps (never creates mixed state);
- the multi-select `FreezerSelectToolBox` uses `isFrozen`/`isActive` + `useSuspend`.

## Testing
- `FreezerModeTest` — 8 pure unit tests (predicates + restore plan incl. mixed state). ✅
- `./gradlew :app:testFossDebugUnitTest assembleFossDebug` — green. ✅
- ⚠️ **Not yet exercised on a device** — on-device pass per backend (Root/Shizuku/Dhizuku) still recommended.

## Known limitations (documented fast-follow)
- Launcher-shortcut **bulk** freeze and the **AppList** multi-freeze remain disable-only (suspend-mode unaware). Pre-existing surfaces, out of scope here.

## Docs
Design spec + implementation plan included under `docs/superpowers/` — happy to drop them from the PR if you'd rather not track them.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -150,25 +150,34 @@ class RootSystemGateway(
         if (!packageName.matches(PACKAGE_NAME_REGEX)) {
             return Result.failure(IllegalArgumentException("Invalid package name: $packageName"))
         }
+        val hasReflection = android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q
+
+        fun isCurrentlySuspended() = getApplicationInfoCompat(packageName)?.run {
+            (flags and android.content.pm.ApplicationInfo.FLAG_SUSPENDED) != 0
+        } ?: false
+
+        // SUSPEND: go through the reflection path (RootMain) so the *suspending package* is our own
+        // app id — setPackagesSuspendedAsUser(caller = BuildConfig.APPLICATION_ID). A bare
+        // `pm suspend` run in a root shell (uid 0) instead records the suspender as "root", a
+        // package that does not exist, so tapping the paused app crashes SuspendedAppActivity with
+        // "IllegalArgumentException: Package root does not exist". We deliberately do NOT fall back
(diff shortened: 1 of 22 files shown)


trinadhthatakula:
@coderabbitai review

Issue #46, opened 2025-06-26
Is root required for app freezing? 


videoMonkey:
I see the section on freezing in the interface, but I don't see how to freeze an app.

trinadhthatakula:
Hi @videoMonkey latest release now supports Shizuku, you can use it instead of root now, https://github.com/trinadhthatakula/Thor/releases/tag/Thor_v1656

I am closing this issue now

PR #256, opened 2026-07-19
fix(audit): remediate P0/P1 findings from dev-branch codebase audit

trinadhthatakula:
## Summary

Remediates the high/medium-priority findings from the **2026-07-20 `dev`-branch codebase audit**. Full report included in this PR: `docs/audits/dev-branch-audit-2026-07-20.md`.

- **20 files** changed · `./gradlew assembleFossDebug` ✅
- Every fix was **adversarially diff-reviewed**; that pass found 3 regressions which are also fixed here.

### P0 (crashes / deadlock)
- **H1 `AppList.kt`** — `rememberSaveable<List<AppInfo>>` → `NotSerializableException` on rotation in multi-select; switched to plain `remember` (+ removed a pointless `derivedStateOf`).
- **H2 `RootSystemGateway.kt`** — the bind handshake ran inside `connectionMutex` with no timeout; a null/absent binder callback could pin the mutex and deadlock every privileged op. Wrapped in `withTimeoutOrNull` + `onNullBinding`.
- **H3 `ExtensionManagerViewModel.kt` / `ExtensionManagerScreen.kt`** — unguarded `pm.getInstalledApplications()` crashed the Extensions screen on open; now `runCatching` + error state + retry UI.

### Memory / resource leaks
- **L1 `InstallerEventBus` / `InstallerViewModel`** — the app-scoped `@Single` `replay=1` bus pinned a decoded `Bitmap` for the process lifetime and replayed terminal state onto other screens. Now `DROP_OLDEST` + `reset()`, cleared in `onCleared()` **guarded by `ownsInstall`** so a non-owning installer VM can't wipe a state another live VM is showing.
- **L2 `InstallerRepositoryImpl`** — per-progress-tick detached `CoroutineScope` (~100+/install) replaced with a conflated `Channel` drained by the install job (cancellation-aware, no post-terminal emissions).
- **L3 `InstallerRepositoryImpl`** — `PackageInstaller` session now abandoned on `openSession` failure.
- Plus installer-UI fixes: `collectAsStateWithLifecycle`, observable `DisposableEffect` key, reset-on-entry, and gating the Extension store refresh on a real store install.

### Robustness
- **#7 `MainShell`** — `isInitMain` reset in `finally` (a transient build failure no longer bricks shell creation for the process); async failure no longer swallowed.
- **#9/#10 providers** — `ExtensionOpsProvider` / `FreezerBridgeProvider` bound their `runBlocking` with `withTimeout` and fail closed, so a hung shell can't pin a binder-pool thread.
- **#25/#26 `FreezerTileService`** — honors `FreezerMode` (suspend vs disable) and moves bulk work off the tile-service scope.
- **#27 `ThorApplication`** — retained `appScope` + `runCatching` around startup work.

### Foundation
- Central Koin dispatcher bindings injected into the IO-bound use cases (**A2 / #46**); `RootSystemGateway` `cachedUserId` hardened.

## Verification
- `./gradlew assembleFossDebug` — **BUILD SUCCESSFUL**.
- Adversarial diff review → fixed: wrong-user `--user 0` caching, cross-VM bus clobber (`ownsInstall`), and the missing Extensions error/retry UI.

## Known limits (please eyeball)
- **#6** — `MainShell`'s async **terminal signal** still needs an API change in `ShellExtensions.kt`/`Shell.kt`; the brick + swallow are fixed, but one async caller path could still not receive a terminal result.
- **#9/#10** — `withTimeout` only interrupts at suspension points, so a fully non-suspending wedged shell call needs one added to be reliably cancellable.

## Deferred to follow-up PRs
A1 (migrate one-off events to `SharedFlow` across ~6 screens), A3 (domain-purity ports / remove `Bitmap`/`Drawable`/`Intent` from domain models), full A2 (VM dispatcher migration), and all P2 items (perf, thread-safety, Compose polish, dead-code, localization).

## Suggested reviewer focus
The concurrency-sensitive changes: `RootSystemGateway` bind/mutex/timeout, `InstallerRepositoryImpl` channel drainer + session handling, `InstallerEventBus` `DROP_OLDEST`/`reset` semantics + `ownsInstall`, and `FreezerTileService`'s app-scoped bulk work.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/ThorApplication.kt
+++ b/app/src/main/java/com/valhalla/thor/ThorApplication.kt
@@ -16,8 +16,10 @@ import com.valhalla.thor.presentation.utils.AppIconKeyer
 import com.valhalla.thor.util.LocaleManager
 import com.valhalla.thor.util.Logger
 import com.valhalla.thor.util.koinLogLevel
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -44,6 +46,11 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
     private val localeManager: LocaleManager by inject()
     private val autoFreezeManager: AutoFreezeManager by inject()
     private val freezerShortcutManager: com.valhalla.thor.data.launcher.FreezerShortcutManager by inject()
+
+    // Retained, cancellable application-lifetime scope. A SupervisorJob keeps one failing
+    // child from cancelling the others, and holding the reference lets us cancel it in
+    // onTerminate so launched work doesn't outlive the process.
+    private val appScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
 
     // Keep the Lazy handle so we can tear the billing client down only if it was actually
     // created this run — resolving the delegate would otherwise spin up a billing connection at
@@ -74,11 +81,15 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
 
         autoFreezeManager.startObserving()
 
-        MainScope().launch {
-            val prefs = preferenceRepository.userPreferences.first()
-            freezerShortcutManager.syncDynamicShortcuts(prefs.addFreezerToLauncher)
-            withContext(Dispatchers.Main) {
-                localeManager.applyLocale(prefs.language)
+        appScope.launch {
+            runCatching {
+                val prefs = preferenceRepository.userPreferences.first()
+                freezerShortcutManager.syncDynamicShortcuts(prefs.addFreezerToLauncher)
+                withContext(Dispatchers.Main) {
+                    localeManager.applyLocale(prefs.language)
+                }
+            }.onFailure { throwable ->
+                Logger.e("ThorApp", "Startup preference sync failed", throwable)
             }
         }
     }
@@ -92,6 +103,7 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
         if (billingProcessorLazy.isInitialized()) {
             billingProcessor.close()
         }
+        appScope.cancel()
         super.onTerminate()
     }
 }
--- a/app/src/main/java/com/valhalla/thor/ThorApplication.kt
+++ b/app/src/main/java/com/valhalla/thor/ThorApplication.kt
@@ -70,6 +70,10 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
         Bypass.setLogger { message, throwable ->
             Logger.e("Bypass", message, throwable)
         }
+        // Install the on-disk offset cache BEFORE the first bypass use (prepareThor below). This
+        // lets the expensive core-oj dex scan be persisted on first launch and reloaded on every
+        // later cold start, instead of re-running the mmap + dex parse each time.
+        Bypass.init(this)
         Bypass.prepareThor()
         ThorShellConfig.init()
 
--- a/app/src/main/java/com/valhalla/thor/ThorApplication.kt
+++ b/app/src/main/java/com/valhalla/thor/ThorApplication.kt
@@ -16,6 +16,7 @@ import com.valhalla.thor.presentation.utils.AppIconKeyer
 import com.valhalla.thor.util.LocaleManager
 import com.valhalla.thor.util.Logger
 import com.valhalla.thor.util.koinLogLevel
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
@@ -93,6 +94,9 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
                     localeManager.applyLocale(prefs.language)
                 }
             }.onFailure { throwable ->
+                // runCatching also catches CancellationException; rethrow it so appScope.cancel()
+                // (onTerminate) isn't logged as a failure and cooperative cancellation is preserved.
+                if (throwable is CancellationException) throw throwable
                 Logger.e("ThorApp", "Startup preference sync failed", throwable)
             }
         }
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -19,12 +19,17 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
 import org.koin.core.annotation.Single
 import java.io.File
 import kotlin.coroutines.resume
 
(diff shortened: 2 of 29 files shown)


trinadhthatakula:
Thanks @gemini-code-assist — all four findings were valid and are now fixed in `621b1b5` (build green: `./gradlew assembleFossDebug`).

| # | Gemini finding | Resolution |
|---|---|---|
| 🔴 | `ExtensionBrowseScreen.kt` — `awaitingInstallResult` is never cleared on `Error`/`Idle`, so a later replayed stale `Success` fires a spurious refresh | Switched the `LaunchedEffect` to a `when` that resets tracking on `InstallState.Error`/`Idle`, gated on the new VM method below |
| 🔴 | `ExtensionBrowseViewModel.kt` — needs a reset method | Added `resetInstallTracking()` (clears `awaitingInstallResult`) |
| 🔴 | `FreezerTileService.kt` — `refreshTile()` after the service is destroyed can throw on `qsTile.updateTile()` (binder gone) | Guarded the post-bulk `refreshTile()` with `if (scope != null)` — the bulk freeze still completes on the process-scoped `appScope`; only the tile UI refresh is skipped when the service is no longer listening |
| 🟡 | `ExtensionManagerViewModel.kt` — hardcoded `Dispatchers.IO` | Now injects `@Named("io") CoroutineDispatcher`, matching the A2 dispatcher-injection pattern introduced in this PR |

Good catch on all three HIGH items — the `ExtensionBrowse` flag leak and the `FreezerTileService` crash were genuine regressions from this PR's own fixes (the `#50` refresh-gating and the `#26` off-service-scope change respectively).

Note taken on the consumer Gemini Code Assist sunset timeline you flagged.

The remaining audit findings not addressed in this PR are catalogued in `docs/audits/dev-branch-audit-2026-07-20.md` and will land in follow-up PRs.

trinadhthatakula:
### Batch 1 — medium-severity audit findings (`60d10f7`)

Second remediation wave landed on this branch (build green: `./gradlew assembleFossDebug`). Fixed the 7 medium findings, then ran an **adversarial self-review over the diffs**, which caught **2 regressions the fixes themselves introduced** — both fixed in the same commit.

| # | Area | Fix |
|---|---|---|
| #16 / #17 | `ShizukuReflector` | exactly one async uninstall + observe the real `IntentSender` result (receiver + `CompletableDeferred` + 15s timeout); reinstall re-queries PM |
| #15 | `AutoFreezeManager` | drop `goAsync()`, dispatch to app scope, `AtomicBoolean` in-flight guard (no ~10s receiver-timeout / no stacking) |
| #48 | `AppList`/`Home`/`MainViewModel` | crash-safe `getInstalledAppsUseCase()` collectors (`Flow.catch`/try-catch, rethrow `CancellationException`, clear loader) |
| #4 | `AppInfoDetailsScreen` | component rows rendered via the parent `LazyColumn` (`componentSection`) instead of `forEach` in a `Column` |
| #5 | `Bypass` | persist the ART-offset disk cache (was never written) → core-oj scan runs once, not every cold start; lazy + double-checked init |
| #8 | `ExtensionManager` | cache extensions by `packageName`+`versionCode` (invalidate on update/uninstall) + single PM scan, no per-resume `PathClassLoader` rebuild + N+1 |

**Self-review regressions caught & fixed:**
- 🔴 `ShizukuReflector` — the new result-broadcast used a **constant action + requestCode 0**, so concurrent (bulk) uninstalls could cross-deliver each other's status → namespaced both by the target package.
- 🔴 `AppInfoDetailsScreen` — the new `LazyColumn` keys weren't unique within a section, so a **duplicate component class name would crash** the screen → keys namespaced by index.

@gemini-code-assist — ready for another pass on `60d10f7`.

trinadhthatakula:
Thanks again @gemini-code-assist — all three findings on `60d10f7` were valid and are fixed in `259a0cb` (build green: `./gradlew assembleFossDebug`).

| # | Finding | Resolution |
|---|---|---|
| 🟡 | `ExtensionManager` — `getInstalledPackages()` can return null → NPE on `.filter` | Added `?: emptyList()` guards — on **both** the `GET_META_DATA` scan in `loadExtensions()` and the lightweight version-code scan |
| 🟡 | `InstallerEventBus` — explicit backing field vs write methods on read-only `SharedFlow` | Refactored to the standard private `_events: MutableSharedFlow` + public `events: SharedFlow = _events`, with `emit`/`reset` on `_events` (public API unchanged) |
| 🟡 | `ExtensionManagerViewModel` — `runCatching` swallows `CancellationException` | Rethrow `CancellationException` in the `onFailure` block before mapping to an error state (matches the pattern used for the `#48` collectors) |

Good call on the `CancellationException` one — I'd added that rethrow to the other three ViewModels in #48 but missed this `runCatching`.

Ready for another pass on `259a0cb`.

trinadhthatakula:
Thanks @gemini-code-assist — all five findings on `259a0cb` were valid and are fixed in `ac74dee` (build green: `./gradlew assembleFossDebug`).

| # | Finding | Resolution |
|---|---|---|
| 🟡 | `ExtensionManager` — `appInfo.sourceDir` null → `PathClassLoader` NPE | `?: return@mapNotNull null` (skip that extension) |
| 🟡 | `ExtensionManagerViewModel` — N+1 `getExtensionPackageName()` PM scans | Fixed at the source: `getExtensionPackageName()` now **reverse-looks-up the in-memory extension cache** first (the passed instances were just loaded, so they're cached), eliminating the per-extension `getInstalledApplications()` scan; PM fallback kept + null-guarded |
| 🟡 | `FreezerTileService` — bulk freeze on the never-cancelled `appScope` can hang forever → leak | Wrapped the `awaitAll` in `withTimeoutOrNull(30_000L)`; a timeout counts as all-failed |
| 🟡 | `AppInfoDetailsScreen` — section expand states lost on config change | `remember` → `rememberSaveable` for the 4 section-expanded flags |
| 🟡 | `ShizukuReflector` — `reinstallApp` re-query races async install | Made `reinstallApp`/`reinstallExistingApp` **`suspend`** and now **await the real `IntentSender` result** via the existing `awaitInstallerResult` (action + requestCode namespaced by target package), instead of the immediate `isAppUninstalled()` re-query. Its only caller (`ShizukuSystemGateway.setAppDisabled`'s unfreeze path) already invokes it inside a suspend `runAction {}`, so no signature ripple |

The `reinstallApp` change is the notable one — it now reports the true async outcome (mirroring the `uninstallApp` #17 fix) rather than racing PMS's state update, so a genuine unfreeze success is no longer misreported as failure.

Ready for another pass on `ac74dee`.

trinadhthatakula:
Thanks @gemini-code-assist — valid, fixed in `d024694` (build green).

`progressJob.join()` removed from the `finally`; the drain coroutine is launched directly inside `coroutineScope { }`, so structured concurrency already awaits it before `coroutineScope` returns — the explicit `join()` was redundant, and (as you noted) a suspending `join()` in a `finally` could mask the real failure (e.g. an `IOException` from `openInputStream`) under cancellation. The `finally` now just does `progressChannel.close()`, which ends the drain loop; the final progress tick still flushes before `commit`.

Ready for another pass on `d024694`.

trinadhthatakula:
Thanks @gemini-code-assist — valid (and a good catch on the hot path), fixed in `537b37a` (build green).

`ExtensionManager.loadExtensions()` now enumerates with `getInstalledPackages(0)` (lightweight), filters to the extension prefix, and fetches `GET_META_DATA` individually only for those few matching packages — instead of a single `GET_META_DATA` scan over every installed app. This removes the `TransactionTooLargeException` / lag risk on devices with many apps (that single-scan tradeoff was introduced by the earlier `#8` change, so this walks it back). Downstream cache/versionCode/sourceDir/metaData logic is unchanged since `getPackageInfo(..., GET_META_DATA)` returns the same populated `PackageInfo`.

Ready for another pass on `537b37a`.

Issue #58, opened 2025-07-30
App lock


nejuni:
Can you also add root method and shizuku method to lock apps as many users have old rooted phone and doesn't have any app lock support root updated recently.
As it is battery friendly and applied more systematically.

trinadhthatakula:
Thanks for filing this, and sorry it's gone a year without a reply.

First, to head off a confusion: v1.93.1's release notes mention "App lock" — that is **Thor's own** biometric lock screen, and has nothing to do with this issue. Locking *other* apps behind a fingerprint or PIN doesn't exist in Thor in any form yet.

This is now written down and tracked in the follow-up list rather than living only as an issue, so it won't get lost. Being honest about the size of it: it needs a Shizuku UserService bound to ActivityTaskManager (or root `am monitor`) to detect launches, an overlay plus a foreground service plus boot persistence to draw the lock, and ongoing per-OEM maintenance — and it carries real Play-policy risk. Dhizuku can't do per-launch gating at all. That's why it hasn't been picked up yet. It's deferred deliberately, not forgotten, and I'd rather say that than leave you guessing.

If you need something in the meantime, Freeze/Suspend in the Freezer gets you partway: a suspended app shows a system dialog instead of opening. It isn't authentication and anyone can dismiss it, but it does stop a casual tap.

I'll update this issue when it moves.

trinadhthatakula:
Status after stable **v1.96.0 (1960)**: locking other apps with Root/Shizuku remains **unimplemented and deferred**.

This request is tracked in the [follow-up list](https://github.com/trinadhthatakula/Thor/blob/v1.96.0/docs/follow-ups/README.md#L363) and [September assessment](https://github.com/trinadhthatakula/Thor/blob/v1.96.0/docs/analysis/community-feature-request-assessment-2026-09.md#L58). It still needs launch detection, an authentication/overlay flow, and lifecycle handling; implementation is not currently scheduled.

Thor's existing biometric lock protects access to Thor itself. It does not fulfill this request to lock other apps, so this issue remains open.

Issue #56, opened 2025-07-25
Thor doesn't work with magisk alpha v29+


selfmusing:
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/fd462346-ab93-49ff-b6b8-24f3a6543c83" />

Check these comments :

https://github.com/topjohnwu/Magisk/pull/9017#issuecomment-3116582870

https://github.com/topjohnwu/Magisk/pull/9017#issuecomment-3116780086

https://github.com/topjohnwu/Magisk/pull/9017#issuecomment-3115187403

Please let me know if you need any logs to debug this .

trinadhthatakula:
New update will be available in a few hours with the bug fix

selfmusing:
> New update will be available in a few hours with the bug fix

Thank you so much :⁠-⁠D

trinadhthatakula:
hi @selfmusing, I have release a new Verison please check it and let me know if everything's working, 
https://github.com/trinadhthatakula/Thor/releases/tag/Thor_v1661

trinadhthatakula:
Thank you, for raising this issue and helping me know what went wrong, I will close this issue now, please keep supporting Thor😁

Issue #55, opened 2025-07-25
Process manager support and profiles for freezing apps


syunkei1:
Can you add support of process manager for both user and system apps by shizuku so that we can know which app is running and how much ram or cpu it is consuming so that we can freeze that app.
Also add support for various profiles used for freezing apps so we can add apps on profiles and freeze them whenever necessary

trinadhthatakula:
Hi @syunkei1 
Profiles for freezer are almost ready, by Sunday there will be a test version available so soon Freezer will have a profile and if all goes well an option to add home screen widget,

Process management is a good idea and I will keep this issue Open and get back to you after checking it's feasibility.

trinadhthatakula:
Half of this shipped in v1.93.1.

**Freeze profiles — done.** The Freezer screen has a "Freeze Profiles" button in the top toolbar. Create a named profile, pick apps into it with a searchable picker, then freeze or unfreeze the whole group in one tap. Each profile has edit and delete, and you can also multi-select apps in the Freezer and use "Save as profile" to turn the selection into one. Deleting a profile only removes the grouping — anything frozen stays frozen. An app can belong to several profiles, freezing profile A then profile B no longer silently skips B, and unfreeze is never filtered, so you can always get back out.

Two things to know: the freeze/unfreeze buttons need an active privilege backend (root/Shizuku/Dhizuku) — creating and editing profiles works without one — and profiles can currently only be run from inside the app, not from the Quick Settings tile or a launcher shortcut.

**Process manager — not shipped.** There's still nothing in Thor showing which apps are running or what they're using. The reason is technical rather than a lack of interest: since Android 8, `getRunningAppProcesses` returns only your own process, so live RAM/CPU means parsing `dumpsys meminfo` / `top` / `ps`, which is brittle across OEMs and OS versions, and CPU% needs sampled polling on top. It's also root/Shizuku-only — **Dhizuku has no shell at all**, so that whole path is a dead end.

Because of that I'm leaning towards building it as an optional **extension** rather than into Thor itself, so the fragile parsing can iterate on its own release cycle without destabilising the app. I want to be upfront that this narrows the audience: Thor's extensions are LSPosed/Xposed modules, so it wouldn't reach a Shizuku-only setup. If that makes it useless for you, please say so — that genuinely changes the calculation, and you're the person who asked for it.

[...]

Issue #51, opened 2025-07-10
New feature requested


sankarru:
If there is a new feature like data backup in future can we also get an (app and data) transfer option between two mobiles as it's missing in many backup apps which is most useful for people like me.

trinadhthatakula:
Update, and an honest one: **what shipped is export, not backup.** This issue stays open.

**What's in v1.93.1:** select as many apps as you like (Apps tab or Freezer) and hit Export — Thor writes one installer file per app into a folder you choose, in a single cancellable run, with progress in the bottom bar that keeps going while you move around the app. Alongside the files it writes a `thor-backup-<timestamp>.json` listing every app it attempted, including the ones that failed, so the folder is never a pile of unlabelled files. Single-app export also gained `.xapk` next to `.apk`/`.apks`.

**What that is not:**

- **No app data.** Settings, game saves, logins, anything under `/data/data` — none of it is captured. This is APK bytes only.
- **No restore.** Nothing in Thor reads a backup folder back. Recovery today means reinstalling each file by hand.
- **No phone-to-phone transfer**, which is literally what you asked for. Rebuilding a Nearby-Share-style transport isn't something I'm going to take on; the share sheet is the answer there.

One correction on the thread: data sync **can't** work on Shizuku. Shizuku runs as the shell uid, which cannot read `/data/data`; Dhizuku has no file access at all; and `adb backup` is dead on modern Android. Real app-data backup is root-only. That's a hard platform limit, not a missing feature.

I also owe you a straight answer about the "recycle bin for uninstalled apps, then a full-fledged backup" I promised on this thread back in 2025 — that hasn't happened. What's here is a manual multi-select export, not automatic capture on uninstall.

I'm rethinking the backup design from scratch rather than bolting a next phase onto this, so I'd rather not put a version number on it than give you one I'd miss.

In the meantime, please try the export half in **[Dev Build v1.93.1 (103)](https://github.com/trinadhthatakula/Thor/releases/tag/v1.93.1-dev-103)** and tell me whether it's useful as it stands. And if you have a specific shape in mind for restore — whole-device or per-app, data included or APKs only, one archive or a folder — say so here. It'll feed straight into the design.

sankarru:
[...]

PR #257, opened 2026-07-20
refactor(audit): A-series architecture cleanups + P2 tail

trinadhthatakula:
## Summary

Second remediation pass from the 2026-07-20 codebase audit (`docs/audits/dev-branch-audit-2026-07-20.md`): the **A-series architecture cleanups** + the **P2 tail** deferred from PR #256. **33 files, net −176 LOC**, `./gradlew assembleFossDebug` ✅, applied via file-disjoint agents and adversarially reviewed after each wave.

## A-series (MVVM / architecture)

- **A1 — one-off events → `SharedFlow`:** moved transient toasts/prompts out of the single `UiState` `StateFlow` onto replay-0 event flows collected via a new lifecycle-safe **`ObserveAsEvents`** helper — `Main` (#62, via the existing `MainSideEffect`), `Settings` (#24/#65), `AppList` (#62), `AppInfoDetails` (#28), `Freezer` (#62), `Permission` (#23). (Freezer/AppList event flows use a 1-slot `DROP_OLDEST` buffer so a load-failure toast emitted from `init{}` isn't dropped before the screen subscribes.)
- **A2 — inject dispatchers (#56):** `@Named("io")` `CoroutineDispatcher` into `Main`/`Settings`/`Home` VMs, replacing hardcoded `Dispatchers.IO`.
- **A4 — framework types out of VMs:** `AppList` `Context`→`UiText` (#20/#49); `Main` (#63) & `Installer` (#59) `PackageManager` removed — `Main` uses the already-resolved `AppInfo.isDebuggable`, `Installer` routes version detection through `AppRepository.getAppDetails`.
- **A5 — single `UiState` (#64): partial** — `Settings` action message moved to events; the standalone `preferences`/`extensionConsentAccepted` flows are still read by `HomeActivity`/`ExtensionManagerScreen`, so the full dedup (needs those readers migrated without coupling startup to the privilege probe) is left as a follow-up. `#58` (Installer split-state) intentionally not done — `installState` genuinely originates from the app-scoped bus, and consolidating would risk the installer hardened in #256.

## P2 tail

- **Perf:** #35 (drop per-action Shizuku probe), #36 (short-circuit gateway probing), #47 (refresh-only detail path), #51 (legacy probe off the main thread), #53 (offload bulk-pin bitmap).
- **Thread-safety:** #33 (suCore `getInstance` `@Synchronized`), #41 (`cachedUserId` `@Volatile`), #55 (Home dashboard stats race).
- **Correctness:** #40 (escape shell args), #37 (suspend Room DAO), #42 (`printStackTrace`→`Logger`).
- **UI/cleanup:** #29 (`rememberSaveable`), #66 (derived flags), #54 (bottom-inset dedup), #30 (reuse string resources), #32/#38/#39 (dead code removed).

## Verification
- `./gradlew assembleFossDebug` green.
- Per-wave adversarial review: Wave 1 → 1 fix (init-time event drop); P2 → 0 regressions. Findings that would create a circular DI (#36) or destabilize the installer (#58) were deliberately handled/skipped with rationale in the commit messages.

## Deferred to a follow-up PR
- **A3 (domain purity, #18/#19/#45):** `ExportAppUseCase`/`ShareAppUseCase` are ~140 lines of MediaStore/SAF I/O returning a framework `Uri`, and removing `Bitmap`/`Drawable`/`Intent` from the domain models ripples into APK icon rendering + the install-confirmation flow. That warrants its own PR **with on-device testing** rather than a compile-only change here.
- **A5 full dedup (#64)** and **#58** as noted above.

## Suggested reviewer focus
`SystemRepositoryImpl.getActiveGateway` selection equivalence (#36), the `ObserveAsEvents` event semantics (A1), and `HomeViewModel`'s reactive stats derivation (#55).


trinadhthatakula:
Thanks @gemini-code-assist — 3 of 4 addressed in `c1ead83` (build green: `./gradlew assembleFossDebug`), and one respectfully declined with rationale.

| Finding | Resolution |
|---|---|
| 🟡 `AppInfoDetailsViewModel` `_events` buffer | ✅ Added `extraBufferCapacity = 1, onBufferOverflow = DROP_OLDEST` (matches AppList/Freezer) |
| 🟡 `PermissionManagerViewModel` `_events` buffer | ✅ Same |
| 🟡 `SettingsViewModel` `_events` buffer | ✅ Same |
| 🔴 `RootServiceServer` → `context.applicationContext` | ⛔ **Not applied — it would break RootService bootstrap** |

On the `RootServiceServer` one: the constructor **requires the context to be a `Callable`** — it does `context.call()` to obtain `[authorizedUid, isDaemon]` and explicitly `throw IllegalArgumentException("Expected Context to be Callable")` otherwise (RootServiceServer.kt:67–80). This is libsu's internal **root-service-process bootstrap context**, not an `Activity`, so there's no Activity-leak risk (the singleton is scoped to a process whose entire purpose is hosting this service). `context.applicationContext` is *not* `Callable`, so applying the suggestion would throw at construction and break root-service IPC. Left as-is intentionally.

Ready for another pass on `c1ead83`.

trinadhthatakula:
Thanks @gemini-code-assist — 3 of the 4 new findings fixed in `ccea9ea` (build green), plus the re-raised `RootServiceServer` one declined again with rationale.

| Finding | Resolution |
|---|---|
| 🔴 `ObserveAsEvents` stale lambda | ✅ Wrapped `onEvent` in `rememberUpdatedState` — good catch, the screen handlers close over `context`/snackbar state |
| 🔴 `AppListScreen` freezer-prompt snackbar not cleared | ✅ Clear the local `freezerPrompt` in `onAddToFreezer` |
| 🔴 `FreezerScreen` freezer-prompt snackbar not cleared | ✅ Same |
| 🟡 `AppInfoDetailsViewModel.refreshDetails` redundant `launch` | ✅ Made it `suspend` and call it directly inside the existing action coroutines (structured concurrency; avoids out-of-order refreshes) |
| 🟡 `RootServiceServer` → `applicationContext` (re-raised) | ⛔ Declined again — see below |

Re-declining `RootServiceServer`: its constructor **requires a `Callable` context** — it calls `context.call()` to read `[authorizedUid, isDaemon]` and `throw IllegalArgumentException("Expected Context to be Callable")` otherwise (RootServiceServer.kt:67–80). This is libsu's **root-service-process bootstrap context**, not an `Activity` (so no Activity-leak is possible — the singleton lives in a process whose sole purpose is hosting this service), and `context.applicationContext` is not `Callable`, so the suggested change would throw at construction and break root IPC.

Ready for another pass on `ccea9ea`.

trinadhthatakula:
### Self-review pass (`478bb5f`)

Ran a holistic adversarial code-review over the whole branch diff; it surfaced **8 confirmed issues in this PR's own changes** — all fixed here (build green):

- 🔴 **One-off events were being dropped (systemic).** The A1 event flows used `MutableSharedFlow(replay = 0, extraBufferCapacity = 1, DROP_OLDEST)`. A `replay=0` SharedFlow **discards** any value emitted while no collector is subscribed — `extraBufferCapacity` only relieves back-pressure for *already-subscribed* collectors, it doesn't retain for a late one. So an init-time load-failure toast (e.g. `FreezerViewModel`, created eagerly by `MainScreen` while another tab is shown) or a rotation/background collector-gap emit was silently lost. Converted `Freezer`/`AppList`/`Settings`/`Permission`/`AppInfoDetails` to **`Channel(Channel.BUFFERED).receiveAsFlow()`** (matching `MainViewModel`), and corrected the `ObserveAsEvents` KDoc. *(Note: the `extraBufferCapacity` buffer suggested in the previous review round doesn't fix this — only a Channel or `replay≥1` does.)*
- 🟠 **`InstallerViewModel`: reverted `#59`.** Routing version detection through `AppRepository.getAppDetails` had two regressions — `AppInfo.versionCode` is `Int` (truncates `versionCodeMajor` apps → wrong downgrade detection), and `getAppDetails` queries `MATCH_UNINSTALLED_PACKAGES` (so uninstalled/archived packages were treated as an "update"). Restored `getPackageInfo(pkg, 0)` + `getLongVersionCode`. The A4 goal for this VM is deferred to the A3 follow-up.
- 🟡 **`ExtensionManagerViewModel`:** the legacy-migration probe no longer flaps `isLegacyInstalled` (dismissed dialog can't re-appear), and checks the single legacy package instead of a second full `getInstalledPackages(0)` per reload.

@gemini-code-assist — ready for another pass on `478bb5f`.

trinadhthatakula:
Thanks @gemini-code-assist — all 6 findings addressed in `9e30984` (build green: `./gradlew assembleFossDebug`).

| # | Finding | Resolution |
|---|---|---|
| 🔴 | `SystemRepositoryImpl` per-app gateway resolution in batches | ✅ Short-TTL (3s) `@Volatile` cache of the resolved `SystemGateway`; a whole batch now resolves once (selection logic moved verbatim into `resolveActiveGateway()`, unchanged). |
| 🟡 | `MultiSelectToolBox` derivations re-scan every recomposition | ✅ `remember(selected)` on all four |
| 🟡 | `PermissionManagerViewModel` hardcoded load-failure string | ✅ `error.message` primary + `UiText.StringResource(R.string.failed_to_load_permissions)` fallback |
| 🟡 | `PermissionManagerViewModel` hardcoded success string | ✅ `UiText.StringResource(R.string.permission_status_updated)` |
| 🟡 | `PermissionManagerViewModel` hardcoded modify-failure string | ✅ `error.message` primary + `R.string.failed_to_modify_permission` fallback |
| 🟡 | `FreezerViewModel.observeApps` no retry on transient failure | ✅ bounded `retryWhen` (≤2 retries, 500ms, never `CancellationException`); the existing catch still fires after retries are exhausted |

(3 new localized string resources added.) Ready for another pass on `9e30984`.

trinadhthatakula:
Thanks @gemini-code-assist — respectfully declining this one; the `@Volatile` here is load-bearing, not redundant.

`@Synchronized` on `getInstance()` only establishes visibility for threads that acquire the *same monitor*. But `RootServiceServer` also exposes a **lock-free** reader:

```kotlin
@JvmStatic fun getInstanceOrNull(): RootServiceServer? = mInstance   // no lock
```

`getInstanceOrNull()` reads `mInstance` without synchronization, so without `@Volatile` there is no happens-before with the write inside the synchronized `getInstance()` — that reader could observe a stale `null` / unsafely-published reference. `@Volatile` is exactly what gives the lock-free reader correct visibility + safe publication.

The `RootServiceManager` comparison doesn't apply: it has *no* lock-free accessor — every `mInstance` access goes through its `@Synchronized getInstance()`, so there `@Volatile` genuinely would be redundant. The asymmetry is intentional (keep `@Volatile` only where an unsynchronized reader exists). Keeping it on `RootServiceServer`.

PR #258, opened 2026-07-20
refactor(audit): A3 — domain-layer purity (#18/#19/#45/#64)

trinadhthatakula:
## A3 — Domain-layer purity (audit findings #18 / #19 / #45 + A5-full #64)

Removes Android **framework** types from the `domain` layer so it's Clean-Architecture-pure and unit-testable without Robolectric. Done as **6 behavior-preserving stages**, each built green, plus a **7th self-review stage** that fixed 3 regressions the holistic adversarial branch review caught.

`20 files changed, +387/−254` vs `dev`.

### Stages
| # | Finding | Change |
|---|---------|--------|
| 1 | #45/#18/#19 | `AppInfo` → pure data class; `mapToAppInfo()` moved to `data/AppInfoMapper.kt`. Export/Share use cases go android-free via a new `AppBundleFileStore` port (impl in data holds MediaStore/SAF/FileProvider). |
| 2 | #45 | `AppMetadata.icon: Bitmap?` → `iconPath: String?`; `AppAnalyzerImpl` persists the parse-time icon to a cache file, `PortableInstaller` renders it with Coil3 `AsyncImage(File(...))`. |
| 3 | #45 | `InstallState.UserConfirmationRequired(Intent)` → payload-less `data object`; the confirm `Intent` now rides a data-layer `PendingInstallIntent` holder (consume-once). |
| 4 | #45 | Deleted dead `ApkDetails` model (carried a `Drawable`, zero references). |
| 5 | #64 | Deduped `SettingsViewModel`'s standalone prefs/consent flows; the two external readers use `PreferenceRepository` directly (single source of truth). |
| 6 | #19 | `AppBundleBuilder` → domain interface + data impl; removes the **last** `android.content.Context` from domain. |
| 7 | review | See below. |

### Stage 7 — self-review fixes (3 CONFIRMED regressions)
- **[med] Installer icon stale-cache** — the stage-2 cache file was keyed by package name only; Coil keys its `File` memory cache by path alone (`addLastModifiedToFileCacheKey` defaults false), so re-analyzing a **new version** of the same package showed the **old** icon. Fixed: key the file by `packageName + versionCode` (busts the cache across versions).
- **[low] Icon write race** — fixed filename + truncating stream let two concurrent same-package analyses corrupt the PNG. Fixed: write to a unique temp then **atomic `renameTo`** (reader always sees a complete file).
- **[low] Consent write cancellable** — `onConsent` wrote on `rememberCoroutineScope` (composition-tied); a config change right after "Accept" could cancel the DataStore edit. Fixed: moved to `ExtensionManagerViewModel.acceptExtensionConsent()` on the durable `viewModelScope`.

### Accepted residual (out of #45 scope)
`android.net.Uri` remains in 3 repository interfaces (`AppAnalyzer` / `InstallerRepository` / `StoreRepository`) + `ShareAppUseCase`'s return — a thin content-handle, not the heavy `Bitmap`/`Drawable`/`Intent`/`PackageManager` mapping #45 targeted (all removed). Removing it = install/store churn for little purity gain. `androidx.compose @Immutable` stability hints on domain models are also retained intentionally.

---

## ✅ On-device verification — DONE (maintainer)
**Manually verified on-device on BOTH debug and release builds — all 5 confirmed working:**

- [x] **Installer icon** renders on the *ReadyTo Install* sheet and the *Installing* state (Coil loads async — a brief styled placeholder box before the icon paints is expected).
- [x] **Icon freshness** — analyze/install **v1** of an app, then a **v2 with a different icon** in the same session: the *new* icon shows (this is the stale-cache fix).
- [x] **Install-confirmation dialog** still appears when `PackageInstaller` requests user action (`STATUS_PENDING_USER_ACTION`).
- [x] **Extensions consent gate** — the liability sheet shows exactly once for an un-consented user, never for a consented one (no first-frame flash); tapping **Accept** persists even across a rotation.
- [x] **Export APK** and **Share APK** flows (Downloads/SAF target + share-sheet) still work and land the file correctly.

> Note: testing surfaced an unrelated **pre-existing** bug (the "Freeze System App?" dialog also shows for *user* apps in the App-details screen; not introduced by this PR — traced to an earlier commit, and this branch doesn't touch that file). It's being fixed in a **separate PR** off `dev`.

CI must be green before merge.


trinadhthatakula:
@gemini-code-assist thanks — all three findings addressed in `7bd8dcd`:

1. **`AppBundleFileStoreImpl` main-safety** ✅ — injected `@Named("io")` dispatcher and wrapped the four suspend members (`writeToDownloads` / `writeToTree` / `isTreeWritable` / `currentTargetLabel`) in `withContext(ioDispatcher)`, so the blocking MediaStore/SAF/disk I/O is now off the caller's thread. `shareUri` stays a plain `fun` (in-memory FileProvider lookup, no I/O).

2. **`AppAnalyzerImpl.persistIcon` temp-file leak** ✅ — wrapped the write + atomic-rename in `try/finally` that always deletes `tmp` when it still exists. A successful `renameTo` consumes `tmp`, so the `finally` is a no-op on the happy path and only cleans up the `copyTo` fallback (or any mid-write failure).

3. **`PortableInstaller` launch crash** ✅ — wrapped `startActivity` in `try/catch` → `viewModel.resetState()` (its existing Idle-reset, consistent with the other call sites). Added a `CancellationException` rethrow since this runs inside a `LaunchedEffect` coroutine.

Build green (`assembleFossDebug`).

PR #260, opened 2026-07-20
feat(home): adaptive bento actions grid + Extension Manager entry

trinadhthatakula:
## Home: adaptive bento actions grid + Extension Manager entry

Surfaces an **Extension Manager** entry directly on Home (previously buried under Settings → scroll → Extension Manager), and restructures Home's stacked action cards into an **adaptive 2×2 "bento" grid**.

### What changed
- **New pure logic** `homeActionRows(...)` + `HomeAction` enum — encodes the visibility matrix, unit-tested (5 cases).
- **New `BentoTile`** — compact, size-adaptive tile (replaces the old full-width `ActionCard`, which is deleted).
- **New `HomeActionsBento`** — the grid, shared by both layout branches:
  - Row 1 = `[Reinstall All?] · [Install from file]`; Row 2 = `[Clear all cache (ROOT)?] · [Extension Manager (privilege≠NONE)?]`.
  - When a row's conditional tile is hidden, the surviving tile spans full width; `animateContentSize` smooths reflow.
- **Extension Manager tile** — visible only when `activePrivilegeMode != null`; opens the shared `ExtensionManager` destination via `homeBackStack`. The shared `ExtensionManager`/`ExtensionBrowse` entries now pop/push the **active tab's** back stack (`currentBackStack`), so back-nav is correct from Home *and* Settings (previously hard-coded to `settingsBackStack`).
- **Adaptive** across `COMPACT / MEDIUM / EXPANDED` width classes; tiles are weight-based; on EXPANDED the content is centered and capped at 1200dp so it doesn't sprawl.
- **QA polish (on-device):** fixed the Support & Community card to align its margins with the App Distribution card in the wide two-pane, and to stack its Play Store/GitHub/Telegram buttons vertically in narrow panes so the labels don't crop.

### Verification
- `HomeActionsTest` 5/5 pass; `assembleFossDebug` + `assembleStoreDebug` green.
- Built via subagent-driven development: fresh implementer + two-stage (spec + quality) review per task, plus a whole-branch review (its one Important finding — an EXPANDED width-cap modifier-order bug — is fixed in `1ab4a9f`).
- **On-device (maintainer):** bento confirmed on multiple phones + emulator; Support & Community fix confirmed.

### Deferred (non-blocking)
- **i18n:** `home_extensions_title` / `home_extensions_subtitle` need `ar/es/fr/zh-rCN` translations (the only lint findings — 2 `MissingTranslation`; the rest is the 28 pre-existing warnings). Same maintainer i18n workflow as before.
- Minor (from review, non-blocking): `BentoTile` dismiss touch target is 28dp; App-Distribution chart `Column` is duplicated across both branches (pre-existing).

### DIFF
--- a/app/src/main/java/com/valhalla/thor/presentation/home/HomeScreen.kt
+++ b/app/src/main/java/com/valhalla/thor/presentation/home/HomeScreen.kt
@@ -100,9 +100,9 @@ fun HomeScreen(
     ) {
         Column(
             modifier = Modifier
+                .then(if (isExpanded) Modifier.widthIn(max = 1200.dp) else Modifier)
                 .fillMaxWidth()
                 .align(Alignment.TopCenter)
-                .then(if (isExpanded) Modifier.widthIn(max = 1200.dp) else Modifier)
         ) {
         // 1. Header (full width always)
         DashboardHeader(
--- a/app/src/main/java/com/valhalla/thor/presentation/home/HomeScreen.kt
+++ b/app/src/main/java/com/valhalla/thor/presentation/home/HomeScreen.kt
@@ -274,7 +274,10 @@ fun HomeScreen(
 
             // 4. Social Links
             Spacer(Modifier.height(8.dp))
-            SupportCommunitySection(onSupportClick = { showSupportSheet = true })
+            SupportCommunitySection(
+                modifier = Modifier.padding(horizontal = 24.dp),
+                onSupportClick = { showSupportSheet = true }
+            )
         }
         Spacer(Modifier.height(32.dp))
         }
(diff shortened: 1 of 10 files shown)


trinadhthatakula:
@gemini-code-assist both addressed in the latest commit:

1. **BentoTile touch target** — removed `.size(28.dp)` from the dismiss `IconButton`, restoring the default 48dp touch target (icon glyph stays 18dp). Prevents accidental main-`onClick` triggers.
2. **SupportCommunitySection threshold** — raised 250 → 320.dp, so a standard 360dp portrait phone (inner ~264dp) now stacks the Play Store / GitHub / Telegram buttons vertically instead of truncating them.

Build green.

PR #261, opened 2026-07-20
fix(ui): make bottom sheets vertically scrollable in landscape

trinadhthatakula:
## Make bottom sheets scrollable in landscape (fix cropped content)

On phones in **landscape** (and any short height), fixed-content bottom sheets overflowed with no way to reach content below the fold. Most visible: the **app-info sheet** (`AppInfoDialog`) — in phone-landscape with detail-view off, only the app header (icon/name/chips) showed and **all the action buttons were off-screen with no scroll**.

### Fix
Added `Modifier.verticalScroll(rememberScrollState())` to the content column of the fixed-content sheets:
- `AppInfoDialog` (the reported one)
- `ExportBottomSheet`
- `FreezerSettingsSheet`
- `PortableInstaller` (installer logic — effects/receiver/viewModel — left untouched)

`ManageFreezerSheet` was **left as-is**: its content is a fixed header + a `LazyVerticalGrid` that already scrolls; adding another vertical scroll would crash. Inner `horizontalScroll` rows are a different axis and unaffected.

On wide/tall devices there's no behavior change (content already fits).

### Verification
- `assembleFossDebug` + `assembleStoreDebug` green.
- Fixed debug build installed on-device for maintainer verification (open an app in phone-landscape → the sheet now scrolls to reveal the actions).

### DIFF
--- a/app/src/main/java/com/valhalla/thor/presentation/appList/ExportBottomSheet.kt
+++ b/app/src/main/java/com/valhalla/thor/presentation/appList/ExportBottomSheet.kt
@@ -8,6 +8,8 @@ import android.widget.Toast
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.foundation.background
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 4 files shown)


trinadhthatakula:
@gemini-code-assist good catch — fixed in the latest commit. Hoisted the sheet's `ScrollState` and added `LaunchedEffect(state) { scrollState.scrollTo(0) }` so the scroll position resets to top on every install-state transition (ReadyToInstall → Installing → Success/Error), avoiding a carried-over offset. Build green.

PR #264, opened 2026-07-21
Odin Phase 1: prep :suCore (MainShell hang fix + ThorRootService→:app + Apache-2.0)

trinadhthatakula:
## Odin Phase 1 — prepare `:suCore` for Maven Central extraction

Phase 1 of publishing Thor's in-house libsu fork **Odin** (`:suCore`, `com.valhalla.superuser`) to Maven Central as `com.trinadhthatakula:odin`. This PR preps the module *in place* — no publishing yet (Phase 2/3). Built via subagent-driven SDD; spec + plan in `docs/superpowers/{specs,plans}/2026-07-21-odin-phase1-sucore-prep.*`.

### What's in this PR
- **MainShell shell-init hard-failure fix** (closes `docs/follow-ups/mainshell-shell-init-hard-failure.md`, the CodeRabbit Major deferred from #263). `getShellAwait()` → `isRootGranted()` could suspend forever on a hard shell-init failure. Added a default no-op `onShellDied(error: Throwable?)` to `Shell.GetShellCallback`, dispatched from `MainShell.get()`'s catch; `getShellAwait()` now resumes exceptionally, and `isRootGranted()` is bounded by a 10s `withTimeoutOrNull` + `getOrElse { rethrow CancellationException else false }` so it resolves to `false` instead of hanging.
- **Make Odin generic** — moved the Thor-specific `ThorRootService.kt` + `IThorRootService.aidl` out of `:suCore` into `:app` (`com.valhalla.thor.rootservice`). Added a reusable `protected fun RootService.enforceCaller()` to Odin's generic base (keeps the `internal` caller-UID enforcement inside the module); replaced hardcoded `com.valhalla.thor`/`.debug` strings with `BuildConfig.APPLICATION_ID`; updated `RootSystemGateway` + manifest. `grep com.valhalla.thor suCore/src` is now empty.
- **Self-containment prep** — declared the `kotlinx-coroutines-android` dependency in `:suCore` (was resolving only via ambient classpath); added a canonical Apache-2.0 `suCore/LICENSE`; flipped the README license note GPL-3.0 → Apache-2.0.

### Verification
- `assembleFossDebug` + `assembleStoreDebug` green.
- **On-device verified (maintainer), debug + release:** root probe resolves on rooted & non-rooted with no UI hang; root suspend / clear-data still work post-move ("Managed by Thor" shown).
- Each task passed an independent spec+quality review; final whole-branch review = ready to merge.

### Deferred to Phase 2 (extraction/publish) — intentional
[...]

PR #265, opened 2026-07-22
fix(app): fix release-only app suspend broken by R8 (RootService daemon)

trinadhthatakula:
## Problem
App **suspend** worked in debug but failed in **release** builds; every other root op was fine. Reported on-device (Android 16 / API 36) with both a `NoSuchMethodException` and, after an initial keep-rule attempt, a `NoSuchFieldError`.

Suspend is the **only root op with no shell fallback** (GH#239: a shell `pm suspend` records the suspender as `"root"` and crashes `SuspendedAppActivity` on tap), so it's the only op that must go through the `ThorRootService` AIDL daemon + hidden-API reflection — and thus the only one R8 could break.

## Root cause — two independent release-only breakages
**1. AIDL daemon stripped/renamed.** Phase 1 (#264) moved `ThorRootService` + `IThorRootService` out of `:suCore` (`com.valhalla.superuser.ipc`, kept by the shell lib's consumer rules) into `:app` (`com.valhalla.thor.rootservice`) with no matching keep rule. The daemon is loaded **by class name** in the root `app_process` and dispatched over Binder, so R8 renaming/shrinking it broke the transaction.

**2. Bundled framework shadows optimized into the release dex.** The compile-time SHADOW stubs `app/src/main/java/android/content/pm/{IPackageManager,IPackageInstaller}.java` were dexed into the app. In release, R8 renamed them (`IPackageManager$Stub -> v91`) and rewrote every reference to match — `Class.forName(...)` string literals **and** `IPackageInstaller::class.java` const-class refs. But at runtime the real bootclasspath classes win **parent-first**, so reflection then targeted the renamed abstract stub instead of the framework class → `NoSuchMethodException` (`asInterface`) → `NoSuchFieldError` → and the same latent break on the Shizuku `installExistingPackage` path. Debug worked only because names aren't renamed and resolve parent-first to the framework.

## Fix
- **`-keep class/interface com.valhalla.thor.rootservice.**`** — protects the reflectively-loaded AIDL daemon.
- **Delete both shadows.** A `-keep android.content.pm.**` rule was a band-aid — it kept the shadow in the dex, so R8 kept optimizing against it (that produced the `NoSuchFieldError`). Removing them means nothing app-side shares those names, so R8 has nothing to rename or string-adapt and every `Class.forName("android.content.pm...")` resolves to the genuine framework class. Their only compile-time *type* use was `IPackageInstaller` as an **opaque handle** in the Shizuku installer path (already driven by the in-house `:bypass` reflection) → replaced with the real supertype `android.os.IInterface` + `Class.forName("android.content.pm.IPackageInstaller")`.

No behavior change in debug; release now resolves the same framework classes debug already did.

## Verification
- `./gradlew :app:assembleFossRelease` succeeds; `mapping.txt` no longer lists the shadows as app classes; `IThorRootService`/`$Stub`/`asInterface` keep original names.
- **On device (API 36):** release suspend + unsuspend work with **both root and Shizuku**; "Managed by Thor" dialog path intact. Logcat: `ROOT caller 'com.valhalla.thor' SUCCEEDED`, `taskResult=true nowSuspended=true`.


trinadhthatakula:
### PR Code Review Complete

**Model Used:** Gemini 3.5 Flash
**Verdict:** **Approved** (Recommended for Merge)

**Summary:**
1. **AIDL Daemon Protection (`app/proguard-rules.pro`):** Keep rules correctly protect the reflectively-loaded daemon/AIDL classes (`ThorRootService`) in release builds, preventing optimization-induced failures in `app_process`.
2. **Hidden-API Shadow Elimination (`android.content.pm.*`):** Removing compile-time shadow files and substituting generic `IInterface` / dynamic `Class.forName` is the correct solution. It prevents R8 from rewriting references in release builds to classes that are shadowed by the real runtime bootclasspath.
3. **Reflector Update (`ShizukuPackageInstallerUtils.kt`, `ShizukuReflector.kt`):** Dynamic class name resolution correctly routes reflection to real bootclasspath classes, matching debug behavior.

PR #276, opened 2026-07-26
fix(installer): gate downgrade detection on a known versionCode + surface the codes

trinadhthatakula:
## The report

Community report via Telegram (screenshot only): *"the Portable Installer considers an upgrade as a downgrade."*
`com.rebelroot.omni` — installed `1.2.4.7`, picked `1.2.5.1`, plain `.apk`.

## Verdict: for that app, Thor is right — but it had no way to say so

The starting hypothesis was that we compare version **names**. We don't. The decision is
`Long`-vs-`Long` on `versionCode` (`InstallerViewModel.kt`), and the only lexicographic
version-name comparison in the repo is `SortBy.VERSION_NAME` (`AppSorting.kt:17`), which feeds the
app-list sorter and never reaches the installer.

Omni's `versionCode` is hand-maintained and **went backwards**. From
[REBEL-ROOT/omni-browser](https://github.com/REBEL-ROOT/omni-browser) `app/build.gradle.kts` at each
release tag:

| tag | versionName | versionCode |
|---|---|---|
| v1.2.4.6 | 1.2.4.6 | 26 |
| **v1.2.4.7** | 1.2.4.7 | **27** ← installed |
| **v1.2.5.1** | 1.2.5.1 | **25** ← picked |
| 1.2.6.0 | 1.2.6 | 25 |

The name reads as an upgrade while the code goes backwards. Android sequences updates by code
alone, so this is a genuine downgrade — the stock installer refuses it too. That's an upstream
versioning defect in Omni, and anyone on 1.2.4.7 is currently stuck: 1.2.6 also ships code 25.

> An earlier revision of this PR claimed a "drops the third name component" encoding
> (1.2.4.7 → 1002007, 1.2.5.1 → 1002001), reverse-engineered from one online sample. That was
> wrong — see `6e974ca`. Their v1.2.3.4 did ship `versionCode = 1002004` from a scheme they
> abandoned one release later, which is where such codes in old listings come from. The verdict
> was right; the reasoning wasn't.

## What *was* broken

**1. `AppAnalyzerImpl.kt:180` — sidecar version code collapsed to `0`.**
```kotlin
val versionCode = (manifest?.versionCode ?: apkmInfo?.versionCode)?.toLongOrNull() ?: 0L
```
Both sidecar fields are `String?`. Any non-numeric value — a version **name** written into
`version_code`, a blank, padding, an absent key — became `0`. And the elvis tested only `null`, so a
blank `manifest.json` value shadowed a perfectly good `info.json` one. This is the one place the
version-name hypothesis was literally true. Extracted to `resolveSidecarVersionCode()`, which tries
each source in turn, trims, and rejects non-positive values.

**2. `0` was then treated as authoritative.** It loses against every installed app, so a file whose
code couldn't be read reported a downgrade of whatever was installed, however new it actually was —
and in NORMAL mode that's a hard refusal with `error_downgrade_privilege`, no override. Added the
pure domain predicate `isVersionDowngrade()`, which treats a non-positive code as *unknown* and
never a downgrade. Same `0 = unknown` convention the extension store already uses
(`ExtensionCatalog.versionCode`); the installer was the outlier.

**3. The panel showed version names only** — the codes the verdict is made on appeared nowhere, so
a correct downgrade call was indistinguishable from a bug. That's precisely how this got reported.
`ReadyToInstall` now carries `oldVersionCode`; both codes render under their names, and the
downgrade warning spells the comparison out. Codes are suppressed when unknown rather than shown
as `(0)`.

## Tests

Pure JUnit4, no new test dependencies:
- `VersionCompareTest` (new, 8) — unknown/negative gate, equality, `versionCodeMajor` packing, and
  pins the Omni case (25 over 27) so nobody "fixes" it by comparing names.
- `BundleAnalysisTest` (+6, 24→30) — blank / version-name / padded / absent / non-positive sidecar codes.

Verified **red-before-green**: a throwaway test ran the pre-fix expressions and confirmed they
really did yield `0` for blank, version-name and padded inputs (and that `0 < 4210` is true), then
was deleted.

## Verification

- 93 unit tests pass, 0 failures.
- `assembleFossDebug` + `lintFossDebug` green — only the 5 pre-existing intentional `VectorPath` warnings.
- Explainer string added in all 5 maintained locales (en/ar/es/fr/zh-rCN).
- No `versionCode` bump — this repo bumps only in `chore(release)` commits.

## Not verified

**No on-device run.** The UI change (version-code lines + the explainer `Column`) is compile- and
lint-verified only; nobody has looked at the rendered layout, including in landscape/narrow widths.
Root cause #1's trigger path (a bundle whose inner APKs won't parse) has no fixture either — it is
covered at the pure-function seam, not end to end.

## Review round 1 — `6e974ca`

Three reviewers, three lenses. Full write-up in the [review comment](https://github.com/trinadhthatakula/Thor/pull/276#issuecomment-5085484758). Headline: `isDowngrade` was doing double duty as both the verdict *and* the `pm install -d` permission flag, so the unknown-code gate in `ee8c40b` silently dropped `-d` and turned a previously-succeeding privileged install into a hard `INSTALL_FAILED_VERSION_DOWNGRADE`. Split into verdict vs. permission. Also: two tests that pinned nothing, `%d` → `%s` for Arabic digit shaping, an overflow guard, and the deletion of dead names-only `getVersionInfo()`.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
+++ b/app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
@@ -6,15 +6,17 @@ package com.valhalla.thor.domain.model
 /**
  * True iff [newVersionCode] is a *known* version code strictly older than [installedVersionCode].
  *
- * A non-positive [newVersionCode] means "unknown" — the analyzer could not read a usable version
- * code out of the picked file — and must never be reported as a downgrade. `0` loses every
- * comparison, so treating it as authoritative would flag an install over *any* installed app as a
- * downgrade, however new the file actually is. Same `0 = unknown` convention as
- * [CatalogEntry.versionCode].
+ * A null [newVersionCode] means "unknown" — the analyzer could not read a version code out of the
+ * picked file at all — and must never be reported as a downgrade, because an unknown would
+ * otherwise have to be represented by some number, and any number we picked would lose against
+ * something. Unknown is deliberately *not* encoded as `0`: `0` is a legal Android version code (the
+ * platform parser defaults a manifest with no `android:versionCode` to zero) and Android compares
+ * it numerically like any other, so installing a genuine `0` over a positive code IS a downgrade
+ * and has to be reported as one.
  *
  * Version *names* are deliberately not consulted: Android sequences updates by version code alone,
  * so a file may carry a newer-looking name (`1.2.5.1` over `1.2.4.7`) and still be a real
  * downgrade. Callers should surface both codes so that verdict is explainable to the user.
  */
-fun isVersionDowngrade(newVersionCode: Long, installedVersionCode: Long): Boolean =
-    newVersionCode > 0L && newVersionCode < installedVersionCode
+fun isVersionDowngrade(newVersionCode: Long?, installedVersionCode: Long): Boolean =
+    newVersionCode != null && newVersionCode < installedVersionCode
--- a/app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
+++ b/app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
@@ -1,0 +1,20 @@
+// SPDX-FileCopyrightText: 2025-2026 Trinadh Thatakula <github.com/trinadhthatakula/Thor>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package com.valhalla.thor.domain.model
+
+/**
+ * True iff [newVersionCode] is a *known* version code strictly older than [installedVersionCode].
+ *
+ * A non-positive [newVersionCode] means "unknown" — the analyzer could not read a usable version
+ * code out of the picked file — and must never be reported as a downgrade. `0` loses every
+ * comparison, so treating it as authoritative would flag an install over *any* installed app as a
+ * downgrade, however new the file actually is. Same `0 = unknown` convention as
+ * [CatalogEntry.versionCode].
+ *
+ * Version *names* are deliberately not consulted: Android sequences updates by version code alone,
+ * so a file may carry a newer-looking name (`1.2.5.1` over `1.2.4.7`) and still be a real
+ * downgrade. Callers should surface both codes so that verdict is explainable to the user.
+ */
+fun isVersionDowngrade(newVersionCode: Long, installedVersionCode: Long): Boolean =
+    newVersionCode > 0L && newVersionCode < installedVersionCode
--- a/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
+++ b/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
@@ -177,8 +177,9 @@ class AppAnalyzerImpl(private val context: Context) : AppAnalyzer {
         val versionName = manifest?.versionName?.takeIf { it.isNotBlank() }
             ?: apkmInfo?.versionName?.takeIf { it.isNotBlank() }
             ?: "Unknown"
-        // 0 when neither sidecar declares a usable code. Consumers must gate on that (see
-        // isVersionDowngrade) — a bundle whose sidecar carries a version *name* in version_code
+        // Null when neither sidecar declares a usable code. This is the ONLY path that can produce
+        // an unknown version code — there is no APK to parse here — and consumers must gate on it
+        // (see isVersionDowngrade): a bundle whose sidecar carries a version *name* in version_code
         // used to land here as 0 and read as a downgrade against every installed app.
         val versionCode = resolveSidecarVersionCode(manifest, apkmInfo)
         val icon = iconBytes?.let { BitmapFactory.decodeByteArray(it, 0, it.size) }
@@ -235,20 +236,23 @@ class AppAnalyzerImpl(private val context: Context) : AppAnalyzer {
      *
      * Keyed by packageName AND [versionCode]: Coil keys its File memory cache by the path only
      * (addLastModifiedToFileCacheKey defaults false), so a fixed per-package path would serve a
-     * STALE icon after a version bump — a distinct path per version busts that. Written to a unique
-     * temp file then atomically renamed, so two concurrent same-(pkg,version) analyses can never
-     * expose a partial/corrupt PNG to Coil (either the old complete file or the new one).
+     * STALE icon after a version bump — a distinct path per version busts that. A null (unknown)
+     * version code gets its own `_unknown` key rather than sharing the `_0` slot with a real APK
+     * that legitimately declares version code 0. Written to a unique temp file then atomically
+     * renamed, so two concurrent same-(pkg,version) analyses can never expose a partial/corrupt
+     * PNG to Coil (either the old complete file or the new one).
      *
      * Best-effort: any decode/IO failure yields a null path (never crashes parsing). This cache
      * file is intentionally NOT deleted in analyze()'s finally — the UI reads it after analyze()
      * returns; it lives in cacheDir so the OS can reclaim it.
      */
-    private fun persistIcon(bitmap: Bitmap?, packageName: String, versionCode: Long): String? {
+    private fun persistIcon(bitmap: Bitmap?, packageName: String, versionCode: Long?): String? {
         if (bitmap == null) return null
         return runCatching {
             val iconDir = File(context.cacheDir, "installer_icons").apply { mkdirs() }
-            val dest = File(iconDir, "${packageName}_$versionCode.png")
-            val tmp = File(iconDir, "${packageName}_$versionCode.${java.util.UUID.randomUUID()}.png.tmp")
+            val key = "${packageName}_${versionCode ?: "unknown"}"
+            val dest = File(iconDir, "$key.png")
+            val tmp = File(iconDir, "$key.${java.util.UUID.randomUUID()}.png.tmp")
             try {
                 FileOutputStream(tmp).use { out ->
                     bitmap.compress(Bitmap.CompressFormat.PNG, 100, out)
--- a/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
+++ b/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
@@ -177,7 +177,10 @@ class AppAnalyzerImpl(private val context: Context) : AppAnalyzer {
         val versionName = manifest?.versionName?.takeIf { it.isNotBlank() }
             ?: apkmInfo?.versionName?.takeIf { it.isNotBlank() }
             ?: "Unknown"
-        val versionCode = (manifest?.versionCode ?: apkmInfo?.versionCode)?.toLongOrNull() ?: 0L
+        // 0 when neither sidecar declares a usable code. Consumers must gate on that (see
+        // isVersionDowngrade) — a bundle whose sidecar carries a version *name* in version_code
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
```diff
@@ -0,0 +1,20 @@
...
+ * comparison, so treating it as authoritative would flag an install over *any* installed app as a
+ * downgrade, however new the file actually is. Same `0 = unknown` convention as
+ * [CatalogEntry.versionCode].
+ *
+ * Version *names* are deliberately not consulted: Android sequences updates by version code alone,
+ * so a file may carry a newer-looking name (`1.2.5.1` over `1.2.4.7`) and still be a real
+ * downgrade. Callers should surface both codes so that verdict is explainable to the user.
+ */
+fun isVersionDowngrade(newVersionCode: Long, installedVersionCode: Long): Boolean =
+    newVersionCode > 0L && newVersionCode < installedVersionCode
```
trinadhthatakula: **[P2] Preserve a real APK versionCode of 0**

`0` is ambiguous here: it means "unknown" only for the sidecar fallback, but `metadataFrom()` also passes through a parsed APK actual `longVersionCode`, which can be `0` (the AOSP APK parser defaults a missing `versionCode` to zero). Android downgrade checks still compare that value numerically, so replacing an installed positive version with such an APK is a downgrade. This gate now returns false, causing NORMAL mode to attempt an install that Package Manager rejects and privileged modes to omit `-d`/requestDowngrade.

Please represent unknown separately from the numeric value—for example:

```kotlin
fun isVersionDowngrade(
    newVersionCode: Long?,
    installedVersionCode: Long,
): Boolean = newVersionCode != null && newVersionCode < installedVersionCode
```

Then have `resolveSidecarVersionCode()` return `null` when unreadable while keeping parsed APK code `0` as a known value (and render it accordingly).

trinadhthatakula:
Review summary:

- The reported Omni transition (`1002007` → `1002001`) remains a downgrade, which is correct: the plain-APK path reads `longVersionCode`, and Android compares those codes rather than the displayed version names. The added UI makes that verdict explainable.
- The sidecar fallback now correctly ignores blank/non-numeric/non-positive fields and tries the next source, preventing unreadable bundle metadata from being treated as an authoritative downgrade.
- One actionable P2 finding was left inline: using `0` as the global unknown sentinel also hides a real parsed APK versionCode of zero. Unknown should be represented separately so Android-compatible downgrade behavior is retained.

Verification on commit `ee8c40b`:

- `./gradlew testFossDebugUnitTest` — passed
- `./gradlew lintFossDebug` — passed (only existing compiler warnings)
- `git diff --check` — passed

No on-device layout/install test was performed.

trinadhthatakula:
## Review round 1 + upstream confirmation (`6e974ca`)

Three reviewers over `ee8c40b` with different lenses (domain/contract propagation, Compose+i18n, adversarial). Two of them independently found the same regression, which is now fixed. Separately, the reporter located Omni's GitHub repo, which settles the diagnosis with real numbers and **corrects the reasoning in `ee8c40b`**.

### The Omni case: verdict CONFIRMED, my decode was WRONG

`ee8c40b` claimed Omni's scheme "drops the third name component" (1.2.4.7 → 1002007, 1.2.5.1 → 1002001), reverse-engineered from one online sample. That is not their scheme. From [REBEL-ROOT/omni-browser](https://github.com/REBEL-ROOT/omni-browser) `app/build.gradle.kts` at each tag:

| tag | versionName | versionCode |
|---|---|---|
| v1.2.4.6 | 1.2.4.6 | 26 |
| **v1.2.4.7** | 1.2.4.7 | **27** ← installed |
| **v1.2.5.1** | 1.2.5.1 | **25** ← picked |
| 1.2.6.0 | 1.2.6 | 25 |

Their versionCode is hand-maintained and **went backwards**. 25 over 27 is a genuine downgrade; the platform installer refuses it identically. So Thor's verdict was right, but for the ordinary reason, not an exotic encoding. Anyone on 1.2.4.7 is currently stuck — 1.2.6 also ships code 25. This is an upstream bug worth reporting to them.

(The 1002005-style code in old listings was real: their v1.2.3.4 shipped `versionCode = 1002004` from a scheme they abandoned one release later. That is where the sample came from.)

The test now asserts on 25 vs 27 and cites the tags instead of a reconstruction.

### Critical, fixed — the gate stripped the downgrade *permission*, not just the verdict

`isDowngrade` was doing two jobs: the UI verdict + NORMAL-mode veto, **and** the `canDowngrade` argument to `installPackage()`, which becomes the literal `pm install -d` (`RootSystemGateway.kt:403`, `ShizukuSystemGateway.kt:105`, `DhizukuSystemGateway.kt:94`, `InstallerRepositoryImpl.kt:391/399/439/447`) and `setRequestDowngrade(true)` (`InstallerRepositoryImpl.kt:489`).

Forcing it `false` for an unreadable version code therefore also dropped `-d`. Failure: app installed at code 200; user picks an `.xapk` whose sidecar carries no usable code and whose base APK does not parse (the Phase 4.5 path this PR exists to serve); root available. **Before this PR** the bogus `isDowngrade = true` passed `-d` and the install succeeded. **At `ee8c40b`** it does not, and the session commits to `INSTALL_FAILED_VERSION_DOWNGRADE` with no override anywhere in the UI. `-d` is permissive-only — a no-op when the install is not actually a downgrade — so the old permissiveness was free and the gate bought nothing on that axis.

Fixed by splitting verdict from permission: `isDowngrade` stays provable-only; `startInstallation()` computes `allowDowngrade = isDowngrade || (versionCodeUnknown && mode != InstallMode.NORMAL)`. Scoped to privileged modes on purpose — in NORMAL mode the flag needs a privilege we do not have, so widening it there buys nothing while putting a reflective `setRequestDowngrade` call on the unprivileged happy path for the first time, which is exactly the kind of change that should not ship without a device run.

### Two tests pinned nothing

Both now fail against a deliberately broken implementation:

- `versionCodeMajorIsRespected` used `0x1_00000001` vs `0x1_00000005` — those truncate to `1` and `5` and **keep their order**, so an Int-truncating implementation passed the test that exists to catch Int truncation. Now `5L` vs `1L shl 32`, which truncates to `5` vs `0` and inverts the verdict.
- `sidecarVersionCode_nonPositiveValuesAreUnknown` put `"0"` in the manifest slot, so the chain already answered `0` for the wrong reason and `.takeIf { it > 0L }` could be deleted without failing anything. Added lone-negative cases for each sidecar.

### Also fixed

- **Arabic digit shaping.** `install_downgrade_code_explainer` took `%d`, which `java.util.Formatter` localises (Arabic-Indic digits under `ar-EG`), while the same two numbers rendered by Kotlin interpolation ~20dp above stay ASCII — the same code, two different ways, on the screen whose entire purpose is comparing them. Now `%s` + explicit `toString()` in all five locales. A version code is an opaque identifier users check against a store listing, not a quantity.
- **Overflow.** The non-update version line grew ~10 characters and sat unweighted in a `Row`; added `weight(1f, fill = false)` so it cannot wrap mid-token at large font scales.
- **Deleted `InstallState.getVersionInfo()`.** Dead code (no call sites) that rendered the verdict from version NAMES only and tested `isUpdate` before `isDowngrade` — a downgrade is always also an update, so its downgrade branch was unreachable and it reported `"Update available: 1.2.5.1 (current: 1.2.4.7)"` for precisely this bug.

### Raised, not taken — filed for follow-up

- `AppInfoMapper.kt:61` does `packInfo.longVersionCode.toInt()`, and `AppInfo`/`AppEntity` store `versionCode: Int`. A real `versionCodeMajor` truncation bug (an app with `versionCodeMajor=1, versionCode=0x80000000` caches as `-2147483648` and sorts below everything under `SortBy.VERSION_CODE`) — but on the app-list path, not the installer, so out of scope here.
- `ExtensionBrowseViewModel.kt:115` open-codes the identical `> 0L` unknown-gate. Two call sites is arguably below the abstraction bar; left alone.
- No coverage of the wiring itself (analyzer → metadata → ViewModel → emitted state); both new helpers are tested only in isolation. Closing that needs a test dependency this module does not have (JUnit 4.13.2 only, no mockk/Robolectric/coroutines-test).
- Reviewer predicted a new `PluralsCandidate` lint warning from `%d` followed by a word. It did not materialise — `lintFossDebug` reports exactly the 5 pre-existing intentional `VectorPath` warnings. Moot now anyway with `%s`.
- TalkBack reads the new codes as bare cardinal numbers with no context. Worth a `contentDescription`, but I would want a device to check it on.

### Verification

93 unit tests / 0 failures / 13 classes. `assembleFossDebug` + `lintFossDebug` BUILD SUCCESSFUL; lint = 5 issues, all pre-existing intentional `VectorPath`.

**Still not verified:** nothing has been run on a device. The Compose changes (the weighted `Column` in the downgrade row, the new `weight(1f, fill = false)`, the code lines under each version name) are compile- and lint-verified only. The `-d` widening for unknown codes on privileged paths is reasoned, not observed.

app/src/main/java/com/valhalla/thor/domain/model/VersionCompare.kt
trinadhthatakula: Valid — fixed in 9f215a9, taking the `null` shape you proposed.

You're right that `metadataFrom()` is the hole: it passes a parsed archive's `longVersionCode` straight through, and the AOSP parser defaults a manifest with no `android:versionCode` to `0`, so a real, fully-parsed APK can legitimately arrive here as `0`. The `> 0L` gate then swallowed its verdict, and installing a genuine `0` over an installed positive code is a downgrade Android refuses like any other.

Unknown is now `null` end to end:

- `AppMetadata.versionCode: Long?`
- `resolveSidecarVersionCode(): Long?` (drops the terminal `?: 0L`)
- `isVersionDowngrade(newVersionCode: Long?, installedVersionCode: Long) = newVersionCode != null && newVersionCode < installedVersionCode`

One deliberate asymmetry: the per-source `sidecarVersionCode()` helper keeps its `takeIf { it > 0L }`, so a sidecar `0` is still unknown. Those fields are free text written by third-party repackers, where `0` is indistinguishable from an unset placeholder — and that filter is also what makes the source fallthrough work, i.e. it's what stops a junk `manifest.json` value from shadowing a good `info.json` one. There's now a test pinning that (`sidecarVersionCode_zeroDoesNotShadowAGoodSiblingValue`). The sidecar-only fallback is the one path with no APK to parse, so it stays the only producer of `null`.

Two smaller things that fell out:

- the installer icon cache keys an unknown code as `_unknown` rather than sharing the `_0` slot with a real version-code-0 APK;
- the downgrade explainer null-checks both codes, so it can't render the word `null` even if the invariant were ever broken.

On the stated consequence, half of it was already closed by 6e974ca earlier in this PR: privileged modes no longer omit `-d` for an unreadable code (`versionCodeUnknown` widens the *permission* while `isDowngrade` stays the *verdict*). What was genuinely still broken — and is what this commit fixes — is a real versionCode-0 APK losing both its downgrade warning and its up-front NORMAL-mode veto message, so the user hit an opaque `INSTALL_FAILED_VERSION_DOWNGRADE` instead of an explanation.

Test changes: `negativeVersionCodeIsTreatedAsUnknown` is dropped rather than migrated — negatives can't reach the predicate any more (the sidecar filter turns them into `null`; a parsed archive can't produce one), so the assertion was false by construction. Replaced by `realVersionCodeZeroIsStillCompared`, which pins exactly the case you flagged. 95 unit tests pass, `lintFossDebug` reports the same 5 pre-existing `VectorPath` issues and nothing new.

PR #277, opened 2026-07-26
fix(apps): stop truncating longVersionCode to Int

trinadhthatakula:
## The bug

`mapToAppInfo` read the platform's version code as `packInfo.longVersionCode.toInt()`.

`longVersionCode` is `(versionCodeMajor shl 32) or versionCode`, so the cast:

- **discards `versionCodeMajor` entirely** — an app declaring `versionCodeMajor = 1, versionCode = 0` arrives as `0`;
- **reinterprets the low 32 bits as signed** — a code above `Int.MAX_VALUE` (2 147 483 647) wraps negative.

Found while reviewing #276 (the Portable Installer downgrade fix). It's a *different* bug on a *different* code path — the installer parses picked files via `AppAnalyzerImpl`, which was already reading `longVersionCode` correctly; this one is the installed-app mapper — so it gets its own branch.

## Why it matters

The truncated value propagates everywhere `AppInfo.versionCode` is read:

| Site | Consequence |
|---|---|
| `AppSorting.kt:16` — `SortBy.VERSION_CODE` | a `versionCodeMajor` app sorts **below** every ordinary app instead of above |
| `AppInfoDetailsScreen.kt:862` | the wrong number is shown to the user as the app's version code |
| `ApksMetadataGenerator` | written into the `version_code` of the `.apks`/XAPK metadata Thor generates on export — read back by other installers **and by Thor's own bundle analyzer** to decide update vs. downgrade |
| Room `AppEntity` | persisted |

Honest severity: **low**. `versionCodeMajor` is rare and Play caps `versionCode` at 2 100 000 000 (< 2³¹), so most devices have nothing wrong. But where it does bite, the export path turns an app into a phantom downgrade on reinstall — the same class of confusion #276 exists to explain.

## The fix

`Int` → `Long` end to end: `AppInfo.versionCode`, `AppEntity.versionCode`, `ApksMetadata.versionCode`, and the mapper drops its `.toInt()`. `AppSorting` and the details screen need no change (a `Comparator` and a string template).

### No Room migration, and that's deliberate

Room maps **both** Kotlin `Int` and `Long` to SQLite affinity `INTEGER`, so the exported schema is byte-identical and the `identityHash` is unchanged. Verified rather than assumed — re-ran KSP with `--rerun-tasks` and diffed `app/schemas/` (empty; hash still `20e4df41d4d04b8b7c74ab54244fc704`). Existing databases open normally, no version bump.

Rows already cached with a truncated value keep it until that package's `lastUpdateTime` changes and `AppRepositoryImpl` re-maps it. That self-heals on the app's next update, and only ever affected the rare apps above — forcing a full re-scan (or a `DELETE FROM apps`, which would also throw away the lazily-computed `installSize` cache) on every user would cost considerably more than it fixes. Documented in a comment on the field.

### Drive-by

Dropped the `@Suppress("DEPRECATION")` on the `AppInfo(...)` construction. It was a leftover from the deprecated `Int` `packInfo.versionCode` this mapper used to read; nothing in the expression is deprecated any more (verified by compiling without it), and leaving it there would silently swallow the next real deprecation on one of those calls.

## Tests

- `AppSortingTest` — two new ordering cases, with operands chosen so a *truncating* implementation inverts the result rather than merely agreeing by luck (`1L shl 32` truncates to `0`; `2_147_483_648L` wraps negative).
- `ApksMetadataGeneratorTest` (new) — pins the generated **wire format**, i.e. that `version_code` carries the full value as an unquoted JSON number (`.apks` meta) and as a quoted string (XAPK manifest). This is the assertion the Kotlin types alone can't give you.

`mapToAppInfo` itself is not unit-tested: it needs a real `PackageInfo`, and this project has JUnit 4 only — no Robolectric/mockk.

## Verification

- 83 unit tests pass, 0 failures
- `lintFossDebug`: 5 issues, all pre-existing `VectorPath` — nothing new
- `assembleFossRelease` + `assembleStoreRelease` green

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
+++ b/app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
@@ -14,11 +14,10 @@ data class AppEntity(
     val versionName: String?,
     // Int -> Long. No schema version bump or Migration is needed: Room maps both Kotlin Int and
     // Long to SQLite affinity INTEGER, so the exported schema and its identityHash are unchanged
-    // (see app/schemas/.../5.json). Rows already cached with a truncated value keep it until the
-    // package's lastUpdateTime changes and AppRepositoryImpl re-maps it — acceptable, because the
-    // only values that were ever wrong belong to apps declaring versionCodeMajor or a code above
-    // Int.MAX, which are vanishingly rare, and forcing a full re-scan on every user to correct
-    // them would cost far more than it fixes.
+    // (see app/schemas/.../5.json) and existing rows survive as-is. Rows written before the change
+    // hold a truncated value; AppRepositoryImpl's scan compares this column against
+    // PackageInfo.longVersionCode and re-maps on a mismatch, so they are repaired on the first
+    // scan without discarding the cache or pushing every package through the mapper.
     val versionCode: Long,
     val minSdk: Int,
     val targetSdk: Int,
--- a/app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
+++ b/app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
@@ -12,7 +12,14 @@ data class AppEntity(
     @PrimaryKey val packageName: String,
     val appName: String?,
     val versionName: String?,
-    val versionCode: Int,
+    // Int -> Long. No schema version bump or Migration is needed: Room maps both Kotlin Int and
+    // Long to SQLite affinity INTEGER, so the exported schema and its identityHash are unchanged
+    // (see app/schemas/.../5.json). Rows already cached with a truncated value keep it until the
+    // package's lastUpdateTime changes and AppRepositoryImpl re-maps it — acceptable, because the
+    // only values that were ever wrong belong to apps declaring versionCodeMajor or a code above
+    // Int.MAX, which are vanishingly rare, and forcing a full re-scan on every user to correct
+    // them would cost far more than it fixes.
+    val versionCode: Long,
     val minSdk: Int,
     val targetSdk: Int,
     val isSystem: Boolean,
--- a/app/src/main/java/com/valhalla/thor/data/repository/AppInfoMapper.kt
+++ b/app/src/main/java/com/valhalla/thor/data/repository/AppInfoMapper.kt
@@ -53,12 +53,16 @@ fun mapToAppInfo(
         ""
     }
 
-    @Suppress("DEPRECATION")
+    // No @Suppress("DEPRECATION") here: nothing in this expression is deprecated any more. It was
+    // a leftover from the deprecated Int `packInfo.versionCode` this mapper used to read, and
+    // keeping it would silently swallow the next real deprecation on one of these calls.
     return AppInfo(
         appName = appInfo.loadLabel(pm).toString(),
         packageName = packInfo.packageName,
         versionName = packInfo.versionName,
-        versionCode = packInfo.longVersionCode.toInt(),
+        // No .toInt(): longVersionCode is (versionCodeMajor shl 32) or versionCode, so truncating
+        // discarded versionCodeMajor entirely and reinterpreted the low 32 bits as signed.
+        versionCode = packInfo.longVersionCode,
         minSdk = appInfo.minSdkVersion,
         targetSdk = appInfo.targetSdkVersion,
         isSystem = (appInfo.flags and ApplicationInfo.FLAG_SYSTEM) != 0,
--- a/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt
+++ b/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt
@@ -112,6 +112,14 @@ class AppRepositoryImpl(
 
                         if (!forceRefresh &&
                             cachedEntry != null &&
+                            // Also a cache miss when the cached code disagrees with the platform's.
+                            // Rows written before versionCode widened to Long hold a truncated
+                            // value, and lastUpdateTime alone never invalidates them: installing
+                            // *Thor* does not change the *target* package's timestamp. Comparing
+                            // the value we already have in hand repairs those rows on the first
+                            // scan, for one Long compare per package and a re-map of only the
+                            // packages that are actually wrong.
+                            cachedEntry.versionCode == packInfo.longVersionCode &&
                             cachedEntry.lastUpdateTime == packInfo.lastUpdateTime &&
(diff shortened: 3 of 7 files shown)


app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
```diff
@@ -12,7 +12,14 @@ data class AppEntity(
...
     val versionName: String?,
-    val versionCode: Int,
+    // Int -> Long. No schema version bump or Migration is needed: Room maps both Kotlin Int and
+    // Long to SQLite affinity INTEGER, so the exported schema and its identityHash are unchanged
+    // (see app/schemas/.../5.json). Rows already cached with a truncated value keep it until the
+    // package's lastUpdateTime changes and AppRepositoryImpl re-maps it — acceptable, because the
+    // only values that were ever wrong belong to apps declaring versionCodeMajor or a code above
+    // Int.MAX, which are vanishingly rare, and forcing a full re-scan on every user to correct
+    // them would cost far more than it fixes.
+    val versionCode: Long,
```
trinadhthatakula, line 21: **[P2] Self-heal version codes already truncated in the cache**

Leaving existing rows untouched means the bug remains for the users who actually hit it. On every startup, `AppRepositoryImpl` first emits the cached row and then reuses it whenever `lastUpdateTime`, enabled, and suspended state match; installing this Thor update does not change the target package timestamp. Consequently, sorting and list-driven export/share paths can keep using the truncated code indefinitely—not merely until this app rescans.

A full cache deletion is unnecessary. The scan already has both values, so make the fast path validate the code and remap only affected packages:

```kotlin
if (!forceRefresh &&
    cachedEntry != null &&
    cachedEntry.versionCode == packInfo.longVersionCode &&
    cachedEntry.lastUpdateTime == packInfo.lastUpdateTime &&
    cachedEntry.enabled == isEnabled &&
    cachedEntry.isSuspended == isSuspended
) {
    // reuse cache
} else {
    // map and persist the current PackageInfo
}
```

`syncCache()` already preserves `installSize`, so this repairs affected rows on the first scan without discarding that cache or forcing every package through the mapper.

trinadhthatakula:
Review summary:

- The `Int` → `Long` propagation is correct for newly mapped packages: `PackageInfo.longVersionCode` now survives the domain model, Room read/write, ordering, details UI, and both generated metadata formats.
- The SQLite schema remains compatible because both Room types use `INTEGER`; no database version bump is required.
- One actionable P2 finding was left inline: existing affected rows are never remapped merely by installing this fix, because the cache fast path only checks `lastUpdateTime` and app state. Comparing `cachedEntry.versionCode` with `packInfo.longVersionCode` repairs only mismatches on the first scan and preserves `installSize` through the existing `syncCache()` logic.

Verification on commit `7f1e5e18`:

- `./gradlew testFossDebugUnitTest lintFossDebug` — passed
- `git diff --check` — passed
- No unexpected schema-file changes were produced

GitHub CI and CodeQL were still running when this review was posted.

app/src/main/java/com/valhalla/thor/data/source/local/room/AppEntity.kt
trinadhthatakula: Agreed, and thanks — the comment I wrote was wrong on the part that matters. Fixed in `feb2bc90` with your patch as-is.

You're right that `lastUpdateTime` is the *target package's* timestamp, so a Thor update never invalidates the row. What I'd add is that the guard has two more clauses, `enabled` and `isSuspended`, and freezing or suspending an app — the thing Thor exists to do — does flip those and remap the row. So the practical window is narrower than "indefinitely" for apps a user actually acts on. But that's not a repair mechanism, it's a coincidence, and it does nothing for the apps nobody touches. Validating the value we already have in hand is the right answer.

I'd considered `DELETE FROM apps` and forcing `lastUpdateTime = -1` on upgrade; both throw away the whole cache to correct a handful of rows. Yours costs one `Long` compare per package per scan and re-maps only the packages that are genuinely wrong, and as you note `syncCache()` carries `installSize` through the REPLACE, so nothing is lost.

I also rewrote the `AppEntity.versionCode` comment, since its stated rationale was exactly the reasoning you refuted.

Verified: 99 unit tests / 0 failures, `lintFossDebug` unchanged at 5 pre-existing `VectorPath` issues, `assembleFossRelease` and `assembleStoreRelease` both green.

trinadhthatakula:
Final state after code review. Two rounds of review on this branch: an external reviewer's `[P2]` on `AppEntity.kt`, and my own multi-lens adversarial pass.

**Change made in response to review** — `feb2bc90`:

The original commit fixed every version code Thor derives from now on, but left rows already written with a truncated one. The reviewer correctly pointed out that my comment's claim ("self-heals on the app's next update") was wrong in the way that matters: `lastUpdateTime` belongs to the *target* package, and installing a Thor update never changes it. So for an app the user never touches, the stale code could feed sorting and the list-driven export/share paths indefinitely.

Applied their fix verbatim — one clause on the rescan's fast path:

```kotlin
cachedEntry.versionCode == packInfo.longVersionCode &&
```

A mismatch falls through to the existing `else` branch, which re-maps and persists the corrected entity. One `Long` compare per package per scan, only genuinely-wrong packages re-mapped, and `syncCache()` carries `installSize` through the REPLACE — so the cache is repaired rather than discarded. Strictly better than the alternatives I'd weighed (`DELETE FROM apps`, or forcing `lastUpdateTime = -1` on upgrade), both of which throw away the whole cache to correct a handful of rows.

Also rewrote the `AppEntity.versionCode` comment, since its stated rationale was exactly what the review refuted.

**Adversarial review pass:** 6 lenses (correctness, tests, Room persistence, UI/behaviour, scope/risk, wire format), every candidate finding put to 3 independent skeptics, plus a completeness critic. **0 confirmed findings.** The most-reported candidate was the same stale-cache issue above — now fixed. One useful correction came out of it: the fast path also keys on `enabled` and `isSuspended`, so freezing or suspending an app already remapped its row. That narrows the practical window but isn't a repair mechanism, and does nothing for apps nobody touches, so the fix stands.

**Verification** (on `feb2bc90`):
- 99 unit tests, 0 failures
- `lintFossDebug`: 5 issues, all pre-existing `VectorPath` — unchanged from `dev`
- `assembleFossRelease` + `assembleStoreRelease`: BUILD SUCCESSFUL
- Room schema: `identityHash` unchanged (`20e4df41d4d04b8b7c74ab54244fc704`); `Int`->`Long` is affinity-identical in SQLite, so no version bump and no `Migration` — verified by re-running `kspFossDebugKotlin --rerun-tasks` and diffing `app/schemas/`

PR #278, opened 2026-07-27
fix(appList): don't charge pull-to-refresh the transition settle delay

trinadhthatakula:
## Problem

`AppListViewModel.loadApps()` unconditionally waited 800 ms before it began collecting
`getInstalledAppsUseCase()`. Because that flow is **cold**, the delay sat *in front of* the scan
rather than overlapping it — the Room read and `pm.getInstalledPackages()` did not start until it
elapsed. Every caller paid it, including pull-to-refresh, which has no entry animation to protect.

## Changes

**1. Only the navigation-entry paths defer** (`2a158305`)

`loadApps(deferForTransition: Boolean = false)`. `init` and `AppListScreen`'s entry
`LaunchedEffect` opt in; `PullToRefreshBox.onRefresh` does not, so a deliberate gesture now starts
the scan immediately.

**2. The delay scales with the Animation Intensity setting** (`6fc3b768`)

`settleDelayFor(intensity)` → LOW `0` · MEDIUM `400 ms` · HIGH `800 ms`.

This tracks real work rather than a guess. `MainScreen` gives **LOW** a `snap()` spec — literally no
animation runs on entry, so there is nothing to settle. **HIGH** additionally enables
`SharedTransitionLayout` shared-element transitions, which animate bounds across destinations and
are the most sensitive to a competing scan, hence the longest head start.

Extracted to its own file so it is unit-testable, following the `HomeActions.kt` / `AppSorting.kt`
precedent. The new DataStore read is guarded with `catch { emit(UserPreferences()) }` —
`dataStore.data` throws `IOException` on a failed read and `PreferenceRepositoryImpl` has no guard
of its own, so a preference read failure must not take down the app list. (`Flow.catch` stays
transparent to cancellation, unlike `runCatching`.)

**3. Keep the refresh indicator readable** (`abe174ca`)

Dropping the delay from the refresh path left the gesture with almost no acknowledgment.
`isLoading` cannot carry it: `AppRepositoryImpl.getAllApps()` sends the Room cache *before* it
triggers the package rescan, and `priv.isReady` has already latched true, so `isLoading` clears one
DAO read later and the indicator flashes out while the real scan is still running.

A dedicated `isManualRefreshing` flag holds the **indicator** for 600 ms while the **scan** starts
at once — so this is strictly better than the old behaviour on both axes, not a restoration of it.
The flag is raised synchronously outside the coroutine and lowered only by a timer that ran to
completion, so a second refresh cannot have its indicator hidden by the teardown of the one it
replaced (`viewModelScope` is `Main.immediate`, which makes that ordering reachable).

It also restores re-entrancy the delay had been providing incidentally: `PullToRefreshBox` ignores
pull gestures while `isRefreshing` is true, so repeated pulls can no longer stack overlapping
package scans.

## Notes

- No behaviour change for cold start or navigation entry at MEDIUM/HIGH; LOW now starts scanning
  immediately, which is correct because it has no animation.
- `versionCode` deliberately not bumped — that belongs in a `chore(release)` commit.

## Verification

- `assembleFossDebug` ✅ · `assembleFossRelease` ✅ · `assembleStoreRelease` ✅
- `testFossDebugUnitTest` ✅ — 105 tests, 0 failures (99 → 105, +6 new)
- `lintFossDebug` ✅ — 5 results, all pre-existing `VectorPath`, none in `appList`


trinadhthatakula:
## Code Review Summary: PR #278

### Key Changes
- **Pull-to-refresh acceleration**: Removes hardcoded 800 ms transition delay from manual refreshes (`loadApps()`), starting the package scan immediately upon pull.
- **Dynamic entry settle delay**: Maps entry animation delay dynamically to `AnimationIntensity` setting (`LOW`: 0 ms, `MEDIUM`: 400 ms, `HIGH`: 800 ms).
- **Indicator legibility fix**: Adds `isManualRefreshing` state floor (600 ms) so the refresh spinner remains visible while background package scanning completes.
- **Unit testing**: Adds `TransitionSettleDelayTest.kt` for delay mappings, intensity hierarchy, and enum safety.

### Suggestions
1. **Redundant launch trigger**: `AppListViewModel.init` launches `loadApps(deferForTransition = true)`. Initial composition in `AppListScreen` triggers `LaunchedEffect(Unit)` when `state.isLoading == true`, cancelling the `init` job and restarting the delay timer. Consider removing or tuning the `LaunchedEffect` condition.
2. **Formatting**: Add a trailing newline at EOF in `AppListViewModel.kt`.

### Verdict
**MERGE** (Model used: Gemini 3.6 Flash)

trinadhthatakula:
## Code Review

### Finding

1. **[P2] Manual refresh can still overlap an active package scan.**

   The indicator becomes false after a fixed 600 ms, regardless of scan completion ([`AppListViewModel.kt:231-241`](https://github.com/trinadhthatakula/Thor/blob/abe174ca378eb6f1d42c69e9ad9ca33704f027af/app/src/main/java/com/valhalla/thor/presentation/appList/AppListViewModel.kt#L231-L241)). After that, another pull cancels the old job without waiting and immediately starts a new one ([`AppListViewModel.kt:150-157`](https://github.com/trinadhthatakula/Thor/blob/abe174ca378eb6f1d42c69e9ad9ca33704f027af/app/src/main/java/com/valhalla/thor/presentation/appList/AppListViewModel.kt#L150-L157)). The synchronous `getInstalledPackages()` call is not cancellation-aware ([`AppRepositoryImpl.kt:84-90`](https://github.com/trinadhthatakula/Thor/blob/abe174ca378eb6f1d42c69e9ad9ca33704f027af/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt#L84-L90)), so the old and new scans can run concurrently.

   **Recommendation:** serialize scans with a `Mutex`, use `cancelAndJoin()` before restarting, or expose explicit refresh-completion state from the repository. Keep 600 ms as the minimum display duration, but do not use it as the re-entrancy boundary.

### PR summary

- Manual refresh begins immediately instead of paying the 800 ms entry-transition delay.
- Navigation-entry delay follows animation intensity: LOW `0`, MEDIUM `400 ms`, HIGH `800 ms`.
- Adds a 600 ms minimum refresh-indicator duration.
- Adds six unit tests for delay mappings and invariants.
- Scope: 4 files, `+173/-7`.

### Suggestions

- Add ViewModel behavior tests covering rapid repeated refreshes, cancellation, indicator timing, and DataStore failure. The current tests only validate constants.
- The `init` load and screen `LaunchedEffect` both initiate loading, causing the first job and delay timer to restart. This is pre-existing but worth consolidating.
- Add the missing trailing newline in `AppListViewModel.kt`.
- CodeRabbit's green status does not represent a completed review; its comment says review was skipped because `dev` is not the default branch.

### Verification

- All GitHub checks currently pass.
- Targeted `TransitionSettleDelayTest` completed successfully locally.

**Model used:** OpenAI Codex, GPT-5

**Verdict:** **Do not merge as-is — request changes** for scan serialization/re-entrancy protection.

trinadhthatakula:
Thanks — the P2 was correct and is fixed at the root in `4c164d6e`, with the review-driven corrections to my own code in `cdee37f3`. Detail below.

## P2 — scan overlap (blocking) → fixed

Confirmed, and worth stating plainly: **this branch made it materially worse, and my own pre-push review round got this wrong.** My reviewers refuted scan-overlap as "pre-existing" and stopped there. The class *is* pre-existing, but the degree changed:

| | interval between consecutive scan starts |
|---|---|
| before this branch | ~1600 ms (800 ms delay + 800 ms delay) |
| after `abe174ca` | ~600 ms (the indicator floor) |

So a scan that previously finished inside the gap now overlaps. That is a regression attributable to these commits, not just inherited behaviour.

**Fixed the cause rather than the timing**, in `AppRepositoryImpl.getAllApps()`:

1. `ensureActive()` at the top of the per-package loop. The rescan is a plain blocking loop over `pm.getInstalledPackages()` with no suspension point, so `worker.cancel()` could never interrupt it. One volatile read per package makes teardown prompt.
2. **`catch (e: Exception)` was swallowing `CancellationException`.** Found this while implementing (1) — Kotlin's `CancellationException` extends `Exception`, so the cancellation would have been caught, printed in debug, and followed by another trip round the trigger channel, defeating both `ensureActive()` and the `awaitClose` teardown. Now rethrown ahead of the broad catch.

On the suggested alternatives: `cancelAndJoin()` would not have worked here — `awaitClose` calls `worker.cancel()` without joining it, so `appsJob` completes while the blocking loop runs on. A `Mutex` would serialise scans but leave the old one running to completion before the new one starts, which is slower than just cancelling it. Making cancellation actually work removes the overlap on **every** path, not only refresh.

Deliberately kept 600 ms as display-duration only, per your recommendation — it is no longer load-bearing for re-entrancy.

## Also fixed — three defects in code this branch added

These came from an internal review pass over `6fc3b768` and `abe174ca`; two were proven by mutation rather than argument.

- **`settleDelayFor`'s KDoc was factually false.** It justified 0/400/800 by which animation runs at each intensity. But `MainScreen` takes `appListViewModel` as an *eager default parameter*, so the delay runs during MainScreen's **first composition at app launch** — `HomeActivity` composes it from a plain `when` with no transition, and neither `NavDisplay`'s `AnimatedContent` nor the bottom bar's `AnimatedVisibility` animates on first composition (`initialState == targetState`). MEDIUM and HIGH also share *identical* entry specs, and the shared-element transitions cited for HIGH fire on the list→detail path, which never calls this function. Applied consistently, the old rationale yields 0/0/0. Rewritten to document what the delay really protects (app-launch frames) and that `AnimationIntensity` is a proxy for how much the user values smoothness over immediacy.
- **The 600 ms floor was not pinned.** `refreshIndicatorFloorIsActuallyVisible` and `refreshIndicatorFloorDoesNotScaleWithIntensity` both reduced to `REFRESH_INDICATOR_MIN_VISIBLE > Duration.ZERO`, because `settleDelayFor(LOW)` *is* `Duration.ZERO`. Verified by mutation: setting it to `1.milliseconds` — exactly the indicator flash the constant exists to prevent — kept all six tests green. Replaced with an exact pin matching how MEDIUM/HIGH are asserted; the same mutation now **fails**. The duplicate is deleted (its named property was unfalsifiable anyway, since the constant takes no intensity parameter).
- **`everyIntensityIsMappedAndNonNegative`'s comment was false.** It claimed to guard a future enum entry; `settleDelayFor` is an expression-body `when` with no `else`, so an unmapped entry is a compile error and the test never runs. Comment corrected to what it actually catches — a negative literal, which `delay()` would silently treat as zero.

## Other review points

- **Trailing newline** in `AppListViewModel.kt` — added.
- **`init` + `LaunchedEffect` double-load** — agreed, and confirmed it does fire (`isLoading` defaults to `true`, so the guard passes against the `stateIn` initial value). Cost is one restarted timer, not two delays, since the calls land milliseconds apart. Pre-existing and unchanged by this branch, so left out of scope; happy to consolidate separately.
- **ViewModel behaviour tests** (rapid refresh, cancellation, indicator timing, DataStore failure) — agreed the current tests only validate constants, and `holdRefreshIndicator`'s ordering is the subtle part. It needs `kotlinx-coroutines-test` + turbine, neither currently in the project (`app/build.gradle.kts` has plain junit only), so it is a dependency decision rather than a test I can just add here.
- **CodeRabbit** — correct, its green status means review was *skipped* because `dev` is not the default branch. Noted, not treated as a signal.

## Verification

`assembleFossDebug` · `assembleFossRelease` · `assembleStoreRelease` all green. `testFossDebugUnitTest`: **104 tests, 0 failures** (105 → 104; the tautological duplicate removed). `lintFossDebug`: 5 results, all pre-existing `VectorPath`, none in `appList` or `AppRepositoryImpl`.

Still open for the maintainer, and intentionally not changed here: with the rationale corrected, `LOW = 0` means a full package scan now starts concurrently with the splash exit and first frames, where every user previously had an 800 ms head start. The 0/400/800 mapping was an explicit product decision, so I have left it as-is rather than quietly reverting it.

trinadhthatakula:
## Verification results

### On-device — passed ✅

Maintainer-verified on hardware across **all three Animation Intensity settings**:

| Intensity | Settle delay | Result |
|---|---|---|
| LOW | 0 ms | smooth, no jank |
| MEDIUM | 400 ms | smooth, no jank |
| HIGH | 800 ms | smooth, no jank |

Screen entry and pull-to-refresh both behave correctly at every intensity. The concern raised during review — that `LOW = 0` starts a full package scan concurrently with app-launch first frames and might need a floor — **did not reproduce on device**. The 0/400/800 mapping stands as-is.

### Android lint — clean ✅

```
./gradlew lintFossRelease lintStoreRelease lintFossDebug   → exit 0
```

All three variants produce identical output: **5 results, all `VectorPath`**, in `drawable/dhizuku.xml`, `drawable/thor_drawn_foreground.xml`, `drawable/thor_mono.xml`. These are pre-existing and intentional — they are the residue of the lossless-only VectorPath compaction in #269, where further savings would have been lossy.

**Zero lint results in `presentation/appList/` or `data/repository/AppRepositoryImpl.kt`.**

### Unit tests — clean ✅

```
./gradlew :app:testFossDebugUnitTest --rerun-tasks   → BUILD SUCCESSFUL
```

**15 classes, 104 tests, 0 failures, 0 errors.** `--rerun-tasks` was required: the first invocation reported `testFossDebugUnitTest UP-TO-DATE` and had not actually executed anything, so the initial green was meaningless.

`TransitionSettleDelayTest` → 5 tests, all passing.

### Kotlin compiler warnings — clean for this branch ✅

Forced recompile of both release variants (`compileFossReleaseKotlin`, `compileStoreReleaseKotlin`, `--rerun-tasks`) surfaces **3 warnings, all anchored in `data/manager/ExtensionManager.kt`** (lines 78, 218, 243) — two "Elvis operator always returns the left operand" plus the Koin-plugin version notice. All pre-existing, none touched by this PR.

**Zero warnings in the four files this PR changes.**

---

### Qodana — deliberately *not* claimed as a passing check ⚠️

Recording this so nobody reads a green Qodana badge as Kotlin coverage on this PR.

`qodana-jvm-community` (the only Docker-free linter available here) exits 0 and reports 6 problems — but it **never analyzes a single line of Kotlin in `:app`**. From its own results directory:

- `projectStructure/Modules.json` — `Thor.app` and `Thor.bypass` both have `"sourceFolders": []`. Only `Thor.vm-runtime` gets `.main`/`.test` source-set submodules, because it is plain `java-library` and syncs *without* AGP.
- `projectStructure/Gradle.json` — `"libraries": []`.
- `log/idea.log` — `PerProjectIndexingQueue - Finished for [Thor]. No files to index with loading content.`
- `qodana.sarif.json` — only two artifact URIs in the entire report: `app/build.gradle.kts` and `settings.gradle.kts`. **No `.kt` file anywhere.** Gradle Kotlin DSL is analyzed through script support, independently of module source roots — which is exactly why the 6 `UnstableApiUsage` findings appear and nothing else does.
- `Project analysis stage completed in 3s` — for a ~186-file Compose module.

Root cause: the community linter ships no Android Gradle plugin, so AGP sync produces no source roots for the Android modules. The Android linter is Docker-only and Docker is unavailable here:

```
$ qodana scan --linter qodana-jvm-android --within-docker=false
✗  Native mode for linter qodana-jvm-android is not supported
```

Two notes for anyone repeating this:

1. A Qodana community run **writes its failed-sync defaults back into the repo** — `.idea/misc.xml` `JDK_21` → `JDK_1_8`, `.idea/compiler.xml` bytecode `21` → `1.8`, and it drops the Kotlin api/languageVersion `2.4` block from `.idea/kotlinc.xml`. Reverted here; worth a `git checkout -- .idea/` after any such run.
2. This also narrows the earlier 2026-07-20 sweep retroactively: that run's 72 problems were all XML / assets / build scripts / `:vm-runtime` — precisely the file classes that don't require `:app` source roots. So it, too, found nothing in `:app` Kotlin, and its "clean per Qodana" conclusion was never supported for Kotlin.

**Net:** Android lint covers the Android surface and is clean; unit tests and compiler warnings cover what they can. The Kotlin/IDE-inspection layer specifically was **not** exercised on this PR — it needs Docker, or `inspect.sh` on a stable (non-Canary) Android Studio. Flagging the gap rather than reporting a green run that inspected two build files.

PR #284, opened 2026-07-28
fix(tile): derive Freezer QS tile state from real per-app freeze state

trinadhthatakula:
## Problem

The Freezer QS tile reported state it had never actually checked.

`FreezerTileService` derived its tile state from **the size of the freezer watchlist**, not from
whether any listed app was currently unfrozen. So once you froze everything, the watchlist was still
non-empty and the tile stayed lit at `STATE_ACTIVE` — "3 apps · tap to freeze" — for apps that were
already frozen. Tapping it ran a batch that did nothing. This is the reported bug, reproduced on
device with root.

Two related defects came out of the assessment:

**Results were reported through a Toast that can never appear.** A `TileService` runs while the
shade is open, and `NotificationManagerService.checkCanEnqueueToast` drops toasts from a package
that is not foreground and holds no notification permission. Even when one is enqueued,
`TYPE_TOAST` sits at window layer 7 and `TYPE_NOTIFICATION_SHADE` at 17 — it draws *underneath* the
shade. Every bulk-freeze result the tile has ever "reported" was invisible.

**The launcher shortcuts had a second, divergent implementation.**
`FreezerShortcutManager.runBulk` filtered candidates its own way and reported nothing at all.

## Changes

The rework replaces the guesswork with real per-app state, and gives the tile and the shortcuts one
shared execution path.

**1. Tile state comes from the actual freeze state of each app** (`33a0a870`, `4c8db239`,
`15365111`)

`AppFreezeStateReader` reads the real per-app state; `freezableCandidates()` (pure, in
`domain/model/FreezeState.kt`) filters it. The tile's `STATE_ACTIVE` now means *at least one listed
app is currently freezable*, which is the thing the label claims.

**2. A pure tile state machine** (`172bf778`, `6379d685`)

`TileVisual.kt` holds `tileVisualFor(...)` and `tileStateFor(visual)` — no Android types, fully unit
tested (13 tests).

The mapping is load-bearing and non-obvious: `Tile.STATE_UNAVAILABLE` for no privilege,
`STATE_INACTIVE` for nothing to freeze, `STATE_ACTIVE` for ≥1 freezable. **`CHECKING` must map to
`STATE_INACTIVE`, not `STATE_UNAVAILABLE`** — `CustomTile.handleClick()` early-returns on
`STATE_UNAVAILABLE`, so an unavailable tile never receives `onClick` at all. Mapping the transient
checking state to UNAVAILABLE would swallow the user's first tap. `tileStateFor` exists so a
regression there fails a test rather than shipping.

(`Tile.STATE_*` are `public static final int` 0/1/2 — JLS constant variables, inlined by kotlinc —
so these assertions load no `android.jar` class and need no new test dependency.)

**3. `BulkFreezeRunner` — one owner for bulk freeze/unfreeze** (`bb7c66d0`, `3ccd6d3d`, `f2e2a40b`,
`0824f142`, `15d0a59d`, `e2927727`, `5cefb34d`, `5f1bc052`)

Both the tile and the launcher shortcuts now go through it, so they share the candidate filter, the
concurrency policy and the reporting path. It:

- awaits `privilegeManager.state.first { it.isReady }` rather than reading an unready snapshot —
  `PrivilegeState` defaults to `active = NONE, isReady = false`, so a cold-start caller that
  short-circuited on `hasAnyPrivilege` would silently no-op;
- serializes conflicting ops (cancel-and-replace) with a bounded `NonCancellable` join, so an
  abandoned FREEZE worker cannot re-disable a package a replacement UNFREEZE batch just enabled;
- hoists its `Semaphore(5)` to an instance field so worker concurrency is bounded across
  generations, not just within one run;
- tags every `BulkResult` with its `op`, so an unfreeze can no longer be reported as "Froze 5 apps".

**4. Results are reported where they can actually be seen** (`057211de`, `2e890ce8`, `14cf72d0`)

The tile surface always carries the result (subtitle/state description), and `BulkResultNotifier`
posts a notification when permitted. `POST_NOTIFICATIONS` is requested by the **ordinary runtime
dialog from a new row in Settings → Permissions**, API 33+ only.

Thor never self-grants it, even holding root or Shizuku: with the permission declared, the dialog
grants the identical capability, and Dhizuku cannot self-grant at all — its DPM alternative would
policy-fix the permission *on*, taking the choice away from the user. The row re-reads
`areNotificationsEnabled()` rather than trusting the launcher result, and deep-links to system
settings once the OS stops showing the dialog (two denials), so it cannot become a dead end
(`7919e370`).

**5. Lifecycle and layering** (`9e252b0d`, `6379d685`, `82139b20`)

`onStartListening` cancels any prior scope before creating a new one, `onDestroy` cleans up, and all
three `listenScope.launch` bodies rethrow `CancellationException` before the `catch (e: Exception)`
— there is no `CoroutineExceptionHandler` anywhere in `:app`, so an exception escaping a `launch`
would take down the process. `BulkResultText` moved to `util/` to remove the codebase's only
`data → presentation` import.

## Notes

- **The tile is freeze-only.** No unfreeze from the tile — that stays in the app and in the
  launcher shortcuts.
- No foreground service, no WorkManager, no ContentProvider. A bulk freeze is short and the
  reporting requirement is satisfied by the tile surface plus a notification.
- `docs/follow-ups/freezer-tile-service-rework.md` is **retired** — its acceptance criteria are
  written around the Toast that never worked, and its claim that the `if (scope != null)
  refreshTile()` guard "should survive" rests on `updateTile()` throwing, which it does not (it
  catches `RemoteException` internally).
- Three new follow-ups filed: `grantPermission` omits `--user` (affects the Permission Manager
  screen today, not this work), Odin caches root availability for the process lifetime, and the
  `BulkFreezeRunner` concurrency-test gap.
- The spec and implementation plan are committed under `docs/superpowers/`.
- `versionCode` deliberately not bumped — that belongs in a `chore(release)` commit.

## Verification

- `test --rerun-tasks` ✅ — **141 tests, 0 failures** on each of foss-debug and store-debug
  (104 before this branch → 141; 4 new test classes, 19 total)
- `assembleFossRelease` ✅ · `assembleStoreRelease` ✅ — the R8 path matters here: SystemUI reaches
  the tile only reflectively, so a release build is the only thing that proves it survives shrinking
- `lintFossDebug` ✅ · `lintFossRelease` ✅ · `lintStoreRelease` ✅ — 0 errors, no warnings beyond
  the 5 known intentional `VectorPath` ones. All four new string groups exist in all five locales.

**Device verification is still outstanding** — not automatable, and it includes the originally
reported bug. To run under **root, Shizuku and Dhizuku**:

1. Apps in the freezer, open QS → tile lit, "N apps · tap to freeze"
2. Tap → apps freeze; reopen the shade → tile dim, "No apps" ← **the reported bug**
3. Revoke privilege, reopen QS → tile greyed, taps do nothing
4. Tap, then collapse the shade immediately → the result still arrives
5. Notification with the permission granted, denied, and with the shade open vs collapsed
6. Settings notification row present on API 33+, absent below
7. Freeze-all shortcut now reports a result where it previously reported none
8. Unfreeze-all shortcut still unfreezes everything
9. Pinned per-app shortcut icons still grey/ungrey after a bulk run
10. Two rapid tile taps → one batch, not two
11. LeakCanary / heap dump after tap+collapse → no retained `FreezerTileService`

One open question only a device can answer: **does `Semaphore(5)` yield real parallelism under
root, or does Odin serialize on a single shell?** If it serializes, the five-wide fan-out only helps
the Shizuku and Dhizuku paths, which changes how likely the 30 s deadline is to bite on a large
watchlist.

## Summary by CodeRabbit

* **New Features**
  * Unified bulk freeze/unfreeze behavior across Quick Settings tile and launcher shortcuts.
  * Added bulk progress, incomplete-result handling, and summary reporting (including grouped notifications).
  * Added Settings controls for notification access, with Android 13+ runtime permission support.
* **Bug Fixes**
  * Improved cancellation handling and prevented stale/misleading bulk results.
  * Refreshed tile/shortcut visuals using accurate per-app freeze state.
* **Localization**
  * Added new translated strings for bulk progress, incomplete outcomes, and notification access.
* **Tests**
  * Added unit tests covering bulk action/result mapping and tile visual state logic.


trinadhthatakula:
## Code review

Adversarial review of the whole branch (`origin/dev...origin/fix/freezer-tile-rework`, 25 commits, +3637−215).

**Method** — I authored this branch, so the review was run by independent subagents rather than by me re-reading my own work. Two workflows, **18 agents total**, all on **claude-opus-5**: 15 agents (finders → dedup → per-finding adversarial verifiers prompted to *refute*, then a completeness critic), then 3 more to adjudicate the critic's late findings, two of which collided with earlier refutations at the same `file:line`. 15 raw findings → 10 unique → **10 confirmed, 4 refuted**. The top three were then re-verified by hand against the sources.

**Verdict: 1 regression vs `dev`, 2 more worth fixing before merge, 7 that can ship as follow-ups.**

---

### Important

**1. `AppFreezeStateReader.kt:26` — wrong `PackageManager` flag. Regression vs `dev`.**

Thor freezes *system* apps with `pm uninstall --user N`, not `pm disable`. The package is then not-installed-for-user, so `getApplicationInfo(pkg, MATCH_DISABLED_COMPONENTS)` throws `NameNotFoundException` → `stateOf` returns `ABSENT` → `freezableCandidates` drops it from the UNFREEZE set → **the Unfreeze-all launcher shortcut silently skips every frozen system app** (empty target list → `run()` returns `null` → no work, no result, no notification). On `dev`, `runBulk` called `forceUnfreeze` unconditionally, so this path worked.

`AppFreezeStateReader` is the sole deviant in the codebase — `AppInfoMapper.kt:72`, `AppRepositoryImpl.kt:119`, `Packages.kt:62`, `ShizukuReflector.kt:107` and `DhizukuReflector.kt:96` all query with `MATCH_UNINSTALLED_PACKAGES` and compute `enabled = info.enabled && (flags and FLAG_INSTALLED) != 0`.

Not Critical only because the tile is freeze-only (unaffected), the shortcut is opt-in, and in-app unfreeze still works. Fix = match `AppInfoMapper.kt:72`.

**2. `FreezerTileService.kt:125` — a cold-start tap can be silently dropped.**

`onClick` awaits `privilegeManager.state.first { it.isReady }` on `listenScope`, which `onStopListening` cancels. Collapse the shade before the IO probe resolves and the freeze never starts, with no feedback. The in-code comment calls this "acceptable, the user navigated away" — but the user *did* tap. It also contradicts device-verification item 4 in the PR body ("tap, then collapse the shade immediately: the result still arrives") in the cold case.

`BulkFreezeRunner.run()` already awaits `first { it.isReady }` itself (`5cefb34d`), so the pre-await exists only to drive the `!hasAnyPrivilege → paint()` feedback branch. Fix = start the run first, then probe for feedback.

**3. `BulkResultNotifier.kt:39` — UNFREEZE has no reporting surface when notifications are off.**

The KDoc claims "the tile's own subtitle is the unconditional surface, so a user who never grants `POST_NOTIFICATIONS` still sees their result." That is false for UNFREEZE: `5f1bc052` nulls `_lastResult` on unfreeze, and line 39 returns early when notifications are disabled. An unfreeze-all shortcut with notifications off therefore reports **nothing, anywhere**. Either the KDoc (and spec §D5) is wrong or the behaviour is.

---

### Minor — proposed as follow-ups

4. **`BulkFreezeRunner.kt:177`** — an UNFREEZE run's post-sweep poisons `_freezableCount`, so the tile paints a wrong state. Display-only: `STATE_INACTIVE` still delivers `onClick`, and `run()` recomputes candidates at tap time.
5. **`AppFreezeStateReader.kt:30`** — `stateOf` catches only `NameNotFoundException`, where the deleted `isFrozen` caught every `Exception`. `FreezerShortcutManager.pinAppShortcut:64` and `FreezerViewModel.pinAllToLauncher:359` are unguarded, so a binder-death `RuntimeException` now propagates where it previously did not.
6. **`SettingsScreen.kt:400`** — the Notifications row is gated on `SDK_INT >= TIRAMISU`, but `areNotificationsEnabled()` is meaningful and user-toggleable on 28–32 too, as `BulkResultNotifier.kt:34-36` itself documents. Sub-33 users who muted Thor get silent drops with no in-app row or deep link.
7. **`BulkFreezeRunner.kt:155`** — `_lastResult` has process-lifetime scope with no expiry; a shortcut-issued result can surface in the tile hours later.
8. **Stale result across non-runner paths** — `_lastResult` is cleared only by the tile displaying it or by a runner UNFREEZE. Unfreezing from Settings, `FreezerViewModel`, `AppListViewModel`, `MainViewModel` or an extension leaves it armed. Cosmetic, lasts one shade-open window, self-heals.
9. **`BulkFreezeRunner.kt:91`** — op-agnostic `_freezableCount`/`_isRunning` (the known deferred B-9).
10. **`BulkFreezeTest.kt:46`** — tautological: asserts only a data-class getter. Would still pass if `bulkActionFor` and `BulkAction` were deleted.
11. **`FreezerShortcutManager.kt:43`** — the new DI edge pulls `PrivilegeManager` into `ThorApplication`'s eager cold-start graph at boot (`ThorApplication.kt:90 → AutoFreezeManager.kt:38 → FreezerShortcutManager.kt:43 → BulkFreezeRunner.kt:51`).

---

### Refuted — no action

- **`BulkFreezeRunner.kt:224`, "unbounded prologue outside `DEADLINE_MS`"** — `TileVisual` tests `!isReady → CHECKING` *before* `isRunning → WORKING`, so "stuck on Freezing…" is structurally impossible; `isReady` is monotonic; and Odin's `RealShellRepository.isRootGranted()` is `withTimeoutOrNull(10_000)` with construction dispatched to `Shell.EXECUTOR`, so no caller thread parks. The residual untimed window (candidate sweep, handoff join, `CANCEL_GRACE_MS`, the `NonCancellable` post-run sweep) is bounded by a user-curated watchlist of client-cached binder calls on `Dispatchers.IO`, and every path reaches the `finally`. Doc-only nit at most.
- **`BulkResultText.kt:31`** — failure/unresolved wording is correct as written.
- **`SettingsViewModel.kt:168`, cross-path race** — real, but pre-existing and strictly *improved* by this PR.

---

On-device verification (11 items × root/Shizuku/Dhizuku) is still the outstanding gap and is unaffected by the above.

trinadhthatakula:
## Code review

**Verdict: Do not merge yet.**

### Findings

1. **High — Unfreeze-all skips frozen system apps.**

   `AppFreezeStateReader.stateOf()` queries with only `MATCH_DISABLED_COMPONENTS`. Thor freezes system apps using `pm uninstall --user`, so those packages are not installed for the current user and this lookup returns `NameNotFoundException`. They become `FreezeState.ABSENT`, are excluded from the `UNFREEZE` target set, and the launcher shortcut silently skips them.

   Evidence: [AppFreezeStateReader.kt#L23-L31](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/data/freezer/AppFreezeStateReader.kt#L23-L31), versus the system-app freeze path in [RootSystemGateway.kt#L223-L245](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt#L223-L245).

   **Suggested fix:** query with `MATCH_UNINSTALLED_PACKAGES` and include `FLAG_INSTALLED` when deriving the frozen state, matching the existing app mapping logic.

2. **High — A cold-start tile tap can be silently lost.**

   `onClick()` waits for privilege readiness inside the TileService listening scope. `onStopListening()` cancels that scope, so collapsing the shade before the cold-start probe completes prevents `runner.launch()` from ever being called. This contradicts the stated “tap, then collapse the shade immediately → result still arrives” verification case.

   Evidence: [FreezerTileService.kt#L116-L136](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/presentation/tile/FreezerTileService.kt#L116-L136).

   **Suggested fix:** hand the accepted tap to the process-scoped runner immediately; keep any privilege-feedback painting separate from the work lifetime. The runner already awaits `PrivilegeState.isReady`.

3. **Medium — Unfreeze shortcuts can complete with no reporting surface.**

   UNFREEZE results are explicitly removed from `_lastResult`, while `BulkResultNotifier` returns without reporting when notifications are disabled or its channel is silenced. Because the launcher trampoline finishes immediately, an Unfreeze-all shortcut can therefore complete without any visible result. This conflicts with the PR’s shared-reporting claim.

   Evidence: [BulkFreezeRunner.kt#L145-L157](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/data/freezer/BulkFreezeRunner.kt#L145-L157) and [BulkResultNotifier.kt#L30-L40](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/data/freezer/BulkResultNotifier.kt#L30-L40).

4. **Medium — UNFREEZE publishes the wrong candidate meaning for the freeze-only tile.**

   The runner’s single `freezableCount` flow is consumed as the tile’s FREEZE count, but the post-run sweep refreshes it using the completed operation. After unfreezing everything, it publishes zero UNFREEZE candidates even though every restored app is now a FREEZE candidate. The next tile refresh normally corrects it, but the cached state is wrong and remains wrong if that refresh fails.

   Evidence: [BulkFreezeRunner.kt#L170-L179](https://github.com/trinadhthatakula/Thor/blob/5f1bc0527b515aeaf9bec5c745e17025af6cc813/app/src/main/java/com/valhalla/thor/data/freezer/BulkFreezeRunner.kt#L170-L179).

   **Suggested fix:** keep the tile-facing count explicitly freeze-specific or maintain separate counts per operation.

### Summary of changes

This PR reworks the Freezer QS tile to derive its state from actual per-app freeze state, introduces a process-scoped concurrent `BulkFreezeRunner`, delegates launcher bulk shortcuts to it, adds tile/notification result reporting, adds the notification-permission settings row and localization, and adds pure state/result tests plus design documentation.

Scope: **28 files, +3,637 / −239**.

### Validation

- GitHub PR CI, CodeQL, manifest checks, and label checks are green.
- CodeRabbit’s status is green, but its actual review was skipped because `dev` is not the default branch.
- Local `./gradlew test --rerun-tasks --max-workers=1` passes.
- A normal parallel local run exposed a pre-existing `BundleZipTest` temporary-directory collision; the failed test passes in isolation.
- `BulkFreezeRunner` still has no executable concurrency/handoff tests.
- The root/Shizuku/Dhizuku device-verification matrix remains outstanding.

### Recommendation

Fix findings 1–3 before merge, preferably finding 4 as well. Add runner concurrency coverage and complete the documented device matrix before approving.

**Review model:** OpenAI GPT-5 Codex, without delegated agents.

trinadhthatakula:
## Review findings addressed — `5f1bc052..ac0108e5`

Fixed the two Important code defects, corrected the false documentation claim, and closed the two Minor items that were cheap. The remaining six are filed rather than fixed, with the reasoning recorded.

| # | Finding | Outcome |
|---|---------|---------|
| 1 | `AppFreezeStateReader.kt` — wrong `PackageManager` flag (regression vs `dev`) | **Fixed** — `c4dedf04` |
| 2 | `FreezerTileService.kt` — cold-start tap silently dropped | **Fixed** — `e48125be` |
| 3 | `BulkResultNotifier.kt` — false "unconditional surface" KDoc | **Fixed** — `3cd67311` |
| 5 | `AppFreezeStateReader.kt` — narrowed catch vs the `isFrozen` it replaced | **Fixed** — `c4dedf04` |
| 10 | `BulkFreezeTest.kt` — tautological op assertion | **Fixed** — `154e3ba6` |
| 4, 6, 7, 8, 9, 11 | op-agnostic runner state · sub-33 notification row · `lastResult` expiry & ownership · eager cold-start graph | **Deferred** — `ac0108e5` |

### Notes on the fixes

**1 + 5** — query with `MATCH_UNINSTALLED_PACKAGES or MATCH_DISABLED_COMPONENTS` and fold `FLAG_INSTALLED` into `enabled`, matching `AppInfoMapper`/`AppRepositoryImpl`/`Packages`/`ShizukuReflector`/`DhizukuReflector`. `stateOf` now also catches every `Exception` → `ABSENT` (with `CancellationException` rethrown first, so a cancelled sweep cannot masquerade as a watchlist of absent packages), restoring what `FreezerShortcutManager.isFrozen` did before this class replaced it.

**2** — `onClick` now hands off synchronously to `runner.launch(BulkOp.FREEZE)` and awaits nothing. The pre-check turned out to be load-bearing only for the failure mode it introduced: `BulkFreezeRunner.run` already awaits `isReady` itself and no-ops without privilege, the collector in `onStartListening` already repaints `NO_PRIVILEGE` on every `PrivilegeManager` emission, and `tileVisualFor` orders `NO_PRIVILEGE` ahead of `WORKING` so an about-to-no-op run cannot flash "Freezing…". Device-verification item 4 is now true by construction in the cold case, not just the warm one.

**3** — documented rather than changed, deliberately. The alternative is to park UNFREEZE results in `lastResult`, which reintroduces exactly the staleness `5f1bc052` fixed. A silent unfreeze beats a freeze tile that says "Unfroze 5 apps" hours later. Giving `FreezerLaunchActivity` a surface of its own — it *is* an Activity, so unlike a `TileService` it can legitimately show one — is in the follow-up.

**10** — replaced with the uncovered case: `total = 5, succeeded = 0, failed = 0`, the deadline firing before a single binder call returns. All five must land in `unresolved`; the pre-rework tile called them failures.

### Gates, re-run on the new tip

- 141 unit tests × fossDebug **and** storeDebug, 0 failures / 0 errors (`--rerun-tasks`)
- `lintFossDebug` + `lintFossRelease` + `lintStoreRelease` — 0 errors; 5 `VectorPath` warnings, all pre-existing and intentional
- `assembleFossRelease` + `assembleStoreRelease` — green
- 3 Kotlin compiler warnings, all pre-existing in `ExtensionManager.kt`

On-device verification (11 items × root/Shizuku/Dhizuku) is still outstanding and is the remaining gate before merge.

trinadhthatakula:
## Code Review Summary

**Model Used:** Gemini 3.6 Flash  
**Verdict:** **APPROVED / MERGE** (Ready to merge once manual device verification passes)

### 1. Executive Summary
PR #284 resolves a critical defect in the Quick Settings (QS) Freezer tile where the tile derived state solely from watchlist size rather than checking whether any app was actually unfrozen. It also fixes invisible toast results and unifies launcher shortcut execution.

### 2. Code Changes Breakdown
* **Domain & State Logic:** Introduced `FreezeState`, `freezableCandidates()`, `BulkOp`, `BulkAction`, and `BulkResult`.
* **Data Layer & Execution:** Created `AppFreezeStateReader` for live state reads, `BulkFreezeRunner` (`@Single`) for unified job coalescing and concurrency control with `Semaphore(5)`, and `BulkResultNotifier` for system notifications.
* **Presentation & UX:** Transformed `FreezerTileService` into a passive visual painter, implemented a pure framework-free state machine in `TileVisual.kt`, and added a `POST_NOTIFICATIONS` permission row in `SettingsScreen.kt`.
* **Testing & Documentation:** Added 4 new test classes bringing test count to 141 tests (0 failures).

### 3. Highlights & Key Architectural Victories
1. **Coroutine Safety:** Robust cancellation handling (`CancellationException` re-thrown ahead of broad catches).
2. **Framework Decoupling:** `TileVisual` and `bulkActionFor` are pure Kotlin unit tests.
3. **No Process Leaks:** `BulkFreezeRunner` as `@Single` prevents service retention leaks on shade collapse.
4. **Android OS Quirks Addressed:** `TileVisual.CHECKING` maps to `STATE_INACTIVE` (prevents AOSP click swallowing), Toast layer issues solved with notifications, and TalkBack string duplication fixed on API 30+.

### 4. Verdict
**APPROVED / MERGE** - Architectural quality is outstanding, tests are thorough, and OS edge cases are handled meticulously.

trinadhthatakula:
## Second review round addressed — `ac0108e5..49471dd6`

Both remaining findings from the GPT-5 Codex review, both of its infrastructure notes, and the two
lowest-impact items from my own deferred list. Six commits.

| Source | Finding | Outcome |
|---|---|---|
| Codex #3 · mine #3 | Unfreeze-all reports nothing when notifications are off | **Fixed** — `0f7e85bc` |
| Codex #4 · mine #4 + #9 | Runner publishes op-agnostic `freezableCount` / `isRunning` | **Fixed** — `3f95fa9d` |
| Codex (validation note) | CodeRabbit skipped this PR — `dev` is not an allowed base | **Fixed** — `233e7765` |
| Codex (validation note) | `BundleZipTest` temp dirs collide across forks | **Fixed** — `f1ab0c9a` |
| mine #6 | Notification row hidden below API 33 | **Fixed** — `82c245b3` |
| mine #7 + #8 | `lastResult` ownerless, never expires | **Still deferred** |
| mine #11 | `PrivilegeManager` in the eager cold-start graph | **Still deferred — "measure this"** |

### Notes

**Per-op state.** `refreshCandidates(op)` → `refreshFreezableCount()`, which always sweeps
`FREEZE`; `isRunning: Boolean` → `runningOp: BulkOp?`, and the tile derives
`isRunning = runningOp == FREEZE`. The fix was smaller than the deferral assumed — the field was
already *named* `freezableCount`, so this makes the name true rather than encoding "the tile is
freeze-only" into the runner. Second-order win: a background Unfreeze-all no longer suppresses a
finished freeze result in the subtitle either.

**Unfreeze reporting.** `launch()` returns `Deferred<BulkResult?>`; `FreezerLaunchActivity` awaits
it and toasts the outcome. That is legal there and only there —
`NotificationManagerService.checkCanEnqueueToast` drops a toast from a background package whose
notifications are disabled, which is exactly what the app-scoped runner is, but the trampoline is a
resumed Activity. The wait is bounded at 2 s, well inside the run's own 30 s deadline: the
trampoline's window is invisible yet touchable, so sitting on it would swallow taps meant for the
launcher underneath. Past the window the user gets "Unfreezing Apps" and the run finishes without
us. Zero new strings — the three cases reuse `bulkResultMessage`, `tile_no_apps_toast` (what
`SettingsViewModel` already shows for Unfreeze-all on an empty watchlist) and the `log_*_batch`
pair, all four locales included.

**Notification row.** The deferral rationale here was wrong on the facts and the follow-up doc now
says so: it asserted an i18n cost from the shape of the change instead of checking the strings
file, where all three strings and `appNotificationSettingsIntent()` already existed. The real
change is a removed version gate plus one branch — 33+ requests the permission, 28–32 deep-links,
as the usage-access row directly above already does. The launcher stays registered inside the
version check so every `POST_NOTIFICATIONS` reference stays behind it for lint's `InlinedApi`.

**Why 7/8 and 11 are still deferred.** #7/#8 want one writer for `lastResult`, which means routing
every bulk unfreeze through the runner — changing who writes shared mutable state in a concurrent
class that still has no tests, which is precisely how the two defects `bulk-freeze-runner-concurrency-tests.md`
documents got in. #11 is unmeasured; restructuring DI on the strength of a call graph with no trace
behind it is how cold starts get slower.

### Gates, re-run on the new tip

- 141 unit tests × fossDebug **and** storeDebug, 0 failures / 0 errors (`--rerun-tasks`)
- `lintFossDebug` + `lintFossRelease` + `lintStoreRelease` — 0 errors; 5 `VectorPath` warnings, all
  pre-existing and intentional
- `assembleFossRelease` + `assembleStoreRelease` — green
- Kotlin compiler warnings unchanged: the same 3 `Elvis operator` warnings in `ExtensionManager.kt`
  (78, 218, 243) plus the Koin-plugin version notice. Nothing new from these commits.

On-device verification (11 items × root/Shizuku/Dhizuku) remains the outstanding gate before merge.
Two new cases join it: Unfreeze-all now toasts its outcome from the trampoline, and the
Settings → Permissions notification row is now visible on API 28–32.

<sub>Review and fixes by Claude Opus 5.</sub>

trinadhthatakula:
@coderabbitai review

trinadhthatakula:
@coderabbitai full review

trinadhthatakula:
## CodeRabbit round — all 24 items triaged

Thanks @coderabbitai. Went through every item: **19 inline comments + 5 nitpicks** across the two reviews (`AppFreezeStateReader` was raised in both, so 23 distinct). Result: **4 code fixes, 16 doc items resolved, 3 refuted, 1 accepted-with-a-different-remedy.**

Two commits: [`96c2748`](https://github.com/trinadhthatakula/Thor/commit/96c2748) (code) and [`5469370`](https://github.com/trinadhthatakula/Thor/commit/5469370) (docs).

---

### The Major one: unbounded `NonCancellable` sweep — finding accepted, remedy rejected

`BulkFreezeRunner.kt:204`. **The finding is right and it was the most valuable thing in this review.** The post-run sweep ran inline inside `withContext(NonCancellable)`, which made it the one unbounded wait in the class: a wedged binder — PMS lock contention right after a batch of `pm disable`/`pm uninstall --user` is not exotic — pinned `_runningOp` and `activeJob` for the process lifetime, leaving the tile on "Freezing…" while every later tap coalesced into the stuck job.

**The proposed diff cannot fire, though.**

```kotlin
withTimeoutOrNull(CANCEL_GRACE_MS) { refreshFreezableCount() }   // ← no-op here
```

`withTimeout` installs its own `TimeoutCoroutine` and *cancels that coroutine*; `withContext` returns only when its body completes. `refreshFreezableCount()` ends in `freezableCandidates()`, a plain **non-suspending** `filter` over one blocking `getApplicationInfo` binder call per watchlist entry. A cancelled-but-non-suspending block has no suspension point to throw at, so the timeout would be recorded and then wait out the full loop anyway. Same trap the file's own **B-10** comment already documents for the worker cancellation path.

The only thing that is actually interruptible is a `join()` on a *separate* job, so that's what shipped:

```kotlin
withContext(NonCancellable) {
    val sweep = scope.launch { sweepFreezableCount() }
    if (withTimeoutOrNull(SWEEP_GRACE_MS) { sweep.join() } == null) {
        // Abandoned, not cancelled: it still publishes when the binder returns.
        Logger.d("BulkFreezeRunner", "post-run candidate sweep exceeded ${SWEEP_GRACE_MS}ms; releasing the run")
    }
}
```

Three things that fell out of it and are worth stating, because they are not obvious from the diff:

- **`NonCancellable` is still load-bearing.** `scope.launch` makes the sweep *complete*; `NonCancellable` is what makes the `finally` *wait* for it. Without it, `sweep.join()` resumes with `CancellationException` immediately under outer cancellation and the bound is meaningless.
- **The abandoned job needs its own failure handling** (new `sweepFreezableCount()`). It runs on `scope`, nothing awaits it, and `SupervisorJob` stops a failing child from cancelling its *siblings* — not from being *reported*. There is no `CoroutineExceptionHandler` anywhere in `:app`, so an escaping exception reaches Android's default handler and kills the process. `CancellationException` is rethrown first, since it is an `Exception` in Kotlin.
- **`SWEEP_GRACE_MS = 10_000`, deliberately not `CANCEL_GRACE_MS`.** That constant bounds an *unwind*; this bounds *real work* against a PackageManager that was just made to disable or uninstall a batch. It should be a wedge detector, not a load limiter — a sweep that misses costs one stale paint, and the tile re-sweeps on every `onStartListening`.

### Fixed as proposed

| # | Item | Resolution |
|---|---|---|
| Nit | `FreezerShortcutManager` — redundant icon-refresh jobs on repeated taps | Fixed. `refreshScheduledFor` guard, at most one refresh per run, matched by **identity** so a conflicting op still gets its own (it leaves the packages in the opposite state). Not cosmetic: `launch()` coalesces same-op taps and the bulk shortcut shows nothing for up to 2 s, so impatient re-taps are the *expected* case — each one cost another N concurrent icon decodes for byte-identical output. |
| Nit ×2 | `AppFreezeStateReader:38` — unused `NameNotFoundException` binding | Fixed. `catch (_: …)` + a line saying why "no such package for this user" is the expected answer here. |
| #14 | `FreezerLaunchActivity:101` — unused `CancellationException` binding | Fixed. `catch (_: CancellationException)`; `ensureActive()` propagation unchanged. Same for the `isSuspended()` catch. |

### Refuted

**1. `log_freezing_batch` / `log_unfreezing_batch` "reuses terminal-log copy as toast text."** The premise is wrong — these are already user-facing strings ("Freezing Apps" / "Unfreezing Apps"), rendered by `FreezeLoggerDialog.kt:115`. They are not terminal-log strings; the `log_` prefix names the dialog they belong to. Adding dedicated keys would cost 2 strings × 5 locales for text that is identical to what already ships.

**2. `SettingsScreen:460` — "toggling the row off is a silent no-op, the switch animates off and snaps back."** Two problems. The described symptom does not happen: M3 `Switch` is fully controlled, `checked` drives the thumb, so with a no-op handler it does not move at all. And the sibling **usage-access** row directly above it (`SettingsScreen.kt:382`) has exactly the same `if (!granted) { deep-link }` shape — changing only one would leave two identical-looking rows behaving differently, which is worse than the thing being fixed. Worth doing as a pair, if at all; not in this PR.

**3. The 16 doc items — resolved, but not the way suggested.** Every one is factually correct: the plan and the spec do describe `sealed TileVisual`, `isRunning`, `refreshCandidates(op)`, `consumeResult()`, `runBulk(): Unit`, a per-run semaphore, "abandoned workers are harmless", and an API-33 gate. None of that ships.

But editing those out of the body would make the documents accurate and useless. A design record is worth keeping precisely because it says what was decided **before** implementation; rewriting it to match the outcome destroys the only evidence that the outcome differed — and the divergence *is* the interesting part, since most of it is review findings landing. So both documents keep their bodies and gain a **Status** line plus an **"As-built deltas"** section — designed / shipped / why it changed — covering all 16 points, with the spec's verification table marked as predating `tileVisualFor`'s `isRunning` branch (#13) and the API 28–32 deep link (#9).

Three corrections *do* go into the spec body, because they are wrong rather than superseded: `TileVisual` is an `enum class` (#2), `BulkResultText` lives in `util/` not `presentation/tile/`, and the `SettingsScreen` change is the permission row — which reads the framework directly and needed no ViewModel change (#4).

`docs/follow-ups/…-deferred-review-findings.md` §1 (#15) is a straight accept: `BulkFreezeRunner` *does* own `_lastResult` — private, published read-only, `consumeResult` the only external mutator. What it lacks is a **lifecycle policy**. Retitled and reworded.

---

### Verification

Everything above was re-checked adversarially before pushing: **Claude Opus 5, 20 agents** — 4 review lenses (concurrency, lifecycle, behaviour, docs) pipelined into per-finding refutation agents, 1.27 M tokens. **0 confirmed findings.**

The near-misses are worth recording, since both were argued well and both turned out to be pre-existing:

- *"The abandoned sweep can overwrite a newer count — `_freezableCount` has no generation guard."* True that there's no guard, false that this diff introduces the race. `FreezerTileService.onStartListening()` already launches its own `refreshFreezableCount()` on `listenScope` on **every shade open**, on `Dispatchers.IO`, unserialised — and once that non-suspending filter is entered, cancelling `listenScope` cannot stop the publish. `_freezableCount` never had totally-ordered writes. A generation stamp would be reasonable hardening for that pre-existing multi-writer field; it isn't this PR's.
- *"The conflicting-op comment still claims the replacement waits for the previous run's sweep."* The sweep performs no mutation — `getApplicationInfo` only — so it cannot violate what that comment guards ("a straggling FREEZE worker can re-disable a package the replacing UNFREEZE batch already enabled"). And the KDoc six lines above already says the handoff is *"bounded, not absolute."*

Gates on the pushed tree: **141 unit tests × fossDebug + 141 × storeDebug, 0 failures** (`--rerun-tasks` — a bare run reports UP-TO-DATE and silently skips); lint fossDebug/fossRelease/storeRelease **0 errors** (5 known intentional `VectorPath` warnings each); `assembleFossRelease` + `assembleStoreRelease` green.

Still outstanding on this PR, unchanged: on-device verification of the 11 items in the description across root / Shizuku / Dhizuku, plus the unfreeze toast from the trampoline, the Settings notification row on API 28–32, and Unfreeze-all against a watchlist containing frozen **system** apps.

PR #286, opened 2026-07-28
fix(freezer): rebuild pinned shortcut icons from the runner, not from one caller

trinadhthatakula:
Fixes item **9** of #284's device-verification list: *"pinned per-app shortcut icons still grey/ungrey after a bulk run"*. Reported failing on device — apps froze, icons stayed coloured.

## The bug

`FreezerTileService.onClick` is one line: `runner.launch(BulkOp.FREEZE)`. That is correct — a QS tile has no reason to know shortcuts exist. But the icon rebuild was bolted to `FreezerShortcutManager.runBulk`, whose only caller is the launcher trampoline. So the side effect was reachable from **one** of the runner's two callers and from nowhere else, and a tile freeze issued no `updateShortcuts` at all.

Reproduced on an emulator (API 37, Pixel Launcher, `dumpsys shortcut`):

| action | package | publisher `Calls:` | shortcut `timestamp` | stored bitmap |
|---|---|---|---|---|
| baseline | `enabled=1` | 0 | …731420 | 30044 B (colour) |
| **Freeze-all** shortcut | `enabled=3` | **1** | **…877330** | **22572 B (grey)** — repainted live |
| **Unfreeze-all** shortcut | `enabled=1` | 1 | …950576 | 30044 B (colour) |
| **QS tile** | `enabled=3` — freeze worked | **1 (unchanged)** | **unchanged** | **unchanged (colour)** |

That also rules out the launcher-side theories: `updateShortcuts` *does* mutate an already-pinned shortcut and Pixel Launcher repaints it with no relaunch; rate limiting was not in play (`Calls: 1` of 10); the shortcut was `disabledReason=[Not disabled]`; and the `FLAG_MATCH_PINNED` id matched byte-for-byte.

Not a regression from #284 — the pre-rework tile never touched `ShortcutManager` either. What the rework changed is that `BulkFreezeRunner` became the single owner of bulk freeze/unfreeze, which is what makes this gap structural and fixable in one place.

## The fix

`BulkFreezeRunner` publishes a `completions: SharedFlow<BulkOp>`; `FreezerShortcutManager` subscribes in its `init`. The dependency direction is forced by DI — the manager already injects the runner, so it observes, and the runner stays free of any launcher concern instead of acquiring a Koin cycle. No future caller can forget it.

Startup is not a race: Koin builds the manager as a constructor argument of `AutoFreezeManager`, and `ThorApplication.onCreate` calls `startObserving()` synchronously, so the collector is live before `onCreate` returns — and `Application.onCreate` always completes before the framework binds a `TileService`.

The per-run dedupe added in #284 goes away with `scheduleIconRefresh`: one run emits one completion, and the buffer (capacity 1, `DROP_OLDEST`) collapses a burst of impatient re-taps into one trailing rebuild. Correctness under that coalescing comes from the rebuild reading live state rather than anything carried on the emission.

## Two more from the same sweep

- **Settings → Unfreeze-all** is the one bulk path that does not go through the runner (its own `awaitAll` over `forceUnfreeze`), so it had the mirror symptom: icons stay **grey** after restoring. It now asks for the rebuild explicitly. Behaviour otherwise unchanged — same packages, same toast.
- **`updateShortcuts` returns `false`, it does not throw.** `ShortcutManager` rate-limits a *background* publisher (~10 accepted calls / 24 h, counter reset whenever the app has been foreground), and the tile path is background by definition. Both call sites discarded that boolean, so a dropped update was indistinguishable on screen from "we never called it". Both now route through `pushShortcutUpdate`, which logs the rejection.

## Left alone, deliberately

`appLabel` (`MATCH_DISABLED_COMPONENTS` only) and `appIcon` (`getApplicationIcon(String)`, no MATCH flags) both drop a system app frozen via `pm uninstall --user N`. Unreachable today — per-app pins are `!isSystem`-gated at four sites — but they go live the moment "v1: user apps only" is lifted, and **they must be fixed together**: patching `appLabel` alone un-masks `appIcon` and turns "no update" into a fallback snowflake glyph. Filed rather than fixed blind.

## Verification

- `testFossDebugUnitTest --rerun-tasks` — **141 tests, 0 failures**
- `lintFossRelease` — **0 errors**, only the 5 known intentional `VectorPath` warnings
- `assembleFossRelease` — green

**Device verification outstanding**, on root / Shizuku / Dhizuku:
1. Pin an app shortcut, freeze from the **QS tile** → icon goes grey ← the reported bug
2. Same app, unfreeze from the tile's watchlist path → icon returns to colour
3. Freeze-all / Unfreeze-all launcher shortcuts → unchanged from #284 (regression check)
4. Two rapid tile taps → one rebuild, not two
5. Settings → Unfreeze-all with a pinned frozen app → icon returns to colour
6. Auto-freeze (screen off) → per-app refresh still works

## Diagnosis method

Claude Opus 5, 7 agents, 656 k tokens: 4 competing hypotheses, each pipelined into two adversarial refuters (facts lens + symptom lens), killed only if both refuted. Three were refuted — one of them empirically on a connected device, which is what produced the table above.

## Summary by CodeRabbit

* **New Features**
  * Added automatic refreshes for pinned app shortcut icons after bulk freeze and unfreeze operations.
  * Added a manual option to rebuild pinned shortcut icons using the latest freezer state.
  * Improved shortcut icon updates following individual app state changes.

* **Bug Fixes**
  * Improved reliability of shortcut icon refreshes during bulk operations.
  * Preserved cancellation handling while rebuilding shortcut icons.

* **Documentation**
  * Documented potential matching issues when resolving system app labels and icons for shortcuts.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/freezer/BulkFreezeRunner.kt
+++ b/app/src/main/java/com/valhalla/thor/data/freezer/BulkFreezeRunner.kt
@@ -22,9 +22,13 @@ import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
@@ -74,6 +78,33 @@ class BulkFreezeRunner(
 
     /** Outcome of the last completed run, consumed once by whoever displays it. */
     val lastResult: StateFlow<BulkResult?> = _lastResult.asStateFlow()
+
+    // extraBufferCapacity + DROP_OLDEST so tryEmit never fails and never suspends the run: a
+    // slow subscriber (the icon rebuild decodes one bitmap per pinned app) must not be able to
+    // hold a batch open, and coalescing a burst down to "rebuild once more afterwards" is
+    // correct because every subscriber re-reads live state rather than the emitted value.
+    private val _completions = MutableSharedFlow<BulkOp>(
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    )
+
+    /**
+     * Emits once per run that actually touched packages — the seam for side effects that must
+     * follow *any* bulk run, whichever surface started it.
+     *
+     * This exists because side effects bolted to one caller are unreachable from the other.
+     * The pinned-shortcut icon rebuild used to hang off `FreezerShortcutManager.runBulk`, so a
+     * freeze started from the QS tile — which calls [launch] directly, having no reason to know
+     * shortcuts exist — froze the apps and left every pinned icon in full colour. Device
+     * evidence: `dumpsys shortcut` showed the publisher's call count and stored icon bitmap
+     * unchanged across a tile freeze that did set `enabled=3`.
+     *
+     * Not emitted for a no-op run ([run] returned null: no privilege, or nothing to act on) —
+     * nothing changed, so nothing needs rebuilding. Not emitted for a cancelled run either;
+     * the conflicting op that replaced it emits its own completion, and subscribers rebuild
+     * from live state, so the later emission subsumes the lost one.
+     */
+    val completions: SharedFlow<BulkOp> = _completions.asSharedFlow()
 
     private val _runningOp = MutableStateFlow<BulkOp?>(null)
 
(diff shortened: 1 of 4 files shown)


trinadhthatakula:
Thanks — three of the four are real. Addressed in `8b797a72`.

**1. `tryEmit` after `notifier.post` — accepted.** ✅
`post()` is a binder call into `NotificationManager`; anything it threw would be caught by the run's outer `catch` and would silently skip the emission for a run whose packages are *already* mutated. That is the same class of bug this seam exists to fix, just reached by a different route. `tryEmit` now runs first — it can neither throw nor suspend, so nothing is owed to the other ordering.

**2. `extraBufferCapacity` without `replay` — accepted.** ✅
Correct: `extraBufferCapacity` only buffers for an *already-subscribed but slow* collector; with zero subscribers `tryEmit` returns `true` and discards. `FreezerShortcutManager` subscribes from `scope.launch` in its `init`, which only *schedules* the collector, so subscription is not ordered against the first run. Added `replay = 1`. It's safe here precisely because subscribers re-read live package state rather than trusting the emitted value — a replayed completion costs one extra idempotent rebuild, never a wrong icon. The `init` comment that claimed "startup is not a race" was overstating the guarantee and has been corrected to say what actually holds.

**3. Future date in the follow-up doc — refuted.** ❌
`2026-07-29` is correct, not tomorrow. The commit that added the file is timestamped `2026-07-29T00:59:35+05:30` — the doc was written just past midnight IST, so it is dated the day it was actually written. No change.

**4. `appIcon` throws — accepted.** ✅
Right, the wording was imprecise: `appIcon` has its own `try/catch`, so what throws is the `getApplicationIcon(pkg)` call *inside* it, and `appIcon`'s `catch` swallows that and returns the `R.drawable.frozen` fallback. That's exactly why the failure surfaces as a white blob rather than an error, which is the point the section is making. Reworded.

Gates on the updated tree: `testFossDebugUnitTest --rerun-tasks` → **141 tests, 0 failures**; `lintFossRelease` → **0 errors** (5 results, all the intentional `VectorPath` warnings); `assembleFossRelease` → BUILD SUCCESSFUL.

PR #290, opened 2026-07-29
chore: the Tier-0/1 deferred-work batch (#5, #11, #14, #16, #17, #18, #20, #21, #22, #28)

trinadhthatakula:
Eight Tier-0/1 items from the 2026-07-29 deferred-work sweep, plus the CI fix that was blocking master. Each commit is one item and carries its own reasoning; this is the map.

## What's in it

| # | Item | Commit |
|---|---|---|
| #5 | master's CI has been red continuously, on the versioning convention being *followed* | `ci:` |
| #11 | delete the unused `room` version bundle; make `kotlinx-coroutines-android` explicit | `build:` |
| #14 | `ExtensionOpsProvider` saw only the `pm disable` half of a freeze | `fix(extensions):` |
| #16 | add `kotlinx-coroutines-test` + turbine | `build:` |
| #17 | `grantPermission` omitted `--user` in all three gateways | `fix(gateway):` |
| #18 | the single-app freeze paths had no tier gate — only a hidden button | `feat(freezer):` |
| #20 | backup rules were still the unedited AGP template | `fix(backup):` |
| #21 | Android Lint was `continue-on-error: true` | `build:` + `ci:` |
| #22 | nobody had measured `PrivilegeManager`'s cold-start cost | `chore(startup):` |
| #28 | drop Qodana | `build:` |

## The two that are worth reading carefully

**#5 — why master was red.** Every push to master published to Play's internal track, but this repo's rule is that `versionCode` moves *only* in a dedicated `chore(release)` commit. So most master pushes deliberately carry an unchanged code, and each one died on `Version code NNNN has already been used` after seven minutes of build. Run [30286348722](https://github.com/trinadhthatakula/Thor/actions/runs/30286348722) is the shape of it: a `fix(shizu):` commit, code unchanged at 1930, then the rejection. The check was firing on the convention being followed correctly.

Master pushes now **always build and sign** — that verification was never the problem. Only the Play upload, the Telegram post and the GitHub pre-release are gated, on whether `versionCode` actually changed against `HEAD^`.

**Play's duplicate rejection is not suppressed anywhere and must not be.** It is the real versioning gate, it is server-side, and a `chore(release)` that reuses a code still fails exactly as it does today. This only decides whether a push was *meant* to publish.

**#18 — what was actually protecting BLOCKED apps.** Nothing in the code. `AppRiskDialog` renders no confirm button for `BLOCKED`, so on all three single-app paths there was simply nothing tappable wired to the freeze. That covered every surface that existed and nothing that didn't — a shortcut, an extension trigger, an automation intent or a widget would reach for the obvious entry point and hand a `packageName` straight to `pm uninstall --user`. `FreezeAppUseCase` is now the backstop under the dialog.

## Not fixed here, deliberately

`master` and `production` both upload the same `chore(release)` version code — to `internal` and `alpha` — and Play allows a code exactly once **per app**, not per track. So one of the two runs must always lose. This PR removes every *non-release* push from that collision, which is the great majority of them and the reason master was permanently red, but the release-commit case survives.

It needs a decision about release flow rather than a workflow edit, so it's written up in `docs/follow-ups/two-branches-one-play-version-code.md` with three options. Promoting `internal` → `alpha` instead of uploading twice is the only one that keeps both tracks *and* one artifact.

## Verification

| Check | Result |
|---|---|
| `:app:testFossDebugUnitTest --rerun-tasks` | **209 tests, 0 failures** (was 104) |
| `lintFossDebug` / `lintFossRelease` / `lintStoreDebug` / `lintStoreRelease` | all green — the debug variants had never been checked before, and both dependency issues surfaced there |
| `assembleFossRelease` / `assembleStoreRelease` | green, including `lintVital` |

Lint went from advisory to enforced with `warningsAsErrors`, so `app/lint.xml` records the three exemptions that are deliberate rather than debt: `VectorPath` (five brand-art paths; #269 already compacted everything compactable losslessly) and `GradleDependency` / `NewerVersionAvailable` ("a newer version of X is available" is a fact about the outside world, and under `warningsAsErrors` it turns every upstream publish into a red build on a commit that changed nothing). All three are `informational`, not ignored — a *new* instance still shows in the report, and `warningsAsErrors` only promotes `WARNING`, so none can fail the build.

## Needs a device — I can't verify these

1. **#17, work profile or Xiaomi Second Space.** Toggle a permission on a package *inside* the profile and confirm it changes there and not on the parent user's copy. Then toggle one on a parent-user package and confirm the same-user case is unchanged. Repeat per privilege mode — Root and Shizuku should match; Dhizuku may refuse the cross-user case, which is fine as long as it *reports* the refusal. Protocol is in `docs/follow-ups/grant-permission-missing-user-flag.md`.
2. **#20, backup.** `adb shell bmgr backupnow com.valhalla.thor`, wipe, restore, and confirm theme / sort / privilege mode / freezer mode survive and nothing else came back.
3. **#18, the three freeze paths.** Freeze a `BLOCKED` system app from the app list, the info sheet and the Freezer tab — each should show the skipped message rather than doing nothing silently. Then confirm an ordinary app still freezes from all three.
4. **#22, cold start.** `adb logcat -s ThorPrivPerf` on a debug build, once per privilege configuration. Protocol in `docs/follow-ups/privilege-manager-cold-start.md`.

## Summary by CodeRabbit

* **New Features**
  * Improved freezing/unfreezing behavior with clearer handling for blocked/unknown apps (safer fail-closed “unsafe/skipped” outcomes).
* **Bug Fixes**
  * Fixed permission grant/revoke on secondary Android users and work profiles by targeting the correct user.
* **Documentation**
  * Updated follow-up guidance, including release/versioning and verification notes.
* **Chores**
  * Hardened CI: stricter linting, path-based workflow skipping for dev checks, and safer production release gating with version-code validation.


trinadhthatakula:
### CodeRabbit review disposition — 8 applied, 9 not (`e3c6123`)

Every finding was checked against the code or the shipped artifact before being applied or dismissed.

**Applied (8)**

| Finding | What was actually wrong |
|---|---|
| 6 s sleep vs. 10 s probe timeout | Real, and the worst of the set. The measurement protocol force-stopped the process 4 s before Odin's own `SHELL_INIT_TIMEOUT_MS`, silently dropping the slowest runs — biasing results *fast*, in the direction that makes the problem look solved. Both blocks now poll; the N-run loop counts rather than greps, because logcat isn't cleared between iterations. |
| Config 3 verdict contradiction | Table said `investigate`, prose said `within budget`. Reconciled as investigate, and the split made explicit: probe is fine (67 ms median), `ready` is not (p90 725 ms) — so the cost sits outside what this doc instruments. |
| Option 2 vs. single-level logger | Understated. `Logger` gates all five levels on one flag *and ships in `thor-extension-api` 3.0.0* — ungating only `e` is a new API surface and an irreversible Central release, not an audit. Recommendation moved to option 3. |
| LeakCanary in the perf session | Correct. "Costs nothing to have running" was wrong; moved to its own pass. |
| Upload vs. promotion semantics | Correct, and the doc contradicted its own recommended option 1. Fixed. |
| MD040 × 4 | Fixed. |
| "Permanent" lockout | Title overstated what the body already got right. Retitled. |
| `hasCheckedAutoPrompt` scope | Right for a reason not given: it's `rememberSaveable` (screen-scoped, not session), **and** it's set when the prompt is *shown*, not when it's declined — so it isn't a decline record at all. |

**Not applied (9)**

- **Two Play upload paths** (Major) — already filed; the comment is on the doc that files it. The proposed fix (make one path build-only) is option 2 there, which the owner explicitly rejected: master publishing to `internal` is wanted behaviour. Option 1 (promote instead of re-upload) is the open decision.
- **`Lazy<BulkFreezeRunner>` names the wrong component** — refuted. The eager chain is `ThorApplication → AutoFreezeManager → FreezerShortcutManager → BulkFreezeRunner → PrivilegeManager`, so the runner *is* the seam that breaks it, exactly as `freezer-bulk-run-deferred-review-findings.md` §2 prescribes.
- **Six "future-dated 2026-07-30"** — the review ran on 2026-07-29; the dates are current.
- **Option 4's `isDebuggable = true`** — already gone, replaced by the `PRIVILEGE_TRACE` flag in `c86aa56`, which is what the finding asked for.

Docstring coverage (16% vs. 80%) is not actioned: this batch is documentation plus small targeted code changes, and blanket-generating KDoc to hit a threshold would add noise, not information.

PR #291, opened 2026-07-30
build(ci): close both open gem advisories and track bundler with dependabot

trinadhthatakula:
## Why

`dev` has **two open Dependabot security alerts**. Both are in `Gemfile.lock`, and both are closed by this PR:

| Severity | Gem | Vulnerable | Patched | This PR |
|---|---|---|---|---|
| **High** | `faraday` | `>= 1.0.0, <= 1.10.5` | `1.10.6` | `1.10.5` → **`1.10.6`** |
| **Moderate** | `excon` | `< 1.5.0` | `1.5.0` | `0.112.0` → **`1.6.0`** |

- faraday: uncontrolled recursion in `NestedParamsEncoder` → stack-exhaustion DoS via deeply nested query parameters.
- excon: does not redact additional sensitive headers when following redirects.

They stayed open because **`.github/dependabot.yml` had no `bundler` ecosystem** — only `gradle` and `github-actions`. So the one dependency surface with no update pressure was the one driving every signed build and both Play uploads, and nothing was ever going to raise a PR for it.

## What changed

**1. `.github/dependabot.yml` — add the `bundler` ecosystem** (the durable half). Weekly, `target-branch: dev`, grouped into one PR, matching the existing `github-actions` entry.

**2. `Gemfile.lock` — one lockfile refresh** (the backlog that had accumulated behind the gap). 17 gems moved, including the google-apis / aws-sdk set the Play upload rides on.

## The one coupling worth knowing

**fastlane 2.236.1 → 2.237.0 is what unblocks excon.** 2.236.1 pinned `excon < 1.0.0`, so excon *could not reach a patched version at all* until fastlane relaxed the constraint to `< 2.0.0`.

**faraday stays on the 1.x line** (`1.10.6`, not `2.14.3`) because fastlane still requires `~> 1.0`. `1.10.6` is the patched release on that line, so the high-severity advisory is fully addressed — this is fastlane's constraint, not a partial fix.

That mutual gating is why the ecosystem is **grouped** rather than per-gem: single-gem PRs would mostly be unmergeable on their own.

## `BUNDLED WITH` is deliberately unchanged

Left at `4.0.14`, not bumped to the local `4.0.17`. All four workflows use `ruby/setup-ruby` with `bundler-cache: true`, which installs the bundler named there and then runs a **frozen** install — so moving that line changes what CI actually executes and belongs in its own commit, not a gem refresh.

Bundler 4.0.17 also wants to add a `bundler (4.0.17) sha256=…` entry to the `CHECKSUMS` block, which `origin/dev` does not have. Dropped for the same reason, so the diff is purely the gems.

## Verification

- `bundle install` resolves and installs all **97 gems** from a clean bundle path.
- `bundle exec fastlane lanes` parses the Fastfile under 2.237.0 and lists all three lanes: `distribute_dev`, `distribute_production`, `build_release_candidates`.
- `.github/dependabot.yml` parses as YAML; all three ecosystems resolve with the intended schedule, target branch and group.

Not run: the lanes themselves, which need the Play service account and keystore secrets. CI exercises `bundle install` on this branch.

## Summary by CodeRabbit

* **Chores**
  * Added weekly automated updates for Ruby dependencies.
  * Grouped related dependency updates into a single update stream.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,3 +21,19 @@ updates:
       actions:
         patterns:
           - "*"
+
+  # Fastlane and its ~96 transitive gems drive every signed build and both Play
+  # uploads, so they were the one dependency surface with no update pressure at
+  # all — the lockfile had drifted far enough that a routine `bundle lock` moved
+  # 17 gems. Weekly and grouped, matching github-actions: these move together
+  # (fastlane's constraints gate excon/faraday), so per-gem PRs would mostly be
+  # unmergeable on their own.
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "dev"
+    groups:
+      fastlane:
+        patterns:
+          - "*"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,8 +8,8 @@ GEM
     artifactory (3.0.17)
     atomos (0.1.3)
     aws-eventstream (1.4.0)
-    aws-partitions (1.1261.0)
-    aws-sdk-core (3.252.0)
+    aws-partitions (1.1275.0)
+    aws-sdk-core (3.254.0)
       aws-eventstream (~> 1, >= 1.3.0)
       aws-partitions (~> 1, >= 1.992.0)
       aws-sigv4 (~> 1.9)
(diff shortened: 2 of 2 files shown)


trinadhthatakula (commented):
## Review summary##

The Bundler ecosystem is now covered by weekly grouped Dependabot updates, and the lockfile refresh correctly moves faraday and excon onto patched releases while respecting Fastlane constraints. I found no blocking or non-blocking code issues in this PR.

Verification performed in an isolated clone at the reviewed head commit:
- The Dependabot YAML parses successfully.
- Bundler 4.0.14 installed all 97 locked gems successfully.
- Fastlane parsed the Fastfile and found all three Android lanes.
- All reported GitHub checks are green.


**Verdict:** Approve.

**Review model:** OpenAI GPT-5 (Codex).

PR #292, opened 2026-07-30
fix(security): give the app lock a way out when no prompt can succeed

trinadhthatakula:
Closes the hard-lockout follow-up: `docs/follow-ups/biometric-lock-restores-without-a-capability-check.md`. The loop is gone on every supported API level, and as of the latest commit the **API 28 residual is closed too**, per the owner's call.

## The trap

`biometric_lock=true` restored onto a device with nothing enrolled was a hard lockout with **no in-app escape**:

1. `authState` came from the preference alone → `Locked` → prompt.
2. The prompt fails the instant it opens → `Error`.
3. The error screen offers **TRY AGAIN** → re-arms the same prompt → step 2.

`finish()` was the only other control. The switch that turns the lock off (`SettingsScreen.kt`, already correctly disabled when the device is incapable) lives behind `MainScreen`, which is never composed. **The guard was real and unreachable in exactly the case it exists for.**

Most likely moment for this to fire: setup-wizard restore, on a fresh or wiped device, *before* the user has been asked to set a screen lock. Also reachable with no backup at all — enable the lock, then remove your screen lock.

## What this does

**The lock is not lifted** — except in the one case where nothing could ever open it, below.

`AuthState` gains `Unavailable`, reached only when the lock is on and the device cannot authenticate. `BiometricUnavailableScreen` says so plainly and deep-links to security settings (`ACTION_BIOMETRIC_ENROLL` on API 30+, falling back to security settings then the settings root, each guarded against `ActivityNotFoundException`). `HomeActivity.onResume` re-queries capability *before* the Shizuku early-return, so the user who leaves, enrolls something, and comes back lands on a prompt that can actually succeed — no restart, no reinstall.

Where `Unavailable` sits in the `when`:

- **Above `Error`** — the prompt fails instantly on such a device, so an error is populated a moment later either way, and TRY AGAIN is the loop restated.
- **Below `authenticated`** — removing a screen lock mid-session must not evict someone who legitimately unlocked.

## The API 28 residual, and the owner's call

> "API 28 with no enrollable biometric will not be able to turn on biometric, we will just not allow him to turn on biometric and show them a toast saying biometric not enrolled. If the setting is restored from backup we do the same and turn off the setting ourself."

On an **Android 9 device with no biometric sensor** the unavailable screen was honest but useless: Android 9's framework prompt takes no device credential, so setting a PIN changes nothing and there is nothing to enrol on. That is not a locked door, it is a wall — the only exits were EXIT and clearing app data.

That is the follow-up's **option 2** (self-heal by writing `biometric_lock=false`) plus a refusal on the way in, and it ships **asymmetrically**, because the two halves cost different things when applied too widely:

- **Refusing to arm is universal.** `setBiometricLock(true)` is refused on *any* device where `canAuthenticate()` is false, and answers with a toast naming what is missing. Refusing costs the user nothing — they are standing in Settings and can go enrol. The switch is no longer `enabled = state.canUseBiometric`: a disabled row swallows the tap (`clickable(enabled = false)`), so the old behaviour was a greyed-out control and silence, with no tap left to answer. It asks `AuthCapability` fresh rather than reading the `uiState` snapshot, which only recomputes when the preferences flow emits.
- **Disarming is confined to the unrecoverable case** — `enrolmentCanFixLockout(SDK_INT, hasHardware)` false. Everywhere else the user can enrol their way in and `Unavailable` takes them there. Silently dropping a lock they could still open would be a security downgrade dressed up as a bug fix.

`enrolmentCanFixLockout` is a named predicate with its own test rather than an inline condition, because it is wrong in opposite directions: a `true` leaked in strands a user at a prompt that can never succeed, a `false` leaked out hands Thor permission to remove a working lock. It asks about sensor *presence*, not enrolment — a sensor with nothing enrolled is the recoverable case precisely because enrolling on it is the fix.

The disarm is driven off the preference flow in `init` rather than checked once, because the preference is read asynchronously: `_biometricEnabled` seeds `false` and the restored `true` lands a moment later, which is the case this exists for. Writing `false` flips that flow, so it settles after one pass. The matching `!capable && !enrolmentCanFix -> NotRequired` branch in `authState` keeps a frame of the dead end from rendering while the write lands, and `enrolmentCanFix` is `lazy` so the extra `BiometricManager` call stays off every cold start (#22).

Not silent: `SecurityViewModel` emits a `UiText` on a buffered `Channel` collected by `HomeActivity` — buffered, not `replay = 0`, because the disarm fires during `onCreate`, before any collector reaches STARTED. Thor removed a security control the user deliberately switched on; saying so is the minimum.

Options 1 (fail open unconditionally), 3 (in-app disable from the error screen) and 5 (`createConfirmDeviceCredentialIntent` on API 28) are now moot — the case each existed to rescue no longer occurs.

## The API 28-29 bug this surfaced

Found in review, and it would have shipped as a *second*, worse lockout.

`BiometricHelper` asked `canAuthenticate(BIOMETRIC_STRONG or DEVICE_CREDENTIAL)`. That is not a supported **combination** below API 30 — androidx's `AuthenticatorUtils.isSupportedCombination` rejects it for `SDK_INT` in `[P, Q]` and returns `BIOMETRIC_ERROR_UNSUPPORTED` without consulting the device at all. Against a `== BIOMETRIC_SUCCESS` test that is indistinguishable from "nothing is enrolled", so `canAuthenticate()` was **hard-false on every Android 9 and 10 device** whatever it had enrolled. minSdk is 28 — these are shipped, supported devices.

Previously that only suppressed the Settings toggle. Gating the *door* on it would have meant an Android 9/10 user is told to set up a screen lock, does, comes back, and is still stuck forever — where the old code showed them a working prompt, because `BiometricPromptHandler` already makes the API split.

`promptAuthenticators(sdkInt)` now asks what the prompt can actually offer, mirroring `BiometricPromptHandler` exactly. Side effect worth knowing: **this re-enables the biometric-lock toggle on Android 9/10 devices with an enrolled fingerprint**, where it was previously disabled for everyone.

## Testability seam

`canAuthenticate()`/`hasHardware()` move onto an `AuthCapability` interface in `domain/repository`. `BiometricHelper` is a final class over `BiometricManager` and `:app` carries no mocking library by design, so the seam had to be an interface for the launch-path decision to be testable at all. `SettingsViewModel` depends on the interface too.

## Verification

- `testFossDebugUnitTest --rerun-tasks` — **230 tests, 0 failures** (up from 214). Eight new `SecurityViewModel` cases — incapable device, recover-and-return, stale-error clearing, no eviction of an unlocked session, no gate while the lock is off, plus the three covering the disarm: that it happens and is *written*, that it is announced, and that a lock the user could enrol their way into is left alone. `PromptAuthenticatorsTest` pins the version split and now `enrolmentCanFixLockout` across 28–36.
- `Build.VERSION.SDK_INT` reads 0 under the mockable android.jar, so every case in `SecurityViewModelTest` runs the API-28-shaped branch — which is also why `FakeAuthCapability.hardware` defaults to `true`, or the pre-existing cases would have silently switched branch.
- `lintFossDebug` and `lintStoreRelease` — clean (`warningsAsErrors = true`); all new strings present in all five locales
- `assembleFossDebug` — builds

## Still outstanding — needs a device

The restore path has not been exercised end to end: `bmgr backupnow` with the lock on → `adb uninstall` → `adb install -t -r` → launch on a device with nothing enrolled. **`pm clear` + relaunch does not test this** — Auto Backup restores only at package *install* time.


trinadhthatakula:
Review findings applied in 972dec25, f42b480e.

**[Major] The Android 9/10 recovery path** — correct, and it was two problems cut at different versions.

*Android 10 is fixable, and is fixed.* `BIOMETRIC_STRONG or DEVICE_CREDENTIAL` is not a combination androidx will **answer** below R, but Q's prompt does **take** a credential — via `setDeviceCredentialAllowed`, the deprecated ancestor of that flag. The prompt now sets it on 29, and capability asks `BIOMETRIC_WEAK or DEVICE_CREDENTIAL`, the one credential-bearing combination supported at every level. Not a weakening: the Q prompt still accepts a strong biometric or the credential, and Android refuses to enrol a biometric without a backup credential, so a credential is implied by every enrolment anyway.

*Android 9 is not fixable* — API 28's framework prompt has no device-credential path at all — so it gets a different sentence rather than a false one ("enrol a fingerprint; a screen lock won't do it here"), chosen from the same version boundary the prompt is built on.

| API | prompt | asked of BiometricManager |
|-----|--------|---------------------------|
| 30+ | `setAllowedAuthenticators(STRONG or CREDENTIAL)` | `BIOMETRIC_STRONG or DEVICE_CREDENTIAL` |
| 29 | `setDeviceCredentialAllowed(true)` | `BIOMETRIC_WEAK or DEVICE_CREDENTIAL` |
| 28 | biometric-only, Cancel button | `BIOMETRIC_STRONG` |

`promptAuthenticators` and the new `promptAcceptsDeviceCredential` have to live in separate files because lint's `NewApi` only accepts a literal `SDK_INT` comparison as a guard for the API-29 call — so the test now pins the *invariant* rather than the versions alone: the levels that ask about a credential are exactly the levels whose prompt accepts one. Drift between those two halves is what produced this bug in the first place.

**Residual, documented, and the owner's call:** an Android 9 device with no biometric hardware still has no way in. Nothing in system Settings satisfies a biometric-only prompt, so only options 1–3 in the follow-up (fail open / self-heal / in-app disable) would reach that user, and each is a security decision this PR deliberately doesn't take.

**[Minor] `biometric_unavailable_message` wording** — the message is now selected by what the prompt on this API level accepts, so neither variant implies a remedy that can't work. Both are translated across ar/es/fr/zh-rCN.

**[Minor] doc wording ×2** — `Unavailable` is now described as "nothing enrolled that the prompt on this API level accepts" in `AuthState`, the follow-up, and the acceptance criteria; the device repro now requires **no accepted biometric *and* no screen lock**, with the reason spelled out (on API 30+ either one alone still unlocks, and a device with a fingerprint enrolled has a screen lock by construction).

`testFossDebugUnitTest --rerun-tasks lintFossDebug assembleFossDebug`: **BUILD SUCCESSFUL**, 218 tests, 0 failures, 0 skipped.

trinadhthatakula (commented):
## Code review summary

The API 29 compatibility split is sound, and the targeted tests pass. One supported-device lockout remains.

### Finding — Android 9 without biometric hardware still has no recovery

`BiometricUnavailableScreen` always directs the user to Security Settings. On API 28 the prompt is biometric-only, so a device without usable biometric hardware can never make `AuthCapability.canAuthenticate()` return true. The only remaining path is to exit and clear Thor data.

**Suggested change:** Add an owner-approved recovery for this impossible state, such as a narrowly scoped in-app reset/disable path. If that security decision remains deferred, do not mark the hard-lockout acceptance criterion as fully fixed across the supported API range.

### Verification

- Reviewed commit `f42b480e76d002b7babe64dad9ed2aeafc6c35a6` in an isolated clone.
- `SecurityViewModelTest` and `PromptAuthenticatorsTest` passed.
- AndroidX 1.4.0-alpha02 confirms the API 29 authenticator combination used here is supported.
- GitHub checks are green.

### Verdict

**Changes requested** — the API 28 no-hardware path remains a hard lockout.

trinadhthatakula:
Taking the second half of the suggestion — the first half is the owner's, so it stays theirs.

> If that security decision remains deferred, do not mark the hard-lockout acceptance criterion as fully fixed across the supported API range.

Done, in `4f55f80d`. Status, the `docs/follow-ups/README.md` row, the PR body and the first acceptance criterion now read **"fixed on API 29+; still a lockout on API 28 with no enrollable biometric."** The criterion is marked *partly* done rather than struck through, with the distinction spelled out: on Android 9 without enrollable hardware `BiometricUnavailableScreen` is a way out of the retry **loop** but not a way **in**, and being honest with that user is not the same as unsticking them.

**One correction to my own text while I was there.** The doc said the residual was "genuinely unfixable without the owner's call". That overstated it — the platform does have a credential path on API 28, and I have recorded it as option 5:

`KeyguardManager.createConfirmDeviceCredentialIntent()` (API 21, deprecated at 29 in favour of `BiometricPrompt`) shows the system PIN/pattern/password screen and returns a result. It is precisely what androidx's `BiometricPrompt` launches for the credential path on 28–29, which is why `BIOMETRIC_WEAK or DEVICE_CREDENTIAL` is a combination androidx will answer at *every* API level. Thor calls the framework prompt rather than androidx's because androidx needs a `FragmentActivity` and `HomeActivity` is a `ComponentActivity` — so the direct intent, behind an activity-result launcher on the unavailable screen, is the cheaper of the two routes.

I did not build it, because it is still a decision, just a smaller one than 1–3: it does not open the lock, it makes a **PIN satisfy** it. API 30+ already accepts the credential, so it is version parity rather than new policy — but it changes what unlocks Thor on a shipped OS version for someone who enabled the lock when only a fingerprint would do. And it only helps a device that *has* a screen lock; one with neither a credential nor biometric hardware still needs option 1, 2 or 3.

Also merged `dev` in (`69661fcb`) — the conflict was one import line in `ViewModelTestDoubles.kt`, where `dev` added `AppShortcutController` next to this branch's `AuthCapability`. `assembleFossDebug testFossDebugUnitTest --rerun-tasks` on the merge → **BUILD SUCCESSFUL, 226 tests, 0 failures**.

⚠️ `build-and-test` will be red until #298 lands: AGP `9.4.0-alpha07` published today and lint's `AndroidGradlePluginVersion` check (severity `ERROR`) now fails `lintStoreRelease` on `gradle/libs.versions.toml:6` for every open PR, this one included.

trinadhthatakula (commented):
## Independent code review

Reviewed the complete diff at commit `8a79a2c65a8b36dc6139827d0bebddde7dc52ffa` in a fresh isolated clone.

### Summary

The change adds an explicit unavailable authentication state, keeps capability and prompt authenticators aligned across API 28, 29, and 30+, refreshes capability after returning from system settings, and adds focused state-machine and authenticator-boundary tests. The current documentation also accurately limits the recovery claim: API 29+ has a working route back to authentication, while API 28 without enrollable biometric hardware remains a documented policy decision rather than being presented as fixed.

### Findings and suggestions

No blocking code findings at this head.

Non-blocking: complete the documented physical-device restore-path test before release, particularly API 28 and API 29 behavior. That coverage depends on framework prompts and install-time Auto Backup behavior that JVM tests cannot prove.

### Verification

### Verdict

**Ready to merge from code review perspective.** Device acceptance testing remains recommended before release.

trinadhthatakula (commented):
## Updated review summary

Re-reviewed the current head `5dfe3d83da5c56e8512d6fc4deb65c1671ad70ea` after it advanced from the previously reviewed SHA. The new merge contributes only the repository dependency updates for Material 3 `1.5.0-alpha25`, Koin plugin `1.1.0`, and Navigation 3 `1.1.5`; it does not alter the biometric-lockout implementation.

The original change still provides a recoverable unavailable-authentication state, API-aware prompt/capability behavior, security-settings guidance, lifecycle re-checking, and focused tests. I found no new correctness issue in the current delta.

Validation: `git diff --check`, GitHub CI/CodeQL, and `./gradlew testFossDebugUnitTest --rerun-tasks` all pass at the current head.

## Verdict

**Ready to merge.** Physical-device coverage across API 28, 29, and 30+ remains recommended because the behavior crosses framework biometric/credential implementations, but this is not a merge blocker.

_Note: GitHub does not allow the PR author to approve their own PR, so this approval verdict is posted as a review comment._

trinadhthatakula:
Owner's decision on the API 28 residual is implemented in `4669b7c0` — the PR body is rewritten to match, and it now **closes** the follow-up rather than parking it.

The instruction was scoped to "API 28 with no enrollable biometric", so it ships asymmetrically:

- **Refusing to arm** applies to any device where `canAuthenticate()` is false. Refusal costs the user nothing — they are in Settings and can go enrol — and `SettingsScreen` drops `enabled = state.canUseBiometric` so the tap actually reaches the view model and can be answered with a toast. A disabled row swallows it.
- **Disarming** is confined to `enrolmentCanFixLockout(SDK_INT, hasHardware) == false` — API 28 with no sensor. Anywhere else, enrolling is the fix and `AuthState.Unavailable` already takes the user there; turning off a lock they could still open would be a security downgrade dressed up as a bug fix.

230 tests / 0 failures, `assembleFossDebug`, `lintFossDebug` and `lintStoreRelease` all clean on AGP 9.4.0-alpha07.

Two things worth a reviewer's eye specifically:
1. The asymmetry above — it is my reading of a one-sentence instruction, and the wider reading (disarm on any incapable device) is the one I deliberately did **not** take.
2. The disarm runs from `SecurityViewModel.init` off the preference *flow*, not a one-shot check, because the restored `true` arrives after `_biometricEnabled` has already seeded `false`.

Still needs a device for the end-to-end restore path (`bmgr backupnow` → `adb uninstall` → `adb install -t -r`); `pm clear` cannot substitute, since Auto Backup restores only at install time.

trinadhthatakula:
Addressed both actionable comments from the latest review, in `4a57fe4f`.

**1. The refusal toast named a fix that would not work on API 28.**

Correct, and worse than "vague": the old copy said "set up a screen lock or enroll a fingerprint" on every device, but the API 28 prompt is biometric-only (`promptAuthenticators(28) == BIOMETRIC_STRONG`). A user there who follows the advice sets a screen lock, comes back, and is refused again by the identical toast.

Added `biometricRefusalMessage(sdkInt, hasBiometricHardware)` in `BiometricPromptHandler.kt`, next to the two predicates it reads — deliberately, since the original drift happened when the "what the prompt accepts" and "what we tell the user" halves lived in different files. Three branches:

| device | copy |
|---|---|
| `!enrolmentCanFixLockout` — API 28, no sensor | `biometric_lock_unavailable_toast` — the lock cannot be armed here at all |
| `!promptAcceptsDeviceCredential` — API 28, sensor present | `biometric_not_enrolled_toast_biometric_only` — a fingerprint is required; a screen lock will not do it here |
| API 29+ | the existing `biometric_not_enrolled_toast` — either one works |

The first case is the point: on that device there is no Settings page to send anyone to, so the toast stops offering one. Both new strings ship in `values`, `values-ar`, `values-es`, `values-fr`, `values-zh-rCN`.

`theRefusalNamesSomethingTheUserCanActuallyDo` in `PromptAuthenticatorsTest` pins all three branches across API 28–36 × hardware true/false.

**2. The doc called the shipped fix "option 5", which was already taken.**

Fixed. The Resolution section now names it descriptively; option 5 stays the `createConfirmDeviceCredentialIntent` route defined in the residual section, which is still not taken.

Also extended the doc's "refusing to arm is universal" bullet to record which toast goes to which device and why, so the next reader does not have to derive it from the `when`.

`testFossDebugUnitTest --rerun-tasks`, `assembleFossDebug`, `lintFossDebug` and `lintStoreRelease` are all green locally.

PR #293, opened 2026-07-30
feat(export/backup): .xapk export (#164) and bulk app backup with a manifest (#51 phase 1)

### LINKED ISSUE #164: .APK, .XAPK, .APKS and Split bundle export?
We kinda already have the APK share function, but I would like to be able to do that, but without sharing, more like exporting to folder/location. Also I think it would be nice to export not only to APK but also to XAPK, APKS and Split bundles.

trinadhthatakula:
Closes #164. Advances #51 (phase 1 of 3).

Two commits, paired on purpose — backup phase 1 is a batch wrapper over the same bundle builder the `.xapk` writer touches, and `docs/follow-ups/app-data-backup-and-xapk-export.md` records the owner's call that they ship together.

## 1. `.xapk` export (`77bfc027`)

`BundleFormat` (`APK`/`APKS`/`XAPK`) now owns the format policy that used to be a single `splitPublicSourceDirs.isEmpty()` branch in the builder, and carries the MIME that decides whether a receiver treats the file as installable.

The sheet offers exactly **two** chips: the app's native container plus `.xapk`. The third is always the wrong offer — `.apks` around a single base APK is a zip that buys nothing, and a monolithic `.apk` of a split app drops the config splits and installs something that will not run. `autoFor()` never returns XAPK, so an export nobody touches produces what it always did.

Three defects fixed in the same code:

- Single-app **share** hardcoded `application/vnd.android.package-archive` even for a `.apks` zip, so an installer accepted the share and then choked on the contents.
- **Nothing ever deleted the staged bundle.** Export frees its own copy in a `finally`; the builder clears its staging dir when a build fails or is cancelled. Share still cannot — its `content://` URI outlives the call.
- `split_apks[]` was built from `AppInfo`, so a split that failed to copy was still *declared*. It is now built from the entries actually staged.

## 2. Backup phase 1 (`0ef42564`)

Multi-select → **Backup** runs the selection through the *existing* `ExportAppUseCase` — the only thing that deletes its staged copy after a successful write, which is the single property keeping `cacheDir` bounded across a 200-app run — and drops `thor-backup-<yyyyMMdd-HHmmss>.json` beside the bundles listing every app attempted, failures included.

`BackupRunner` is a `@Single` on a process-lifetime scope (the shape `BulkFreezeRunner` already uses): the run outlives the sheet, the view model and often the Activity, **without a foreground service** — which would mean a new permission, a mandatory FSU type on API 34+, a time cap on 35+ and a class of start-not-allowed crashes. Two apps stage at once, gated by a semaphore owned by the *runner* rather than the run. A non-modal bar in the bottom bar reports it and can stop it; a cancelled run still writes its manifest.

### Decisions worth reviewing on their merits

| Decision | Why |
|---|---|
| **One manifest per run, timestamped** | Both file-store paths write by name and delete a collision first, so a fixed `thor-backup.json` would let Tuesday's export silently replace Monday's manifest while leaving Monday's bundles undescribed. Readers glob `thor-backup-*.json` and merge. |
| **Failures recorded in the manifest, same flat shape** | `error` set, `fileName`/`sizeBytes` null. No polymorphic discriminator, so a `jq` one-liner or a v1 restore tool does not have to learn Kotlin's `"type"` convention. |
| **`schemaVersion` written even when it equals its default** | It is the one field a foreign reader needs before it can parse the rest. A test pins that a v1 reader survives a v2 document, so phase 2 can add a per-entry data file without breaking a phase-1 folder. |
| **Non-modal progress bar** | The confirm dialog tells the user to keep Thor open, not to sit and watch it. A blocking dialog would contradict its own copy. |
| **Pre-flight against `StorageManager.getAllocatableBytes`** | `File.usableSpace` under-reports on Android by ignoring the clearable cache the platform would evict, so it would refuse batches the device could run. Measured in `data/` and passed in as a `Long`, so the domain stays free of `Context` — and an unmeasurable volume **fails open**. |

### Four self-review findings fixed before this PR

Each produced a wrong sentence in front of the user, and each now has a test:

1. **"N of M apps were saved" counted attempts, not files.** A cancel after eight attempts of which three failed saved five. `BackupProgress` now carries `completed` (drives the bar) and `saved` (drives the sentence) separately.
2. **The saved count was read from the shared progress flow** — which a replacement run zeroes on its way in, so a cancel-by-replacement reported 0 while the folder held five. Now a per-run local captured inside the `async`.
3. **A replaced run reported "Export stopped".** It is cancelled exactly like an explicit cancel, but the replacement is about to report for the same UI. A `superseded` flag keeps it quiet.
4. **Three taps stranded the first run.** B parked in `A.cancelAndJoin()`, B's own cancellation made that join throw, and A carried on staging into the same cache as C. The handoff is now `NonCancellable`, so waiting for B transitively waits for A. That wait is unbounded by construction, so the builder's copy and zip loops check for cancellation **per 8 KB chunk** — promptness has to come from the work, not a timeout.

Plus: an unexpected failure now emits `BackupRunResult.Failed` rather than logging and vanishing. Nothing awaits the run in the common case, and a progress bar that disappears with nothing said reads as success.

## Verification

- `./gradlew testFossDebugUnitTest --rerun-tasks` → **255 tests, 0 failures** (up from 209 on `dev`). New: `BackupRunnerTest` (5), `BackupAppsUseCaseTest`, `BackupIndexTest`, `BundleFormatTest`, plus `MainViewModelTest` coverage for the replay/consume handshake.
- `./gradlew lintFossDebug assembleFossDebug` → **BUILD SUCCESSFUL**, clean under `warningsAsErrors`. All 6 new strings exist in all 5 locales.

## Left to the owner — device work the desk cannot do

Both are recorded in `docs/follow-ups/app-data-backup-and-xapk-export.md`; **neither is a regression risk to this branch**, and a green install hides both:

1. **Missing OBB.** A real XAPK carries `Android/obb/<pkg>/*.obb` and Thor exports none. Export an app that has one, install, launch — it will install cleanly and fail at play time. The explain string says so up front; assert the absence deliberately.
2. **`manifest.json` against a third-party reader.** Thor writes it for its own consumption, so nothing has validated it against SAI or APKPure. Round-trip a real APKPure `.xapk` in both directions.

## Not done here

- **#51 phase 2** (root-only data tar) — a separate branch behind an explicit root gate. `README.md`'s "BackUp App Data" promise stays half-kept until it lands, so the README is deliberately unchanged.
- **Closing #164 on GitHub** — the issue is the owner's to close.

Docs updated: `docs/feature-request-roadmap.md` (#164 → done, #51 → phase 1 done) and the follow-up narrowed to phase 2 + the two verifications.

## Summary by CodeRabbit

* **New Features**
  * Added multi-app bulk export with an always-visible global progress bar, stop/cancel controls, and bulk completion summaries.
  * Added export format selection (APK/APKS/XAPK), including improved XAPK packaging and icon handling.
  * Bulk backup/export now writes per-run timestamped backup index manifests with per-app outcomes.
* **Bug Fixes**
  * Improved cancellation responsiveness during large exports and zip/copy work.
  * Fixed bundle naming and split-app metadata to prevent duplicate/incorrect ZIP entries.
* **Documentation**
  * Updated roadmap and follow-up notes for backup/export behavior.
* **Localization**
  * Added/updated bulk export and format-selection UI strings across supported languages.


trinadhthatakula:
Review findings applied in 8aae224a, 60b1537c, 9a87ba31.

**Manifest name collision** — timestamp is now `yyyyMMdd-HHmmss-SSS`. Same-second isn't hypothetical: cancel-and-replace is the ordinary way two runs finish close together (the cancelled run writes its manifest under `NonCancellable` while the replacement is already going), and both file-store paths delete a collision before writing. Two tests added, one of them pinning that the name still sorts chronologically as text across a second boundary.

**`consumeCompletion()` in a plain `finally`** — the reported *scenario* doesn't hold (if the first send never landed, the outcome was never shown and replaying it is correct), but the bug is real one branch over: `Finished` + `!indexWritten` sends **twice** over a rendezvous channel, so a ViewModel cleared between the two deliveries left the outcome unconsumed and the next instance replayed the whole thing, support prompt included. Consuming unconditionally would have traded that for the opposite failure — silently dropping an outcome nobody saw. It's now guarded on at least one message having landed, still in `finally` and still non-suspending. Reporting was split into a pure `exportReport` so the acknowledge-once rule lives in one place rather than five `when` arms.

**A third fault the new test caught:** `_effect` was declared *below* `init`. `viewModelScope` runs on `Dispatchers.Main.immediate`, so the collector `init` launches starts inline, and a replayed completion reaches it *during* the constructor with the channel field still null — i.e. reopening Thor after an export that finished with no UI attached NPE'd on the main thread inside `viewModelScope`, where nothing handles it. Declaration moved above `init` with a comment saying why it has to stay there.

**Docs** — "process-lifetime" now states plainly that it means surviving Thor's *UI*, not its *process*, with the manifest named as the deliberate and only recovery and the two rejected alternatives costed. #164 is marked closeable *with its scope stated*, and "APK-only" reads "bundles only, no app data" everywhere.

`testFossDebugUnitTest --rerun-tasks lintFossDebug assembleFossDebug`: **BUILD SUCCESSFUL**, 258 tests, 0 failures, 0 skipped.

trinadhthatakula (commented):
## Code review summary

The new XAPK and batch-export tests pass, but the batch path can report success while losing or misplacing bundles.

### Finding 1 — Same-name apps overwrite each other

Output filenames contain only the app label and version. Two distinct packages with the same label and version therefore write the same destination name. Both file-store implementations overwrite by name, yet the run counts both writes as successful and records two manifest entries pointing to one file.

**Suggested change:** Make batch filenames package-unique, or allocate unique names before building. Add a regression test with two packages sharing a label and version, using a fake store that models overwrite-by-name.

### Finding 2 — The destination can change during a run

Each `ExportAppUseCase` call re-reads `exportDirUri`, and `writeIndex` reads it again. Because the batch UI is non-modal, the user can change the destination while the run is active. A tree can also become invalid mid-run. Bundles may then be split across folders while the single manifest describes only the current destination.

**Suggested change:** Resolve one immutable destination at batch start and use it for every bundle and the manifest.

### Finding 3 — Single export races batch staging

All exports of a package share `export_temp/<packageName>`. A simultaneous single-app export can delete the directory or staged file used by the batch worker.

**Suggested change:** Use a unique run-scoped staging directory per export, or serialize all exports for the same package with a shared mutex.

### Verification

- Reviewed commit `9a87ba315ec76ee940f94927de3307edd867a116` in an isolated clone.
- Backup, runner, index, format, metadata, and MainViewModel tests passed.
- GitHub checks are green.
- Existing test fakes use package-derived filenames and do not model overwrite semantics, so they miss Finding 1.

### Verdict

**Changes requested** — the batch export currently has data-integrity risks.

trinadhthatakula:
Latest review pass applied in `c79bcbda`.

**1. A null `versionName` rendered as the literal `null`** (Minor, CodeRabbit)

`StringBuilder.append(Any?)` writes the four letters `null`, so an app that declares no version exported as `MyApp_null.apk`. The version segment is now dropped entirely when it is absent, and blank counts as absent — a trailing underscore is the same defect wearing less. Covered by `an app that declares no version is not named after the word null`, which pins the null case, the blank case, and that the package discriminator still applies on top.

**2. Two source paths with the same leaf name collapsed into one entry** (Minor, CodeRabbit, outside diff)

Real, and the consequence is worse than a lost split: the same `File` landed in `apkFiles` twice and `zipFiles` threw `ZipException: duplicate entry`, failing the export of an app whose APKs were every one of them readable.

Fixed with `stagedApkNames()`, which settles every destination name before the first byte is copied:

- an **exactly repeated path** is dropped — `publicSourceDir` turning up again in `splitPublicSourceDirs` is the realistic way this collides, and it is the same bytes, so staging it twice would double the zip for nothing;
- a **genuine collision between two different paths** is renamed (`base.apk`, `base_2.apk`) rather than dropped, because those are two different APKs and a bundle missing one will not install. Nothing reads a split's file name back — the installer takes the split's identity from its manifest, and the `metadata.json`/`manifest.json` sidecars are generated from these names.

Comparison is exact rather than case-insensitive here, unlike `uniqueBundleFileNames`: this names files in the app's own cache dir on ext4 and entries in a zip, both of which tell `Base.apk` from `base.apk`. The SAF destination — which may not — is the other function's problem.

Seven tests in `StagedApkNamesTest`, including the hidden-file edge (`.apk` is a whole name, not an extension on an empty stem).

**3. Roadmap claimed this branch had merged** — corrected to "In review (#293)" on both rows.

`assembleFossDebug testFossDebugUnitTest --rerun-tasks` → **BUILD SUCCESSFUL, 273 tests, 0 failures**.

---

⚠️ **`build-and-test` is red for a reason that has nothing to do with this branch.** AGP `9.4.0-alpha07` published today and lint's `AndroidGradlePluginVersion` check — which ships at `ERROR` — now fails `lintStoreRelease` on `gradle/libs.versions.toml:6`, a line this PR does not touch. Every open PR went red at the same moment. #298 fixes it in `app/lint.xml`; this branch needs `dev` merged in afterwards to go green.

trinadhthatakula (commented):
## Independent code review

Reviewed the complete diff at commit `5b6c8e54b157db7fcd434271e8e97154f7c8e2bf` in a fresh isolated clone. The earlier batch filename, destination, and staging-scope findings are fixed, but three blocking issues remain.

### 1. Incomplete split bundles are reported as successful

`AppBundleBuilderImpl.kt:87-99` uses `mapNotNull` and only fails when every APK copy fails. If the base or any required ABI/density split fails to copy, the builder still emits an APKS/XAPK, `ExportAppUseCase` reports success, and backup records a successful entry. Such a bundle is not a faithful backup and can be un-installable or unusable after install. Filtering the manifest to match the damaged archive makes its description internally consistent but does not make the app restorable.

**Suggested change:** make staging all-or-fail for the base plus every installed split. On any failed copy, delete staging and return failure. Add regressions for a failed base copy and one failed split copy.

### 2. Renamed collision entries are not reflected in XAPK manifest metadata

`stagedApkNames` correctly renames different source paths with the same leaf to names such as `base_2.apk`, but `ApksMetadataGenerator.xapkManifest` reconstructs entries from the original leaf names and only filters by a set of staged names (`ApksMetadataGenerator.kt:124-145`). For the collision case the manifest can list `base.apk` twice and never name the actual `base_2.apk` zip entry. This undermines the collision fix for XAPK readers.

**Suggested change:** pass the full source-path to staged-name mapping into metadata generation and build `split_apks` from that mapping. Add an integration test covering two different source paths with the same leaf.

### 3. Stop cannot interrupt the final destination copy

The builder copy and zip loops check cancellation, but `AppBundleFileStoreImpl.kt:45-58,84-122` still uses blocking `copyTo` / `File.copyTo` for the equally large write to SAF, MediaStore, or legacy Downloads. A Stop or replacement run that arrives there waits for the entire multi-gigabyte copy before the old run can unwind, contrary to the cancellable progress UI and the runner handoff assumptions.

**Suggested change:** use a cancellation-aware buffered copy in all destination writers, deleting the partial `DocumentFile`, MediaStore row, or legacy file on cancellation. Add a test with a blocking destination stream to prove Stop ends the write promptly.

### Verification

- `git diff --check` — passed

### Verdict

**Changes requested.** The first two findings can produce backups that are recorded as valid but are not faithfully restorable; the third breaks the user-visible cancellation contract.

trinadhthatakula (commented):
## Updated review summary

Re-reviewed current head `958c14ac22fb8f3f0432f0ce90bcec2464a2a0dd`. Its only delta after the previously reviewed SHA is the shared dependency merge (Material 3 `1.5.0-alpha25`, Koin plugin `1.1.0`, Navigation 3 `1.1.5`), so the export/backup implementation and the three blocking findings remain unchanged.

## Blocking findings

1. **[P1] Fail the bundle when any required APK cannot be staged** — `AppBundleBuilderImpl.kt:91-97` uses `mapNotNull` and succeeds when at least one APK copies. A failed base or split copy therefore produces and reports a partial bundle as a successful backup. Require every deduplicated source APK to stage, and clean up/fail the app export if any copy fails.

2. **[P1] Preserve collision-renamed APK names in XAPK manifest entries** — `stagedApkNames()` can rename a genuine leaf collision to names such as `base_2.apk`, but `ApksMetadataGenerator.xapkManifest():127-145` reconstructs declarations from the original paths and merely filters by the staged-name set. The renamed APK remains in the zip but disappears from `split_apks`, so readers that trust the manifest omit a required split. Carry the resolved name and split identity together into manifest generation.

3. **[P1] Make destination copies cooperatively cancellable** — `AppBundleFileStoreImpl.kt:52-54,104-105,122` uses blocking `copyTo` for SAF, MediaStore, and legacy Downloads writes. Cancelling Stop or cancel-and-replace cannot interrupt a large destination write, so the old run can continue occupying I/O and delay handoff. Use a chunked copy loop with coroutine cancellation checks, while retaining the current partial-output cleanup.

Validation: `git diff --check`, GitHub CI/CodeQL, and `./gradlew testFossDebugUnitTest --rerun-tasks` pass at the current head.

## Verdict

**Changes requested.** The green suite does not cover these partial-copy, name-collision, and in-progress destination cancellation paths.

_Note: GitHub does not allow the PR author to submit a formal Request changes review, so this verdict is posted as a review comment._

trinadhthatakula:
All three fixed in 1cbbc923. They turned out to be one theme: every one of them produced an export that *looks* like it worked.

## 1. A split that would not copy was dropped, not failed

`mapNotNull` + `isEmpty()` meant "at least one APK made it" counted as success. An app whose ABI split sits on a mount the shell cannot read got exported without it, and the user finds out when they restore — a bundle that installs missing its resources, or refuses to install at all.

There is no partial backup of a split app; an `.apks`/`.xapk` is only meaningful if it holds the whole set. Every deduplicated source APK must now stage or the build throws, and the existing `catch` wipes the staging dir with it:

```kotlin
val apkFiles = plan.map { (path, name) ->
    val destFile = File(tempSplitDir, name)
    if (!copyFileSafely(path, destFile)) {
        throw IllegalStateException("Failed to copy APK: $name")
    }
    destFile
}
```

## 2. A renamed APK vanished from the manifest

Exactly as you described. `stagedApkNames` disambiguates `base.apk` → `base_2.apk`, the manifest was rebuilt from the *source* paths and then filtered by the staged-name set, so `base.apk` was named twice, `base_2.apk` never, and the zip carried an APK `split_apks` did not mention.

Name and identity now travel together:

```kotlin
data class StagedApk(val sourcePath: String, val entryName: String)
```

The builder passes those pairs; `xapkManifest` builds `split_apks` straight from them, so the manifest describes the *container* rather than the app. The split `id` is still derived from the source leaf — the rename is a container detail, the id is what an installer matches the split on. `xapkManifest_namesARenamedApkByWhatTheZipCallsIt` pins the exact case (`base.apk` + a colliding split → `["base.apk", "base_2.apk"]`, id still `base`).

The `null` branch is unchanged in meaning and now says so: no list is a claim about the app, not about what was staged.

## 3. Writing the bundle out was not cancellable

All three sinks used the stock `copyTo` — one uninterruptible call over a bundle that is routinely hundreds of megabytes, so cancelling did nothing until the whole thing had been pushed through SAF or MediaStore. Now chunked with an `ensureActive()` per 8 KB, mirroring `AppBundleBuilderImpl.copyCancellable`, which already did this for the staging copies.

Cleanup: the SAF and MediaStore paths already deleted their partial output, and since `CancellationException` is an `Exception` those `catch` blocks now cover cancellation too — worth stating in a comment rather than leaving as a happy accident. The legacy Downloads path had **no** cleanup, because there was nothing to clean up while the copy could not be interrupted; it has one now. A truncated `.apk` sitting in Downloads under the right name is the one failure shape a user cannot tell from a good export.

---

284 tests, 0 failures. `lintFossDebug` / `lintStoreRelease` clean, `assembleFossDebug` builds.

Still unexercised on a device, and noted as such on the PR: reading a Thor-produced `.xapk` back with SAI/APKPure, and the OBB path.

PR #294, opened 2026-07-30
feat(app-list): filter the app list by permission (#285)

### LINKED ISSUE #285: [Feature]: Filters by App Permission
### Pre-flight checklist

- [x] I searched existing issues and discussions and didn't find a duplicate.

### Problem / motivation

I wanted to see which app use a certain permission. For example, what are the apps that use camera permission?

### Proposed solution

Make a filter to group apps by permission

### Area

Permissions

### Alternatives considered

_No response_

trinadhthatakula:
Closes #285.

A third `FilterType` alongside Source and State: chips for Camera, Microphone, Location and the rest, each listing the apps that declare a permission in that group.

## The scope question the roadmap held this behind

> *"where the data comes from … this likely wants a permission index in the Room cache, populated on scan — which turns a UI feature into a schema change."*

**Answered: no.** `getInstalledPackages(GET_PERMISSIONS)` returns every package's declared permissions in a *single* binder call, so the whole device is indexed in one sweep. `PermissionRepository.buildPermissionIndex()` runs it while the filter is selected and keeps `group -> packages` in memory on the ViewModel. Persisting it would have been worse than useless — any install, uninstall or update invalidates the index, so Room buys a migration and a staleness bug in exchange for nothing.

## ⚠️ The part that actually needs review

**`PermissionInfo.group` is vestigial, and this feature was built on it before review caught that.**

Since API 29 the framework manifest declares *every* dangerous platform permission with `android:permissionGroup="android.permission-group.UNDEFINED"`; the real mapping moved into PermissionController's own hardcoded table. Verified on an API 37 emulator:

```
$ adb shell pm list permissions -g -d
group:android.permission-group.CAMERA        (empty)
group:android.permission-group.MICROPHONE    (empty)
group:android.permission-group.LOCATION      (empty)
…all 47 runtime permissions sit under UNDEFINED
```

So the obvious implementation produces **no Camera chip, no Microphone chip and no Location chip on any device from the last six years** — only whatever custom groups third-party apps still declare. It would have shipped looking like it worked, on an emulator, for the reviewer who only had third-party apps installed.

`PlatformPermissionGroups` mirrors PermissionController's `Utils.PLATFORM_PERMISSIONS` and is consulted first. `PermissionInfo` is still the right answer for **custom** permissions — that is where the field is still honest, and the protection level is the only way to know the permission is dangerous at all. A permission in neither is dropped rather than guessed.

The group *names* are real even where the mapping is not, so `getPermissionGroupInfo(...).loadLabel()` still resolves each to the label the system's own permission dialog uses. **The chips read exactly like the prompts the user has already seen, in their language, with Thor shipping zero translations of "Camera".**

Since a hardcoded string table gets no help from the compiler, `PlatformPermissionGroupsTest` (11 tests, pure JVM) asserts key/value shape, that UNDEFINED is never produced, that the headline six resolve, that the tiered location and granular media permissions land where the system prompts for them, that non-dangerous and custom permissions fall through, and pins the table size so a rebase cannot silently eat a line.

## Three things the sweep gets right that are easy to get wrong

| | |
|---|---|
| **`MATCH_UNINSTALLED_PACKAGES \| MATCH_DISABLED_COMPONENTS`** | The same universe `AppRepositoryImpl` scans. Thor freezes system apps with `pm uninstall --user N`, so a frozen system app is not installed for this user; a default sweep drops it and every chip then quietly omits exactly the apps Thor exists to show. |
| **`ensureActive()` in the package loop** | Plain blocking work with no suspension point of its own, so without it a torn-down sweep runs to completion on its IO thread and overlaps its replacement, each paying a full set of binder calls. `CancellationException` is rethrown ahead of the catch-all, so switching filters cannot surface as "Couldn't read permissions". |
| **Rebuild on install/uninstall/update** | Keyed on `packageName@lastUpdateTime`, derived from the app list Thor already refreshes — no `PACKAGE_ADDED` receiver, no polling. A package the index has not heard of matches *no* group, so a freshly installed camera app would otherwise be missing from the Camera chip with nothing to indicate why. The key stays put for freeze, suspend and a size arriving, so a session that installs nothing pays for one sweep. |

## Two UI states that were saying the wrong thing

- **While the index builds**, the list showed "No matching apps found". There is no index yet, so everything is filtered out — a pending answer rendered as a wrong one. `isLoadingPermissions` now feeds `isLoading`. (During a *rebuild* the previous index stays in state, so nothing flashes.)
- **A failed build** was indistinguishable from an empty one. The toast is long gone by the time the user looks at the empty chip row, and "no permission groups found on this device" would be a lie. `permissionIndexFailed` is its own state and the row says so. The status line is also no longer `maxLines = 1` — the es/fr/ar translations are sentences and were being cut in half.

## Two things that came along because the code was already open

- **`filterApps()` lifted into domain, verbatim.** The rules are real — "Frozen" is `!enabled`, Active and Suspended deliberately overlap, an unrecognised *persisted* chip passes everything through rather than showing an empty list — and none were reachable from a test inside a private method on a ViewModel with eleven constructor dependencies. `AppFilteringTest` covers all three branches, not just the new one. Behaviour is unchanged; I diffed it against `HEAD:AppListViewModel.kt`.
- **`updateAppFilter`'s `if (State) "STATE" else "SOURCE"` is now an exhaustive `when`.** The old shape would have silently persisted SOURCE for the new filter, and for every filter added after it. Reading back falls through to Source for an unknown token, so a prefs file from a newer Thor degrades instead of failing.

## Verification

- `./gradlew testFossDebugUnitTest --rerun-tasks` → **235 tests, 0 failures, 0 skipped** (209 on `dev`; +26 from `AppFilteringTest` and `PlatformPermissionGroupsTest`).
- `./gradlew lintFossDebug assembleFossDebug` → **BUILD SUCCESSFUL**, clean under `warningsAsErrors`. All 6 new strings present in all 5 locales (`MissingTranslation` is fatal).
- The UNDEFINED behaviour was confirmed on an API 37 emulator, not inferred from docs.

## Not done here

- **`SortBy.asGeneralName()` is hardcoded English**, one tab down in the same sheet — the identical bug to the `FilterType.asGeneralName()` one this PR fixes. Left alone deliberately: 8 strings × 5 locales is 40 machine-translated entries nobody would review, in a PR about permissions. Filed as [`docs/follow-ups/sort-labels-are-hardcoded-english.md`](docs/follow-ups/sort-labels-are-hardcoded-english.md), with the note that `SortOrder.asGeneralName()` is dead code to delete rather than translate.
- **Closing #285 on GitHub** — the issue is yours to close.
- **On-device confirmation.** Everything above was verified on an emulator and in unit tests; the permission chips have not been exercised on the physical device.

## Merge note

`docs/feature-request-roadmap.md` is also edited by the open backup branch (#293). Each edits only its own rows, but the ranking table and the "In review" line will conflict textually — whichever merges second needs a rebase.

## Summary by CodeRabbit

* **New Features**
  * Added an “Permission” option to app list filtering, with chips grouped by Android permission group and labeled using localized names.
  * Added permission index status messaging in the filter UI while permission data is being prepared (loading, unavailable, or failed).
* **Bug Fixes**
  * Improved saving/restoring of the selected filter type to correctly persist and reload the new Permission option.
  * Refined filter reset behavior when switching app list categories, returning the selected filter to “All” appropriately.
* **Documentation**
  * Updated the roadmap and follow-ups to reflect permission filtering progress and related label localization work.


trinadhthatakula (commented):
## Code review summary

The filtering and grouping tests pass, but the index can expose permissions that do not exist on the current Android version.

### Finding — Future permission declarations become false capability matches

`resolveRuntimeGroup()` returns the static platform mapping before consulting `PackageManager`. `requestedPermissions` includes manifest declarations even when the current OS does not define that permission.

For example, an APK declaring `POST_NOTIFICATIONS` on API 28, or `READ_MEDIA_IMAGES` on API 32, will be placed in a Notifications or Photos group even though that permission is unknown and grants no capability on that device. This contradicts the stated rule that only current dangerous/runtime permissions are indexed and can create misleading chips across Thor’s supported API range.

**Suggested change:** Use the static table only to override the unreliable group field. Still verify that `getPermissionInfo()` succeeds on the current device and that its base protection level is dangerous. Alternatively, store the permission introduction API in the table and gate by SDK level. Add tests for future permissions on older SDK levels.

### Verification

- Reviewed commit `d1d5c4b8d6fd369d3a587818529d74f7749a3372` in an isolated clone.
- `AppFilteringTest` and `PlatformPermissionGroupsTest` passed.
- GitHub checks are green.
- The current pure table tests validate mapping shape, but they do not model whether a permission exists on the runtime OS.

### Verdict

**Changes requested** — permission chips can claim capabilities that are unavailable on the current device.

trinadhthatakula:
Finding applied in `5c57ad6b`, plus a second defect the same read turned up.

**The ordering rule is now device-first.** `resolveRuntimeGroup` shrank to the binder call; the decision moved into a pure `runtimeGroupFor(permission, declared)` in `PermissionIndex.kt`:

- `getPermissionInfo` throwing `NameNotFoundException` is the authoritative "this OS does not define it", and it is consulted **before** the table. A permission the device has never heard of is dropped, whatever the table says about it.
- The device also has to call it `PROTECTION_DANGEROUS`. The table now only overrides the *group* of a permission the device has already confirmed exists and is dangerous — which is the narrow job it was added for, since API 29+ reports `UNDEFINED` for every dangerous platform permission.
- A custom dangerous permission keeps its own honest group, exactly as before.

So an APK declaring `POST_NOTIFICATIONS` on API 28, or `READ_MEDIA_IMAGES` on API 32, now produces no chip at all. I went with the runtime check rather than an introduced-in-API column in the table: the SDK-level version is a second thing to keep correct, and it still cannot see a permission an OEM removed or a GSI never shipped.

Splitting the pure function out is what made it testable — `PackageManager` is abstract and `:app` has no mocking library. `RuntimeGroupForTest` (7 tests) covers the undefined-on-this-device case, the not-dangerous case, the table-overrides-the-group case, the custom-permission case, and the "declared later, grouped once the device defines it" case.

**Separately: the table keyed voicemail on `android.permission.ADD_VOICEMAIL`.** `Manifest.permission.ADD_VOICEMAIL` is actually `com.android.voicemail.permission.ADD_VOICEMAIL`, so that entry matched nothing any device has ever declared and the Phone chip silently skipped voicemail apps. The key-shape test that should have caught it was asserting the `android.permission.` prefix, which is exactly the assumption that was wrong; it is now an explicit exception list plus a test pinning both spellings.

Also merged `dev` in (`361638e4`) — the app list's Android-bound collaborators became domain ports there while this branch was adding a permission repository to the same constructor. The ports and injected dispatchers are kept, and `AppListViewModelTest` gets a counting `FakePermissionRepository`.

`assembleFossDebug`, `testFossDebugUnitTest` and `lintFossDebug` are green on the merge commit.

trinadhthatakula (commented):
## Independent code review

Reviewed the complete diff at commit `ee2b1f874453edc59bd042369be2e33de8c2f629` in a fresh isolated clone. The earlier current-OS existence check is fixed, but the static group mapping still has a blocking correctness gap.

### Finding — The platform mapping is already stale and is not API-aware

`PlatformPermissionGroups` is described as mirroring PermissionController, but it differs from the canonical current mapping in two user-visible ways:

1. `android.permission.RANGING` is a dangerous platform permission mapped by current PermissionController to `NEARBY_DEVICES` when the ranging flag is enabled. It is absent from this table. On a device that defines it, `declaredPermission()` correctly returns a dangerous permission, but `runtimeGroupFor()` falls back to the framework `UNDEFINED` group and drops it. Apps using generic ranging therefore never appear in the Nearby devices chip.
2. `ACCESS_MEDIA_LOCATION` is always assigned to `STORAGE`. PermissionController assigns it to Storage before API 33 and to `READ_MEDIA_VISUAL` on API 33+, matching the Photos and videos prompt. On current devices Thor puts those apps in the wrong chip. The stated double-count concern does not apply because `packagesByGroup` uses sets; mapping this permission to the current platform group does not duplicate an app within that group.

The test that pins the local table size to 42 only proves that this local snapshot did not lose one of its own entries; it cannot detect a permission added upstream, which is exactly what happened with RANGING.

**Suggested change:** make the mapping API-aware, add `RANGING -> NEARBY_DEVICES` (the existing device-existence check already gates feature availability), and switch `ACCESS_MEDIA_LOCATION` from Storage to Read Media Visual on API 33+. Add tests for both sides of the API 33 boundary and for RANGING on a device where it is declared dangerous.

### Verification

- Compared the table with AOSP PermissionController `PermissionMapping.kt` and the current framework manifest

### Verdict

**Changes requested.** The feature can omit a current dangerous permission entirely and misclassify another across much of Thor’s supported OS range.

trinadhthatakula:
Both points taken — and the second one corrects a claim I had written into the code as a justification. Fixed in c78b2a92.

**`packagesByGroup` is `Map<String, Set<String>>`.** So the "would double-count" comment defending `ACCESS_MEDIA_LOCATION -> STORAGE` was simply wrong: an app already in a group cannot be added to it twice. With the only argument for the pin gone, the pin has to go too — a single static answer is wrong on one side of the Tiramisu boundary whichever side it picks. On 33+ a photo app requesting `READ_MEDIA_IMAGES` + `ACCESS_MEDIA_LOCATION` and nothing else was landing in a Storage chip the system's own dialog would never show it under.

**`RANGING` was worse than ungrouped — it was dropped.** The fallback reads `PermissionInfo.group`, which is `UNDEFINED` for every dangerous platform permission since Android 10, so an app holding only `RANGING` left the Nearby devices chip entirely rather than landing in a wrong one. And you are right that it needs no version condition: `runtimeGroupFor` reaches the table only once the device says it *defines* the permission **and** calls it dangerous — which is exactly the pair `Flags.rangingStackEnabled()` governs. Both guards are now asserted independently.

## Shape of the fix

`groupOf` and `runtimeGroupFor` take the running API level, and a `regrouped` map carries the permissions the *platform* moves across versions (one entry today):

```kotlin
private val regrouped: Map<String, Pair<Int, String>> = mapOf(
    P + "ACCESS_MEDIA_LOCATION" to (TIRAMISU to READ_MEDIA_VISUAL)
)

fun groupOf(permission: String, sdkInt: Int): String? {
    regrouped[permission]?.let { (since, group) -> if (sdkInt >= since) return group }
    return byPermission[permission]
}
```

The level is a **parameter, not a read of `Build.VERSION.SDK_INT`**, for two reasons. The table is deliberately pure Kotlin — it exists because the device cannot be asked, so a test that asked a device would be testing the wrong thing. And `SDK_INT` reads **0** under the mockable android.jar, so a default argument reading it would have silently put every unit test on the pre-Tiramisu side of the boundary while looking like it exercised the device's answer. `PermissionRepositoryImpl` passes it in at the one call site; the per-permission memo cache is unaffected, since the level is fixed for the process.

## Tests

257 total, 0 failures (was 249). New:

- `accessMediaLocationMovesToTheMediaChipWhereThePlatformMovesIt` / `accessMediaLocationIsGroupedWhereThisDeviceWouldPromptForIt` — 28, 32, 33, 36 through both entry points.
- `rangingCountsTowardsNearbyDevicesOnceTheDeviceDeclaresItDangerous` — undefined on old *and* new devices → null; defined-but-not-dangerous → null; both → `NEARBY_DEVICES`. Each guard failing on its own.
- `everyNearbyDevicePermissionSharesOneChip` — all six, one group.
- `everyRegroupedPermissionAlsoHasABaseRow` — an override without a base row would resolve on new devices and disappear on old ones, which is the silent-drop failure this file exists to catch wearing a new hat.
- `everyPermissionInTheTableResolvesWhenTheDeviceDefinesIt`, `knownPermissionsAndGroupOfAgree` and the headline-permission pins now run at **both** API 28 and 36 rather than once, so the new parameter cannot strand an entry on one end of the supported range.
- Table-size pin 42 → 43.

`lintFossDebug` and `lintStoreRelease` clean, `assembleFossDebug` builds.

trinadhthatakula:
All three applied in `21b403a9`.

**1. Cold start with a persisted Permission filter shows the wrong sentence (actionable).**

Confirmed, and the diagnosis is exact. `AppQuickFilters` picks between three strings while `permissionIndex.isEmpty`, and the empty-app-set branch cleared `isLoadingPermissions` on its way out — so "waiting on the package scan" and "the scan finished and found nothing" collapsed into one state, and the row fell through to `permission_filter_unavailable`. A claim about the user's *device*, shown right next to the spinner that was the real reason, for every returning user who left the filter there.

Split as suggested: `type != Permission` still clears the flag, `Permission` with an empty app set now raises it. `permissionIndexFailed` is deliberately left alone rather than cleared — loading outranks it in that selector anyway, and the next emission re-answers it from a real build; clearing it would only matter if the app list could empty *after* a failure, and then the honest state is still "loading".

**2. `indexBuilds` is wired in but never asserted (nitpick — taken).**

Fair, and it was pointing at real gaps. Three tests now, in `AppListViewModelTest`:

- `a persisted permission filter reads as loading until the app list arrives` — the regression for (1). Verified it fails against the previous condition, not just passes against the new one.
- `the index is built once per app-set change and never for a sort or a search` — one build on the first real app load, still one after `updateSearchQuery` + `updateSort` + `updateSortOrder`, two after the same package reappears with a new `lastUpdateTime`. That last one is the point of the `packageName@lastUpdateTime` key: an update invalidates, a freeze or a size arriving does not.
- `a failed index is announced and then admitted to in the chip row` — the toast is sent, `permissionIndexFailed` is set, and the stale index is dropped rather than silently filtered on.

`userApp` gained a `lastUpdateTime` parameter for the second one; that field is half the invalidation key and there was no way to say "the same app, updated".

**3. MD018 on the roadmap line starting with `#285` (nitpick — taken).**

Reworded to "see the **#285** row below".

`testFossDebugUnitTest --rerun-tasks`, `assembleFossDebug`, `lintFossDebug` and `lintStoreRelease` green locally.

PR #295, opened 2026-07-30
feat(freezer): named freeze profiles (#55a)

### LINKED ISSUE #55: Process manager support and profiles for freezing apps

trinadhthatakula:
Closes #55 (the #55a half — the process manager, #55b, stays declined).

A **freeze profile** is a named set of apps you can freeze or unfreeze in one tap.

- Two new Room tables, `freeze_profiles` + `freeze_profile_apps`, via **auto-migration 5 → 6**. New tables only, nothing existing touched, so a shipped database gets them and keeps its watchlist. Schema export committed.
- A profiles sheet (per-row freeze / unfreeze / edit / delete, with a delete confirmation), an editor sheet that reuses the watchlist's app picker, and a **"save this selection as a profile"** entry in the multi-select toolbox.
- New toolbar button on the Freezer screen. Not privilege-gated — naming a set of apps is not freezing them.

## Profiles are not a subset of the watchlist

Watchlist membership is a standing instruction that the QS tile and both launcher Freeze-all shortcuts act on **with no UI at all**. Folding profiles into it would mean creating a "Games" profile silently enlists every game in the tile's Freeze-all. The two lists stay independent; what they share is `BulkFreezeRunner`.

## The three risks the roadmap named are closed

1. **Tier gate.** Profile runs route through `BulkFreezeRunner`, so `targetsFor` → `freezableCandidates(...)` applies the list-level `FreezeTier.BLOCKED` filter for free. Routing *through the runner* is what earns the gate — a fourth surface calling `setAppDisabled` itself would have been the one bulk path able to freeze what every dialog in the app refuses to render a confirm button for. The editor additionally raises `AppRiskDialog` at *add* time, because a profile is a standing instruction that later runs act on with no UI.

2. **Coalescing key.** The runner's job slot keyed on `BulkOp` alone would have made "freeze profile A" then "freeze profile B" return the first `Deferred` and silently never freeze B. The key is now `BulkRequest(op, scope)`:
   - same request → coalesce (unchanged);
   - **same op, different scope → serialize** (`join()`), because two freezes do not contradict each other and cancelling A to start B would leave A half-frozen with nothing said about it;
   - conflicting op → replace (`cancelAndJoin()`), unchanged in spirit but now cancelling the **whole unsettled chain** rather than just its tail. Found in self-review: the tail of a serialized chain is the run that has not started yet, the head is the one holding packages open, so cancelling the tail alone left a freeze running alongside the replacing unfreeze.

3. **Launcher restore.** `FreezerBridgeProvider` now gates on the watchlist **∪** every profile's membership. An app frozen only by a profile it is the sole member of used to be un-unfreezable from the launcher — a dead icon tap with nothing said about why.

## Tests

`FreezeProfileTest` — 17 cases over the two pure pieces: the name rule (trimmed, capped at 40, duplicate detection **case-insensitive to match the `COLLATE NOCASE` unique index**, so the inline error and the database cannot disagree) and `BulkRequest` identity, including a regression test that two profiles under the same op are *not* the same request.

Suite is **226 tests, 0 failures** (was 209 on `dev`). `assembleFossDebug`, `testFossDebugUnitTest --rerun-tasks` and `lintFossDebug` all green — lint runs with `warningsAsErrors`, so all five locales are translated.

## Notes for review

- **Not fixed here, pre-existing:** `BulkFreezeRunner`'s outer catch returns `null` for a genuinely failed run, which is the same value it uses for "no-op". The profiles sheet therefore reports a rare hard failure as *"Nothing to do"*. Changing it would alter what the tile and both shortcuts report, so it did not belong in this branch.
- While a profile run is queued behind another run, `runningRequest` already names the queued one, so the spinner moves to the pending row a moment early. Self-healing (the tile re-sweeps on `onStartListening`) and left alone deliberately.
- **Needs a device:** both new sheets are unverified visually, and a profile run has not been exercised against a real privilege backend.

## Summary by CodeRabbit

* **New Features**
  * Added freeze profiles for saving, editing, deleting, and reusing groups of apps.
  * Added profile creation from selected apps, app search and filtering, validation, and risk confirmations.
  * Profile-based freeze and restore operations now include all associated apps.
  * Added clearer progress, completion, no-op, and failure feedback for bulk operations.
  * Added database migration support and localized profile-related messaging.

* **Bug Fixes**
  * Improved operation tracking when multiple freeze requests run or queue concurrently.
  * Updated restore authorization to recognize apps included in freeze profiles.


trinadhthatakula (commented):
## Code review summary

The profile schema, repository transactions, tier filtering, launcher restore gate, and request identity are internally consistent. I found no blocking correctness issue.

### Non-blocking suggestions

1. **Test runner behavior, not only request equality.** The central risk in this PR is the new same-request coalescing, same-op serialization, and conflicting-op chain cancellation in `BulkFreezeRunner`. Current profile tests prove `BulkRequest` identity but do not execute those coroutine handoffs. Deterministic runner tests would protect the behavior that closes the roadmap risk.

2. **Keep editor state until persistence succeeds.** `FreezerScreen` closes the editor immediately after dispatching the asynchronous create/update. A constraint or I/O failure shows a toast after the draft has been discarded. Consider returning save state/result to the UI and dismissing only on success.

### Verification

- Reviewed commit `c93472be89e06207a1ca0fc818abcde5b7d4a438` in an isolated clone.
- `FreezeProfileTest` passed.
- `assembleFossDebug` succeeded.
- GitHub checks are green.
- Room schema 6 contains the expected profile, membership, foreign-key, and index definitions.

### Verdict

**Approve with non-blocking suggestions.**

trinadhthatakula:
Latest CodeRabbit finding applied in `a9ef0976`.

**"Keep bulk-run failures distinct from intentional no-ops"** (Major) — confirmed, and it is worse than the report suggests, because it is the *only* thing the surface can say.

`BulkFreezeRunner` returned `null` from two places that mean opposite things: `run()` returning null for "no privilege, or nothing left after the tier filter", and the `catch (e: Exception)` that keeps a Room or binder failure from killing the process. Every caller that awaited a run therefore reported a failure as **"Nothing to do for this profile"** — the one reading that is certainly wrong, on the run most likely to have left the watchlist half-frozen.

The awaited value is now `BulkOutcome`:

```kotlin
sealed interface BulkOutcome {
    data class Completed(val result: BulkResult) : BulkOutcome
    data object NothingToDo : BulkOutcome
    data class Failed(val cause: Throwable) : BulkOutcome
}
```

`NothingToDo` stays a first-class outcome rather than a degenerate `Completed`, for the reason B-8 already gave: `BulkResult(0, 0, 0)` renders as "Froze 0 apps", a false report of a freeze that never ran.

Both awaiting surfaces were fixed, not just the profile one — `FreezerLaunchActivity` had the same conflation through `run.isCompleted && !run.isCancelled`, so an Unfreeze-all launched from a pinned shortcut also said "no apps" when the batch had blown up. The failure wording is deliberately vague about what got frozen (`bulk_run_failed`: "That run did not finish — check the apps before relying on it") because the runner genuinely does not know: the throw can land before the first package or halfway through the batch.

**One thing the finding did not name.** `notifier.post(result)` sat inside the same try, so a run that froze every app it was given was reported as a *failure* when only the notification about it failed to post. It now catches around `post` alone: the packages are already mutated by that line, and failing to describe a freeze is not failing to perform it.

Strings added in all five locales. `assembleFossDebug testFossDebugUnitTest --rerun-tasks` → **BUILD SUCCESSFUL, 228 tests, 0 failures**.

No test covers the new `Failed` path: `BulkFreezeRunner` takes `Context`, the notifier and a `SystemGateway`, and `:app` has no mocking library. The mapping is three lines with no branching beyond the `when`, but that is a reason it is low-risk, not a claim that it is verified.

---

⚠️ **`build-and-test` red on this branch is not this branch.** AGP `9.4.0-alpha07` published today and lint's `AndroidGradlePluginVersion` check (severity `ERROR`, so `warningsAsErrors` was never what promoted it) now fails `lintStoreRelease` on `gradle/libs.versions.toml:6`. Every open PR failed within the hour on a line none of them touched. #298 downgrades it in `app/lint.xml` alongside the two sibling checks that were already downgraded for exactly this reason; once it lands, this branch needs `dev` merged in.

trinadhthatakula (commented):
## Review summary

This PR adds named freeze profiles, Room persistence and normalization, profile management UI, bulk freeze/unfreeze execution, result reporting, and supporting tests. The persistence and profile-flow changes are internally consistent, and the full Foss debug unit suite plus `git diff --check` pass at `4b2291091180f8f70914f6ad47e5b222bb7e11ca`.

## Blocking finding

**[P1] Coalesce against the matching operation suffix, not only a completely uniform unsettled chain**  
`app/src/main/java/com/valhalla/thor/data/freezer/BulkFreezeRunner.kt:303-305`

The `unsettled.all { it.request.op == request.op }` guard misses a valid duplicate of the newest active request while a cancelled opposite-operation predecessor is still unwinding. For example:

1. A `FREEZE` request is active.
2. B `UNFREEZE` arrives, cancels A, and becomes the active replacement.
3. Before A leaves `unsettled`, the same `UNFREEZE` request arrives again.
4. The chain is temporarily `[A(FREEZE), B(UNFREEZE)]`, so the `all` check fails and C is enqueued behind B.
5. B and C both execute the same unfreeze batch and report twice.

Please search only the contiguous same-operation suffix of `unsettled` for an active identical request (for example, iterate from the end while `request.op` matches), so stale cancelling predecessors do not prevent coalescing while an earlier same-op request across a conflicting operation is still excluded. A focused regression test for the A/B/repeated-B timing would protect the intended invariant.

## Verdict

**Changes requested.** The feature is otherwise in good shape, but the duplicate bulk action can repeat package operations and user-visible results.

_Note: GitHub does not allow the PR author to submit a formal Request changes review, so this verdict is posted as a review comment._

trinadhthatakula:
P1 fixed in `a8cf5ab7`. The analysis is right and so is the suggested shape — the trailing same-op run isn't just a cheaper approximation of "not doomed", it's equal to it:

> an entry E is doomed by some later entry X whose op differs from the entry before X; if everything from E to the end shares E's op, no such X can exist after E.

So the suffix answers the repeated-B case with B while still refusing to hand back the cancelled A — which is the case the uniformity check was added for, and the one a plain unconditional search over the chain would get wrong in the other direction. Both failure modes are silent, which is why it now has its own tests.

On the regression test: `BulkFreezeRunner` can't be constructed on a JVM (four final collaborators over `Context`/`PackageManager` — the `BulkFreezeWorkerTest` KDoc lists them), so the rule is lifted out as `coalesceTargetIndex(size, request, requestAt, isActive)` and asserted directly. Nine cases in `CoalesceTargetIndexTest`, including the exact A/B/repeated-B chain you described, its mirror (a repeat of the doomed A must **not** coalesce), the multi-boundary chain `[F1, U, F, F2]`, and the scope-identity rule.

245 tests / 0 failures; `assembleFossDebug`, `lintFossDebug`, `lintStoreRelease` clean.

PR #308, opened 2026-07-31
fix(freezer): removing an app from the watchlist always restores it

trinadhthatakula:
## The bug

Taking an app off the Freezer watchlist is supposed to thaw it first, so it can never end up frozen but no longer tracked. Three surfaces can remove an app from the watchlist, and only one of them actually did.

| Surface | Removes from watchlist | Restores the app |
|---|---|---|
| `FreezerViewModel.removeFromFreezer` / `toggleManaged` | yes | **yes** |
| `AppListViewModel.toggleFreezerMembership` | yes | **no** |
| `AppInfoDetailsViewModel.addOrRemoveFromFreezer` | yes | **no** |

On the two broken paths the watchlist row was dropped, the launcher shortcut was greyed out, and a success message was shown — while the app stayed disabled or suspended. That strands it: the Freezer screen no longer lists the app, so the user's only route back is the import-already-disabled flow. Which answer you got depended on which screen you happened to tap.

## The fix

Both paths now restore before reporting success:

- **`restoreApp(pkg, enabled, isSuspended)`** when the `AppInfo` is resolvable, so an app that is *both* disabled and suspended has both halves undone (the mixed state from #239).
- **`forceUnfreeze(pkg)`** when it is not — before the app-info sheet's first detail load, or when a search/filter has hidden the row in the app list. Both of its halves are no-ops on an already-active app, so the fallback is safe rather than merely tolerable.
- A **failed thaw surfaces the error** instead of the removal message. Previously a dead shell still reported "removed".

`AppListViewModel` also patches its local state optimistically, exactly as `freezeApp` already does, so the row stops reading as frozen without waiting for a full rescan.

Resolution is against `_rawState`, not the filtered list, for the same reason the add path is: a search or filter that hides the app must not decide its fate.

**The add path is deliberately unchanged.** The asymmetry is the point — adding never freezes, removing always restores — and the KDoc now says so at both sites.

## Testability changes

The app-info sheet could not be constructed in a JVM test at all, so the second half of the fix would have shipped unverified. Two narrow changes:

- `AppShortcutController` gains `refreshAppShortcut`, and `AppInfoDetailsViewModel` takes that port instead of the concrete `FreezerShortcutManager`, which needs a `Context`. `AppListViewModel` already depended on the interface, so this makes the two consistent rather than introducing a new pattern.
- That view model's five hardcoded `Dispatchers.IO` uses become an injected `@Named("io")` dispatcher, per the rule in `CLAUDE.md`. Without it the work escapes the test scheduler and nothing is deterministically assertable.

`FakeAppRepository` gained a settable `details` map (defaulting to empty, so `getDetailedAppInfo` still returns null — itself a state under test).

## Tests

Eight new regression tests. **Each was verified to fail against the pre-fix code**, not merely to pass against the new code:

`AppListViewModelTest` (3)
- `removing a suspended app from the watchlist unsuspends and enables it` — was `expected:<[setAppSuspended:a:false, setAppDisabled:a:false]> but was:<[]>`
- `the row stops reading as frozen without waiting for a rescan` — was `AssertionError: enabled again`
- `a failed restore is reported instead of the removal message` — previously got the removal `PluralsResource`

`AppInfoDetailsViewModelTest` (5, new file)
- `removing a disabled and suspended app undoes both halves`
- `removing before details load still restores, unconditionally`
- `a failed restore is reported instead of the removal message`
- `adding to the watchlist freezes nothing` — pins the other half of the asymmetry
- `removal retires the launcher shortcut`

Full suite: **384 tests, 0 failures, 0 skipped** (`:app:testFossDebugUnitTest --rerun-tasks`).

## Note

This also unblocks a FAQ answer for the website, which currently carries a "don't publish this until the code matches" gate on exactly this behaviour.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removing an app from the freezer watchlist now restores its previous enabled or suspended state.
  - Failed restorations display an error instead of reporting success.
  - App status updates immediately after removal, without waiting for a rescan.
  - Launcher shortcuts are disabled and refreshed when apps are removed or their freezer state changes.

- **Tests**
  - Added coverage for restoration, error handling, shortcut updates, and immediate status changes.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/launcher/FreezerShortcutManager.kt
+++ b/app/src/main/java/com/valhalla/thor/data/launcher/FreezerShortcutManager.kt
@@ -102,7 +102,7 @@ class FreezerShortcutManager(
 
     /** Update an already-pinned per-app shortcut so its icon reflects the app's current state.
      *  No-op if no such shortcut exists. Call after any freeze/unfreeze of the package. */
-    fun refreshAppShortcut(packageName: String) {
+    override fun refreshAppShortcut(packageName: String) {
         scope.launch { updateShortcutIcon(packageName) }
     }
 
--- a/app/src/main/java/com/valhalla/thor/domain/repository/AppShortcutController.kt
+++ b/app/src/main/java/com/valhalla/thor/domain/repository/AppShortcutController.kt
@@ -12,4 +12,10 @@ package com.valhalla.thor.domain.repository
 interface AppShortcutController {
     /** Disable (and hide) any shortcut targeting [packageName] — it is gone or no longer launchable. */
     fun disableAppShortcut(packageName: String)
+
+    /**
+     * Re-render an already-pinned shortcut for [packageName] so its icon matches the app's current
+     * frozen/active state. No-op when nothing is pinned. Call after any freeze or unfreeze.
+     */
+    fun refreshAppShortcut(packageName: String)
 }
--- a/app/src/main/java/com/valhalla/thor/presentation/appList/AppInfoDetailsViewModel.kt
+++ b/app/src/main/java/com/valhalla/thor/presentation/appList/AppInfoDetailsViewModel.kt
@@ -6,19 +6,19 @@ package com.valhalla.thor.presentation.appList
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.valhalla.thor.R
-import com.valhalla.thor.data.launcher.FreezerShortcutManager
 import com.valhalla.thor.domain.model.DetailedAppInfo
 import com.valhalla.thor.domain.model.FreezeTier
 import com.valhalla.thor.domain.model.freezeTier
 import com.valhalla.thor.presentation.freezer.FreezerPrompt
 import com.valhalla.thor.domain.repository.AppRepository
+import com.valhalla.thor.domain.repository.AppShortcutController
 import com.valhalla.thor.domain.repository.FreezerRepository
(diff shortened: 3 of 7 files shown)


trinadhthatakula:
Addressed CodeRabbit's findings in `1a5e5f13`.

**Findings 1 and 3a — valid, fixed.** Both removal paths did the durable thing first and the fallible thing second, so a failed restore dropped the watchlist row and left the app frozen with nothing pointing at it. Restore now runs first in both `AppInfoDetailsViewModel.addOrRemoveFromFreezer` and `AppListViewModel.toggleFreezerMembership`; the Room delete, the shortcut greying and the success toast all happen only after it succeeds. Removal is all-or-nothing.

**Finding 2 — valid, applied on top.** Worth noting it could not have been applied on its own: both assertions assert `true`, and under the old ordering membership was already gone by assert time, so committing that suggestion alone would have turned the build red. With the reorder in place the assertions bite correctly — I verified by restoring the old ordering and re-running: exactly those two tests fail and no others. 384 tests, 0 failures with the fix in.

**Finding 3b — valid, out of scope, filed as #310.** `FreezerViewModel.removeFromFreezer` has the same shape and is worse: it discards every `Result` in the loop and emits the success plural with `packageNames.size` regardless, so twelve failed restores still report "Removed 12 apps". That needs a per-package failure tally and a new string (with all four translations), which does not belong in this diff.

**Finding 3c — not applicable.** `AppListViewModelTest`'s failed-restore test asserted only on the emitted event list; it made no claim about membership either way. It does now.

PR #309, opened 2026-07-31
chore(build): let a keystore-less clone build a release, and add the F-Droid guide

trinadhthatakula:
Three things, all pointing the same way: making Thor buildable by someone who is not us.

## 1. A fresh clone cannot build a release APK

The `release` signing config has a three-way branch — local `jks.properties`, CI env vars, then a fallback that only logs a warning — but the release build type assigned that config **unconditionally**. With no credentials the config is empty and still assigned, so `validateSigning<Flavour>Release` fails with

```
Keystore file not set for signing config release
```

and the build dies before producing anything. Anyone rebuilding from source — F-Droid's buildserver included — hits this on a clean clone.

`signingConfig` is now assigned only when credentials actually exist:

```kotlin
val hasSigningCredentials: Boolean =
    keystorePropertiesFile.exists() || System.getenv("KEY_ALIAS") != null
...
signingConfig = signingConfigs.getByName("release").takeIf { hasSigningCredentials }
```

Verified both directions on this branch:

| Condition | Command | Result |
|---|---|---|
| credentials present | `validateSigningFossRelease` | BUILD SUCCESSFUL — unchanged |
| credentials absent | `assembleFossRelease` | **BUILD SUCCESSFUL**, emits `app-foss-release-unsigned.apk` (previously failed outright) |

CI and local signing are untouched, because both satisfy the condition. The distribution copy task renames by wildcard (`rename(".*\\.apk", ...)`), so the `-unsigned` suffix is absorbed there too — no name coupling to break.

## 2. The fastlane changelog was five releases stale

`fastlane/metadata/android/en-US/changelogs/` held only `1600.txt` while `versionCode` is `1931` — so F-Droid and IzzyOnDroid have shown no "What's New" text since v1.60.0. Added `1931.txt`, copied from the existing `release-notes/v1.93.1/playstore.txt`.

Worth wiring into the release script as a follow-up: the notes are already written each release, nothing was copying them here.

## 3. `docs/fdroid-submission.md`

What it would actually take to get Thor into the **main F-Droid repo** (IzzyOnDroid is a separate third-party repo and confers no standing there), checked against the current Inclusion Policy and Build Metadata Reference rather than from memory.

Short version:

- ✅ The `foss` flavour is already clean — Play Billing is `storeImplementation`-only, no proprietary deps reach it.
- ⛔ **Blocker:** the pre-release AGP (`9.4.0-alpha07`) is a policy problem the build recipe cannot work around. That's what sets the date.
- ⛔ **Blocker:** the signing config — fixed by this PR.

The doc also contains a full drafted `metadata/com.valhalla.thor.yml`, the buildserver docker test loop, Route A (RFP issue) vs Route B (`fdroiddata` MR) with a recommendation, an explicit warning that `AllowedAPKSigningKeys` + `Binaries` reproducible builds is a **one-way door**, and a "what I could not verify" section.

## Testing

- `assembleFossRelease` with and without credentials, as tabulated above. `jks.properties` restored after each run.
- No source changes — only the build script, a metadata file, and a doc.

## Summary by CodeRabbit

* **New Features**
  * Added Freeze Profiles, batch app export, `.xapk` export, permission-based filtering, and consolidated app information.
  * Improved freezer tile and shortcut reliability.
  * Enhanced App Lock safety.
* **Bug Fixes**
  * Release builds now complete without signing credentials, while preserving signing when credentials are available.
* **Documentation**
  * Added guidance for submitting Thor to F-Droid, including metadata, validation steps, and known requirements.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -39,6 +39,12 @@ if (keystorePropertiesFile.exists()) {
     keystoreProperties.load(FileInputStream(keystorePropertiesFile))
 }
 
+// Whether this machine can sign a release at all: a local jks.properties, or the CI env vars.
+// Single-sourced here because both the signingConfig block and the release build type need the
+// same answer — see the comment on `signingConfig` below for why the build type has to ask.
+val hasSigningCredentials: Boolean =
+    keystorePropertiesFile.exists() || System.getenv("KEY_ALIAS") != null
+
 // --- VERSIONING HELPERS (Private & Modernized) ---
 
 // 1. Resolve Code: Checks property 'versionCode' first, falls back to 'initialVersionCode'
@@ -129,7 +135,14 @@ android {
                 getDefaultProguardFile("proguard-android-optimize.txt"),
                 "proguard-rules.pro"
             )
-            signingConfig = signingConfigs.getByName("release")
+            // Only assign the config when credentials actually exist. The `release` signingConfig
+            // above leaves every field unset in the no-credentials case, and an assigned-but-empty
+            // config fails validateSigning<Flavour>Release with "Keystore file not set for signing
+            // config release" — so a fresh clone (and anyone rebuilding from source, F-Droid's
+            // buildserver included) cannot produce a release APK at all. Unsigned output is the
+            // correct result there; signing happens downstream. CI and local signing are unchanged,
+            // because both satisfy the condition.
+            signingConfig = signingConfigs.getByName("release").takeIf { hasSigningCredentials }
         }
         debug {
             isMinifyEnabled = false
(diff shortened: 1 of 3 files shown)


trinadhthatakula:
Addressed in `6035f373`.

**Signing-blocker status — valid, fixed.** The doc was written before the build change and still read as a proposal. The verdict table listed the keystore abort as an open ⛔ blocker, §4 said "the fix, in this repo", and §10's suggested order opened with two steps this very commit had already done. All three now say landed, and §4 carries the verification result for both directions.

**Stale line reference — valid, fixed by removing the number.** `versionNameSuffix = "-foss"` moved from 170 to 183 because the signing change above it added lines, which is a good argument for a doc not citing line numbers at all. Now references the `foss` flavour block by name.

**Missing fence language — valid, fixed.** `bash` added.

**Document date — not applicable.** The doc says 1 August 2026 and that is today's date; the sources were fetched the same day.

While in there I also rewrote the stale-changelog note. `1931.txt` now exists, but the useful finding isn't the missing file — it's that nothing *copies* `release-notes/vX/playstore.txt` into `fastlane/.../changelogs/`, which is how it drifted five releases behind unnoticed. The doc previously said "put it in the release checklist"; a checklist item is exactly what already failed, so it now says to put it in the release script.

PR #255, opened 2026-07-18
chore(release): merge dev into master for v1.92.1 Release

trinadhthatakula:
This PR merges all the stable modernizations, the new Auto Reinstall Apps feature, the Odin root engine integration, and modern platform bypasses from `dev` into `master` for the stable `v1.92.1` release.

## Summary by CodeRabbit

- **New Features**
  - Added Auto Reinstall Apps, preserving Google Play Store as the installer for background updates.
  - Introduced a faster, more reliable root communication engine with improved security and lifecycle handling.
  - Added support for suspending apps and clearing app data through the updated root service.

- **Bug Fixes**
  - Improved compatibility across Android versions and custom ROMs.
  - Reduced crash risks, connection leaks, and main-thread blocking.

- **Style**
  - Refined settings interactions, button styling, tab navigation, and scrolling text behavior.

PR #314, opened 2026-08-01
fix(freezer): freeze system apps by disabling them, not by uninstalling

trinadhthatakula:
## The bug

Freezing a **preinstalled** app destroyed its data.

Every privilege mode ran `pm uninstall --user N` — with no `-k` — as the *first and only* mechanic for system apps. Unfreezing ran `pm install-existing`, which brings the package back **factory-fresh**: logins, settings and local content gone. The UI reported an ordinary, reversible freeze. Nothing warned about it, and nothing in the codebase distinguished the two outcomes.

Both gateways now **try to disable first**, and only escalate where the platform genuinely leaves no alternative.

## The chains

**Root** — `RootSystemGateway`

1. `pm disable --user N` — data preserved.
2. `pm uninstall -k --user N` — **gated off entirely**. A failure to disable now surfaces as a failure instead of silently changing mechanic.

`--user N` rather than the bare form is deliberate: `PackageManagerShellCommand.runSetEnabledSetting` seeds `userId = USER_SYSTEM`, so a bare `pm disable` issued from a work profile or Second Space targets user 0 and then fails its own verification. This file already documents that exact trap for `pm grant`/`pm revoke`.

**Shizuku** — `ShizukuSystemGateway`

1. `IPackageManager.setApplicationEnabledSetting` via `:bypass`.
2. `pm disable-user --user N`.
3. unprivileged `PackageManager` — free, and almost always a no-op.
4. `pm uninstall -k --user N` — **gated**.

## The fallback no longer destroys anything

`-k` sets `DELETE_KEEP_DATA`, which leaves `/data/user/N/<pkg>` and `/data/user_de/N/<pkg>` in place instead of having `installd` destroy them. Measured on the HyperOS device that refuses to disable system packages at all — the device this fallback exists for:

```
pm uninstall -k --user 0 <pkg>      -> Success, exit 0
pm install-existing --user 0 <pkg>
ceDataInode / deDataInode           -> byte-identical before and after
```

For a system app the data survives indefinitely: the package record never goes away, because the APK is still on the read-only partition and `pm uninstall --user` only clears `PackageUserState.installed` for one user. (The adb client's "there is no way to remove the remaining data" warning about `-k` is about orphaned data after a *full* uninstall of a user app. It does not apply here, and it did not appear on the device.)

**The grants survive too — this was measured after an earlier revision of this PR guessed otherwise.** Five sites in the branch said runtime permission grants were "expected not to survive the round trip". They do. On a stock AOSP API 36 emulator at uid 2000:

```
pm grant com.android.egg READ_EXTERNAL_STORAGE  -> granted=true
pm uninstall -k --user 0 com.android.egg        -> Success
pm install-existing --user 0 com.android.egg    -> Success
READ_EXTERNAL_STORAGE                           -> granted=true, flags unchanged
ceDataInode / deDataInode                       -> byte-identical
```

`-k` retains the whole `PackageUserState`, not just the data directories. The pessimistic guess had been stated in two user-visible log strings as if it were fact; all five sites are corrected. What the rung *does* still change is `FLAG_INSTALLED`, which is the part the rest of the app has to handle.

**The rung does not exist at shell uid on API 37.** Identical command, same uid, AOSP emulators:

```
pm uninstall -k --user 0 com.android.egg
  API 36 -> Success, exit 0
  API 37 -> Failure [only root can delete system app for a particular user], exit 1
```

The package is left untouched, so on Android 17 a Shizuku at uid 2000 ends the chain at "could not disable" with an honest failure rather than a wrong state. A Shizuku started as root is unaffected. Documented at the call site, with an explicit note not to add a rung below it.

`freezeSystemAppForUser` is deliberately a **separate function** from `uninstallApp`, in both the Shizuku helper and the reflector — adding `-k` to the shared one would silently make the user-facing "uninstall this app" feature leave data behind on every app it removes. The new one is shell-only with no `PackageInstaller` fallback, because that fallback cannot express `DELETE_KEEP_DATA` and would quietly turn a data-preserving freeze back into a destroying one.

## The gate — and why it is not a version check

`uninstallFreezeFallbackAllowed` in `domain/model/FreezePolicy.kt` is the single home for the escalation decision. It answers `true` for exactly one combination: **Shizuku, on a system app, where a privileged rung was refused by the platform.** Everything else fails loudly rather than switching mechanic behind the user's back.

It briefly answered `sdkInt >= 36` instead, on the report that shell-uid disabling of system apps stopped working on Android 16. **That boundary does not exist.** Measured on a stock AOSP Android 16 (API 36) emulator as uid 2000:

| command | result |
|---|---|
| `pm disable --user 0 com.android.egg` | `SecurityException: Shell cannot change component state for null to 2`, exit 255 |
| `pm disable-user --user 0 com.android.egg` | **exit 0**, `enabled=3`, `installed=true` |

Same on `com.android.printspooler`, `com.android.wallpaper.livepicker` and `com.android.traceur`; all four reversed with `pm enable` and verified back at `enabled=1`. Disabling a system app from the shell uid on Android 16 works, and the data survives.

The command that fails is `pm disable` — `COMPONENT_ENABLED_STATE_DISABLED` (state 2), which the shell uid has been forbidden from setting **since API 25**; it may only set `DEFAULT`, `ENABLED` or `DISABLED_USER`. That is a *command* constraint, not a version one, and the Shizuku path already satisfies it by sending `pm disable-user`. AOSP's guard block in `PackageManagerService.setEnabledSettings` is **byte-identical** across android14-, android15-, android16-, android16-qpr1- and android16-qpr2-release; the 15→16 diff of that method contains tracing and metrics and no security logic.

The restriction users actually hit is **Xiaomi's**: a vendor `PackageManagerServiceImpl.shouldRestrictEnabledSettingsChange` — a class with no AOSP counterpart, which 404s at every `android.googlesource.com` tag from 13 to 17 and lives in `/system_ext/framework/miui-services.jar` on the device — throws `SecurityException("Cannot disable system packages.")` for `callingUid == 2000` on a system package. It was first reported on HyperOS running **Android 14**, roughly a year before Android 16 shipped. Reproduced here on `25053PC47G` (HyperOS OS3.0, `BP2A.250605.031.A3`), where it covers `/system/app`, `/system/priv-app`, `/product/app` and `UPDATED_SYSTEM_APP` alike while third-party packages disable normally.

So a version test is wrong in both directions at once: it would strip the wrong mechanic onto a Pixel that could have disabled the app, and refuse to freeze at all on the devices that are the entire reason the fallback exists. What separates the two cases is not the release but whether the platform **refused** — so that is what the gate asks. `disableRefusedByPolicy` is set only by a `SecurityException` from `PackageManagerService`, never by a non-zero exit, a timeout or an unreadable read.

Confining escalation is the whole point. Without that distinction, any transient failure to disable — a busy `PackageManager`, a binder timeout, a package caught mid-update — silently swaps the mechanic and produces no bug report, because from the outside it looked like it worked.

**Root answers `false` whatever the flag says.** The two refusals this fallback was built for both key on the **shell** uid (2000), and root is uid 0. One refusal genuinely can reach root — `ProtectedPackages` (device provisioning package, DO/PO, DPM owner-protected) throws `Cannot disable a protected package: <pkg>` and is *not* uid-gated; reported on Amazon Fire and on Infinix XOS. That case still answers `false`, and deliberately: a package the platform protects that hard is one where "remove it for the user instead" is the wrong reading of the refusal. Root's flag is still computed and passed honestly, so both gateways call the gate identically and letting root escalate would be a change in the policy, not at the call site.

**Unfreeze deliberately does not consult the gate.** Devices in the field carry apps frozen by the old unconditional-uninstall builds, and a single package can be *both* disabled *and* uninstalled-for-user, so both gateways walk `install-existing` → `enable` and verify the end state. Gating the thaw would strand exactly the users who most need it.

## No rung trusts an exit code

`pm` prints `Success` for a no-op and returns non-zero from commands that did take effect, so every rung is judged by **re-reading `ApplicationInfo`**. The ordering and short-circuit rules are extracted as pure functions — `firstRungThatSticks`, `RootFreezeChain`, `orderRungs` — to make them reachable from plain JVM tests.

Rungs report a `RungResult` (`RAN` / `FAILED` / `REFUSED_BY_POLICY`) rather than a bare `Boolean`, and `firstRungThatSticks` returns a `ChainOutcome` whose refusal flag is **sticky**: a refusal on the reflection rung is a fact about the device, so a later rung merely failing must not erase it. The unprivileged rung can never report a refusal — Thor holds no `CHANGE_COMPONENT_ENABLED_STATE`, so it throws `SecurityException` on every device, and counting that would green-light the destructive fallback permanently.

The `setApplicationEnabledSetting` reflection is now the one and only copy in the codebase. Its two conditionals are load-bearing and documented as such: `COMPONENT_ENABLED_STATE_DISABLED` vs `DISABLED_USER`, and `com.android.shell` vs Thor's own package name — `PackageManagerService` validates `callingPackage` against the calling uid, which differs between a shell-uid Shizuku (2000) and a root Shizuku (0, owns no package).

## Two UI leaks, both widened by the new mechanic

- **The Freezer import prompt filtered on `!isSystem`**, which skipped every system app Thor itself had frozen. That clause predates the tier model by five weeks and was never a no-op: `AppInfo.enabled` folds in `FLAG_INSTALLED`, so uninstalled-for-user system apps have *always* matched `!enabled` and have always been excluded. Replaced by `importableDisabledApps`, gated on `isBlockedFromFreeze` instead — importing a BLOCKED app is a one-way door, since Unfreeze-all would enable it and Thor would then refuse to re-freeze it.
- **`AppList` drew a second, differently-styled badge** for `!isInstalled` — a branch that is unreachable once `!enabled` is tested first. Deleted, and the badge extracted to `AppStatusBadge` so the list and grid cannot drift again. `cd_uninstalled` had no other use and is removed from all five locales (465 → 464 entries each; named-resource sets verified identical, so `MissingTranslation` stays clean).

## Prose

~22 comments, KDoc blocks and test-assertion messages asserted the old mechanic as fact. All corrected.

One is a real hazard rather than stale prose: `AutoFreezeManager`'s `getApplicationInfo(pkg, 0)` looks like an oversight next to every other lookup in the codebase, but **flags `0` is load-bearing**. Widening it to `AppFreezeStateReader.MATCH_FLAGS` would make an uninstalled-for-user package resolve with `enabled == true`, and auto-freeze would start issuing freeze commands against apps it can already see are frozen. Now annotated as do-not-widen at the call site.

## Known gaps — deliberate, and worth a reviewer's attention

- **`DhizukuSystemGateway` still uninstalls unconditionally.** Converting it without a device check risks removing Dhizuku's only working system-app freeze, so it is left unconverted rather than half-converted. After this ships, the same button preserves data under Root/Shizuku and wipes it under Dhizuku, with no UI difference.
- **The refusal predicate is text-matching on the shell rung.** `Shizuku.execute` hands back stderr as a string, so `isPolicyRefusal(String?)` looks for `SecurityException` in it; the reflection rung gets the real throwable and walks its cause chain (bounded at 10 hops, so a cyclic chain terminates). An OEM that refuses with a differently-worded exception would read as an ordinary failure — which fails *closed*, costing that device the fallback rather than costing its owner a mechanic change they did not ask for.
- **`pm suspend` was measured working on the refusing device** and would be a better rung than uninstall-for-user — it keeps the data *and* the launcher icon with no uninstall/reinstall round trip. It is not added here because Thor already exposes suspend as a **user-chosen freeze mode** (the v1.92.0 Freeze Mode Selector), and silently substituting it inside the *disable* mode is a product decision, not a bug fix. Worth its own issue. (`pm hide` is blocked on that device, so it must not go ahead of suspend in any future ladder.)
- **API 37's new required-packages guard is live, not inert.** An earlier revision of this section had it backwards: it reasoned that because shell's manifest holds zero occurrences of `ALLOW_CONTROL_SYSTEM_REQUIRED_PACKAGES`, the restriction was inert. Holding that permission is the **bypass** — holding none means shell is *subject* to it. Measured on Android 17 (`CP31.260623.005`):

  ```
  pm disable-user --user 0 com.google.android.packageinstaller
  -> SecurityException: Cannot disable required package
     at PackageManagerService.setEnabledSettings(PackageManagerService.java:4218)
  ```

  It fires for a small allowlist — `packageinstaller` and `permissioncontroller` — and not for `settings`, `systemui`, `gms`, `nexuslauncher` or `intentresolver`, all of which still disable normally at uid 2000. It is a `SecurityException`, so `isPolicyRefusal` already classifies it correctly and the existing gate handles it with **no code change**: the chain records the refusal, finds the uninstall rung unavailable on 17, and fails closed. That is the right outcome for a package the platform says the device requires.
- **Pre-existing, out of scope:** the non-system branch of `RootSystemGateway.setAppDisabled` still uses bare `pm disable` with no `--user`, so on a work profile it targets user 0 — the same bug fixed six lines above it.

## Verification

- `assembleFossDebug` — green
- `testFossDebugUnitTest --rerun-tasks` — **429 tests, 0 failures**, 45 of them new
- `lintFossDebug` — clean (`warningsAsErrors`, `abortOnError` both on)
- Zero new Kotlin compiler warnings
- **Device-tested on three devices.** A stock AOSP **Android 16 (API 36)** emulator and an **Android 17 (API 37)** Pixel 10 Pro Fold emulator, both via the shell uid 2000 that Shizuku runs as, plus the HyperOS handset `25053PC47G` for the OEM refusal and the original `-k` inode measurement. Every package mutated during the probes was restored to `default-state` and verified on both emulators.
- **The branch build runs on Android 17.** `app-foss-debug.apk` (`versionCode=1931`, `targetSdk=37`) installed on the Pixel 10 Pro Fold: `HomeActivity` resumes, no `AndroidRuntime` fatal, and the privilege probe reports `active=SHIZUKU` — the exact configuration this PR changes.

### DIFF
--- a/PROJECT_CONTEXT.md
+++ b/PROJECT_CONTEXT.md
@@ -88,10 +88,12 @@ for the presentation layer.
 - **Advanced Insights**: Installer source (resolved from package labels, not hardcoded), split APK
   indicators, version codes, SDK targets, `isSuspended`, `isDebuggable`.
 - **System App Support**: Uninstall or freeze system apps (requires any privilege mode). A freeze
-  disables the package and keeps its data; it removes the package for the current user only where
-  disabling is unavailable — Shizuku on Android 16+, and Dhizuku, whose gateway is not yet
+  disables the package and keeps its data; it removes the package for the current user
+  (`pm uninstall -k`, so the data directories survive) only where the platform actually *refused*
+  the disable — an OEM restriction, not an Android version, plus Dhizuku, whose gateway is not yet
   converted. `domain/model/FreezePolicy.kt` owns that rule; every reader of freeze state must
-  handle both, since devices carry system apps frozen the old way.
+  handle both, since devices carry system apps frozen the old way, when the fallback ran first,
+  unconditionally, and without `-k`.
 - **Security**: Biometric/device-credential lock for app access. Per-session authentication state.
 - **App Metadata Caching**: Room DB cache for `AppInfo`, invalidated via `lastUpdateTime`.
 - **Preferences** (`UserPreferences`): theme, AMOLED, dynamic color, biometric lock, sort/filter
--- a/PROJECT_CONTEXT.md
+++ b/PROJECT_CONTEXT.md
@@ -87,7 +87,11 @@ for the presentation layer.
   with multi-user support.
 - **Advanced Insights**: Installer source (resolved from package labels, not hardcoded), split APK
   indicators, version codes, SDK targets, `isSuspended`, `isDebuggable`.
-- **System App Support**: Uninstall or freeze system apps (requires any privilege mode).
+- **System App Support**: Uninstall or freeze system apps (requires any privilege mode). A freeze
+  disables the package and keeps its data; it removes the package for the current user only where
+  disabling is unavailable — Shizuku on Android 16+, and Dhizuku, whose gateway is not yet
+  converted. `domain/model/FreezePolicy.kt` owns that rule; every reader of freeze state must
+  handle both, since devices carry system apps frozen the old way.
 - **Security**: Biometric/device-credential lock for app access. Per-session authentication state.
 - **App Metadata Caching**: Room DB cache for `AppInfo`, invalidated via `lastUpdateTime`.
 - **Preferences** (`UserPreferences`): theme, AMOLED, dynamic color, biometric lock, sort/filter
--- a/PROJECT_CONTEXT.md
+++ b/PROJECT_CONTEXT.md
@@ -88,11 +88,16 @@ for the presentation layer.
 - **Advanced Insights**: Installer source (resolved from package labels, not hardcoded), split APK
   indicators, version codes, SDK targets, `isSuspended`, `isDebuggable`.
 - **System App Support**: Uninstall or freeze system apps (requires any privilege mode). A freeze
-  disables the package and keeps its data; it removes the package for the current user
-  (`pm uninstall -k`, so the data directories survive) only where the platform actually *refused*
-  the disable — an OEM restriction, not an Android version, plus Dhizuku, whose gateway is not yet
-  converted. `domain/model/FreezePolicy.kt` owns that rule; every reader of freeze state must
-  handle both, since devices carry system apps frozen the old way, when the fallback ran first,
+  disables the package and keeps its data. Only where the platform actually *refused* the disable
+  does it fall back to `pm uninstall -k --user N` — a **per-user removal, not a real uninstall**:
+  the APK stays on the read-only partition, the package record survives, and `-k` keeps both the
+  data directories and the runtime permission grants, so `pm install-existing --user N` restores
+  the app intact. That refusal is an OEM restriction, not an Android version. Dhizuku is the one
+  exception still on the old path: its gateway is not yet converted, so it removes for the current
+  user unconditionally and without `-k`. Note the fallback is unavailable to Shizuku at shell uid
+  on API 37, which answers it with `only root can delete system app for a particular user`.
+  `domain/model/FreezePolicy.kt` owns the rule; every reader of freeze state must handle both
+  mechanics, since devices carry system apps frozen the old way, when the removal ran first,
   unconditionally, and without `-k`.
 - **Security**: Biometric/device-credential lock for app access. Per-session authentication state.
 - **App Metadata Caching**: Room DB cache for `AppInfo`, invalidated via `lastUpdateTime`.
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@
 - **Extension Manager** — an in-app catalog of optional add-ons, each signature-verified and SHA-256 checked before install
 - **Redesigned Home** — an adaptive bento grid with one-tap access to the Extension Manager
 - **Universal Android Debloater (UAD) Integration** — safety recommendation chips (Recommended, Advanced, Expert, Unsafe) dynamically shown for system packages
-- **Safe System App Debloating & Freezing** — uninstalls system apps for the current user (`pm uninstall --user`) and restores them (`pm install-existing`) to support modern Android versions safely
+- **Safe System App Debloating & Freezing** — freezes a system app by disabling it (`pm disable`), which keeps its data, and only removes it for the current user (`pm uninstall --user`, which does not) where the platform allows nothing else: Shizuku on Android 16+, and Dhizuku. Unfreezing undoes either mechanic (`pm install-existing`, then `pm enable`)
 - **Adaptive UI Layouts** — vertical navigation rail for tablets/foldables, optimized viewport layouts, and split landscape detail screens
 - **Safety Gating** — blocks freezing of system apps marked as **Unsafe** by UAD to prevent bootloops, and warns on **Expert** packages
 - Root Support
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@
 - **Extension Manager** — an in-app catalog of optional add-ons, each signature-verified and SHA-256 checked before install
 - **Redesigned Home** — an adaptive bento grid with one-tap access to the Extension Manager
 - **Universal Android Debloater (UAD) Integration** — safety recommendation chips (Recommended, Advanced, Expert, Unsafe) dynamically shown for system packages
-- **Safe System App Debloating & Freezing** — freezes a system app by disabling it (`pm disable`), which keeps its data, and only removes it for the current user (`pm uninstall --user`, which does not) where the platform allows nothing else: Shizuku on Android 16+, and Dhizuku. Unfreezing undoes either mechanic (`pm install-existing`, then `pm enable`)
+- **Safe System App Debloating & Freezing** — freezes a system app by disabling it (`pm disable`), which keeps its data, and only removes it for the current user (`pm uninstall -k --user`, which keeps the data too) where the device itself refuses to disable system packages — some OEM builds do, whatever the Android version — plus Dhizuku, whose gateway is not yet converted. Unfreezing undoes either mechanic (`pm install-existing`, then `pm enable`)
 - **Adaptive UI Layouts** — vertical navigation rail for tablets/foldables, optimized viewport layouts, and split landscape detail screens
 - **Safety Gating** — blocks freezing of system apps marked as **Unsafe** by UAD to prevent bootloops, and warns on **Expert** packages
 - Root Support
--- a/app/src/main/java/com/valhalla/thor/data/freezer/AppFreezeStateReader.kt
+++ b/app/src/main/java/com/valhalla/thor/data/freezer/AppFreezeStateReader.kt
@@ -45,7 +45,7 @@ class AppFreezeStateReader(
         val info = packageManager.getApplicationInfo(packageName, MATCH_FLAGS)
         // MATCH_UNINSTALLED_PACKAGES is not optional. A *system* app is frozen with `pm disable`
         // where that works and with `pm uninstall --user N` where it does not — the gated
-        // destructive fallback (`destructiveFreezeFallbackAllowed`), plus every package the
+        // destructive fallback (`uninstallFreezeFallbackAllowed`), plus every package the
         // uninstall-only builds froze and that is still frozen today. A package in that second
         // state is not installed for this user, so the lookup throws NameNotFoundException without
         // the flag — the app then reads ABSENT and freezableCandidates drops it, which silently
@@ -103,7 +103,7 @@ class AppFreezeStateReader(
          * Both mechanics are live, and neither flag can be dropped. `pm disable` is now what
          * freezes a system app wherever the platform permits it, and `pm uninstall --user N` is
          * what freezes one where it does not — the gated destructive fallback, see
-         * `destructiveFreezeFallbackAllowed`. Even if that gate closed everywhere tomorrow,
+         * `uninstallFreezeFallbackAllowed`. Even if that gate closed everywhere tomorrow,
          * MATCH_UNINSTALLED_PACKAGES would stay load-bearing: devices carry system apps that the
          * uninstall-only builds froze, and they read as ABSENT without it, which is precisely how
          * `Unfreeze all` would lose the packages that most need unfreezing.
--- a/app/src/main/java/com/valhalla/thor/data/freezer/AppFreezeStateReader.kt
+++ b/app/src/main/java/com/valhalla/thor/data/freezer/AppFreezeStateReader.kt
@@ -21,7 +21,9 @@ import org.koin.core.annotation.Single
  *
  * [MATCH_FLAGS] covers both of Thor's freeze mechanics — MATCH_DISABLED_COMPONENTS for a
  * disabled package, MATCH_UNINSTALLED_PACKAGES for a system package uninstalled for the user —
- * and FLAG_SUSPENDED (API 24+) catches the suspend-mode case.
+ * and FLAG_SUSPENDED (API 24+) catches the suspend-mode case. Both mechanics are live: a system
+ * app is frozen with `pm disable` where the platform allows it and with `pm uninstall --user N`
+ * where it does not, and packages frozen the second way by earlier builds are still out there.
  */
 @Single
 class AppFreezeStateReader(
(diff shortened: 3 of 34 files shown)


trinadhthatakula:
## Two corrections since the first push — one of them reverses this PR's own gate

### 1. The Android 16 boundary does not exist

The gate shipped in `b59c1f64` as `sdkInt >= 36`. That was wrong, and it took a device to show it.

On a **stock AOSP Android 16 (API 36) emulator**, as uid 2000:

| command | result |
|---|---|
| `pm disable --user 0 com.android.egg` | `SecurityException: Shell cannot change component state for null to 2`, exit 255 |
| `pm disable-user --user 0 com.android.egg` | **exit 0**, `enabled=3`, `installed=true` |

Same on `com.android.printspooler`, `com.android.wallpaper.livepicker`, `com.android.traceur`. All four restored with `pm enable` and verified back at `enabled=1`.

The command that fails is `pm disable` — `COMPONENT_ENABLED_STATE_DISABLED` (state 2), which the shell uid has been forbidden from setting **since API 25**. That guard block in `PackageManagerService.setEnabledSettings` hashes identically (md5 `61926cfa2f445cb628f3e9ec190d66c0`) across `android14-`, `android15-`, `android16-`, `android16-qpr1-` and `android16-qpr2-release` and tag `android-16.0.0_r4`; the 15→16 diff of that method is `Trace.instant()` instrumentation and a metrics object.

The restriction that **is** real is Xiaomi's, and it was reproduced on `25053PC47G` (HyperOS OS3.0, `BP2A.250605.031.A3`):

```
$ pm disable-user --user 0 com.android.egg
java.lang.SecurityException: Cannot disable system packages.
    at com.android.server.pm.PackageManagerServiceImpl.shouldRestrictEnabledSettingsChange(...)
EXIT=255
```

`PackageManagerServiceImpl` 404s on `android.googlesource.com` at every tag 13→17 and on `main`; it lives in `/system_ext/framework/miui-services.jar`, in the same string pool as Xiaomi's own `"Cannot disable miui core packages."`. It was decompiled in Brevent #902 on **HyperOS/Android 14** in May 2024 — thirteen months before Android 16 shipped — and corroborated on Xiaomi hardware in Hail #255, AppManager #1557 and UAD-ng #1322. A GitHub-wide search for the exact string returns four issues, every one Xiaomi.

So the version gate was wrong in both directions: it would have changed mechanic on Pixels that could disable the app, and refused to freeze at all on the devices that are the entire reason the fallback exists. `6ebdaa17` replaces `sdkInt` with `disableRefusedByPolicy`, set only by a `SecurityException` from `PackageManagerService`.

### 2. The fallback never had to destroy data

`cf739fc0` adds `-k`. Measured on the same Xiaomi device: `pm uninstall -k --user 0` → `pm install-existing --user 0` returns the app with **byte-identical `ceDataInode` and `deDataInode`**.

Unmeasured caveat, stated as such in the code: runtime permission grants and app-ops are per-user package state and are expected **not** to survive the round trip. The app should come back with its data and without its grants.

`freezeSystemAppForUser` is a separate function from `uninstallApp` on purpose — adding `-k` to the shared one would silently make the real "uninstall this app" feature leave data behind.

Renamed `destructiveFreezeFallbackAllowed` → `uninstallFreezeFallbackAllowed`. Leaving a symbol called "destructive" on a path that measurably no longer destroys anything is the same stale-prose defect this PR is otherwise fixing.

### 3. Also corrected

- **Root's justification was too strong.** It claimed every refusal keys on uid 2000. `ProtectedPackages` (device provisioning package, DO/PO, DPM owner-protected) throws `Cannot disable a protected package` and is **not** uid-gated, so it refuses root identically — reported on Amazon Fire and Infinix XOS. The branch still answers `false`, now for a reason that survives contact with that case.
- **Struck "Shizuku on Android 16+"** from `README.md`, `PROJECT_CONTEXT.md` and `docs/follow-ups/shortcut-match-flags-system-apps.md`. It was never measured, and merging it would have left three documentation sites that a future search would read as independent corroboration of a claim this branch invented.
- **`Targets.B`** (`SDK_INT >= 36`) has zero references repo-wide — a dead constant from a bulk API-37 bump, not a mirror of anything.

### Still open

`pm suspend` was measured working on the refusing device and is strictly better than uninstall-for-user — data *and* launcher icon, no round trip. Not added here: Thor already exposes suspend as a user-chosen freeze mode, so substituting it inside *disable* mode is a product decision. Worth its own issue.

429 tests, 0 failures. `assembleFossDebug` and `lintFossDebug` green.

trinadhthatakula:
## Review response — 3 fixed, 2 rejected on measurement

### ✅ Fixed

**`ExtensionOpsProvider.kt:121` — apply the freeze safety gate to extension requests.** Correct, and the best finding on this PR. `Op.FREEZE` called `ManageAppUseCase.setAppDisabled`, whose own KDoc says it is *"the raw primitive, with no `FreezeTier` check"*. `FreezeAppUseCase` is the gate, and its KDoc already named this caller by name:

> *"a shortcut, an **extension trigger**, an automation intent or a widget would reach for the obvious entry point and hand a `packageName` straight to the freeze, which on an Unsafe system app means disabling a package the device needs to boot"*

An extension trigger was the one such caller still going around it. Signature attestation does not close this — a legitimately pinned extension still supplies an arbitrary package list, so being trusted buys the right to *ask*, not the right to freeze something that stops the device booting.

The documented reason the batch paths skip the gate does not transfer: they classify a whole target list against one shared snapshot, so routing them through it would mean a redundant per-package re-read and a second refusal for apps the user was already told about. An extension does no such filtering, so it gets the per-package resolve. Unfreeze stays on `forceUnfreeze` — it is the way *out* of a bad freeze, and a block that caught it would trap the app it protects. Default `FreezerMode.FREEZE` preserves the disable-only rule for extensions.

Not unit-tested at the provider level, stated plainly: a `ContentProvider` cannot be instantiated in a JVM test and this project carries no Robolectric (verified, not assumed). The routing is compiler-verified and `isBlockedFromFreeze` is already covered.

**`shortcut-match-flags-system-apps.md:105` — wrong line reference.** Correct, and worse than reported: the call is at line **141**, and the file is under `data/service/`, not `data/manager/`. Both references fixed and anchored to the enclosing `screenReceiver` rather than a bare line number.

**`PROJECT_CONTEXT.md:94` — clarify the removals are user-scoped.** Fair. "Removes the package for the current user" reads as an uninstall. Restated: the APK stays on the read-only partition, the package record survives, and `-k` keeps the data directories **and** the runtime permission grants, so `pm install-existing --user N` restores the app intact. Also now records that Dhizuku is the one gateway still on the old unconditional path.

### ❌ Rejected — both ask for a version gate that hardware refuted

**`FreezePolicy.kt:155` — "require an Android 16 system-package disable failure before destructive escalation."** This is the gate `6ebdaa17` removed, and it was removed because it is measurably wrong. On a **stock AOSP Android 16 (API 36) emulator** at uid 2000:

| command | result |
|---|---|
| `pm disable --user 0 com.android.egg` | `SecurityException: Shell cannot change component state for null to 2`, exit 255 |
| `pm disable-user --user 0 com.android.egg` | **exit 0**, `enabled=3`, `installed=true` |

Confirmed since on **Android 17 / API 37** (`CP31.260623.005`) — identical results, the failing guard now at `PackageManagerService.java:4206`. Disabling a system app from the shell uid works on both releases; the command that fails is `pm disable` (state 2), forbidden to shell **since API 25**. That is a *command* constraint, not a version one, and the Shizuku path already satisfies it by sending `pm disable-user`.

A version test is wrong in both directions: it would change mechanic on a Pixel that could disable the app, and refuse to freeze at all on the OEM builds (Xiaomi HyperOS, first reported on **Android 14**) that are the entire reason the fallback exists. `disableRefusedByPolicy` asks whether the platform actually refused, which is the thing that varies.

**`RootSystemGateway.kt:418` — "confirm device support for `pm disable` on protected system packages."** Confirmed, and the answer is already in the code: `ProtectedPackages` is **not** uid-gated and refuses root identically (`Cannot disable a protected package: <pkg>`), which is why the root branch answers `false` rather than escalating. A package the platform protects that hard is one where "remove it for the user instead" is the wrong reading of the refusal.

### 📌 New measurements since the last push (`4524387d`)

Android 17 turned up two things, one of which corrected this PR's own body.

**The uninstall fallback does not exist at shell uid on API 37.** Identical command, same uid 2000, AOSP emulators:

```
pm uninstall -k --user 0 com.android.egg
  API 36 -> Success, exit 0
  API 37 -> Failure [only root can delete system app for a particular user], exit 1
```

The package is left untouched, so the chain ends in an honest `Result.failure` rather than a wrong state. A Shizuku at uid 2000 on Android 17 stops at "could not disable"; a Shizuku started as root is unaffected. Documented at the call site with an explicit note not to add a rung below it.

**Runtime permission grants survive `-k` after all.** Five sites in this branch said they were "expected not to". Measured on API 36:

```
pm grant com.android.egg READ_EXTERNAL_STORAGE  -> granted=true
pm uninstall -k --user 0                        -> Success
pm install-existing --user 0                    -> Success
READ_EXTERNAL_STORAGE                           -> granted=true, flags unchanged
ceDataInode / deDataInode                       -> byte-identical
```

`-k` retains the whole `PackageUserState`, not just the data directories. The pessimistic guess was stated in two user-visible log strings as if it were fact; all five sites corrected.

**And the PR body had the API 37 required-packages guard backwards.** It claimed the new restriction was *inert* because shell holds zero occurrences of `ALLOW_CONTROL_SYSTEM_REQUIRED_PACKAGES`. Holding that permission is the **bypass** — holding none means shell is subject to it:

```
pm disable-user --user 0 com.google.android.packageinstaller
-> SecurityException: Cannot disable required package
   at PackageManagerService.setEnabledSettings(PackageManagerService.java:4218)
```

It fires for a small allowlist (`packageinstaller`, `permissioncontroller`) and not for `settings`, `systemui`, `gms`, `nexuslauncher` or `intentresolver`, all of which still disable normally. It is a `SecurityException`, so `isPolicyRefusal` already classifies it correctly and the existing gate handles it with **no code change**.

All packages mutated during these probes were restored to `default-state` and verified on both emulators.

---

429 tests, 0 failures. `assembleFossDebug` and `lintFossDebug` green.

PR #319, opened 2026-08-01
feat(web): the Thor landing page, with a claims gate that reads the built HTML

trinadhthatakula:
Adds `web/` — an Astro static site for Thor, deployed separately and sharing nothing with the Gradle build but the repository it lives in. Eight routes: home, features, download, FAQ, privacy, extensions policy, build-an-extension, and a 404.

The app's own typefaces are self-hosted so the site and the app read as one thing, and `src/lib/tokens/` is the single source for colour, spacing, type scale and motion.

## The part worth reviewing

This site's entire argument is "check it yourself", so a confidently wrong sentence costs more than a missing one. Most of the work here is the machinery that stops one shipping.

`check:claims` reads the **built HTML** and fails the build on fourteen named claims — things that were once true of Thor and are not any more, or that are true of one build flavour and get written as though they were true of both. Each rule carries its rationale, the source that settles it, and a pair of fixtures asserting it both fires on the tempting wrong phrasing and stays silent on the corrected one. A meta-test rejects a rule whose fixtures are missing, whose `fail` fixture fires on a bare phrase with no surrounding prose, or whose `pass` fixture is shorter than its `fail`.

Alongside it: `check:links` resolves every internal link and fragment against the built output, so a renamed heading breaks the build rather than the page; `check:markup` and `check:sitemap` cover the structural invariants; `check:a11y` runs axe over all eight pages. `src/lib/repo-facts/` reads minSdk, targetSdk and the version from `libs.versions.toml` and `gradle.properties` at build time, so the numbers on the site cannot drift from the ones the app compiles with.

## One real bug found in the gate itself

`toSentences` treats every newline as a sentence boundary and `extractText` pushed text nodes through verbatim. The prose here is hard-wrapped at about a hundred columns, and that wrapping reaches `dist` **inside** the paragraph's text node — so a claim straddling a wrap was split in half, no pattern spanning the break could match it, and the rule went unenforced on that sentence with the build still green.

Which claims escaped was decided by where the editor happened to wrap. On the built privacy page, "Thor's own code makes exactly one kind of network request" was arriving as two fragments, and the rule written for that exact sentence never saw it.

Fixed by collapsing whitespace inside text nodes, leaving only the block markers the extractor pushes deliberately. `pre` is exempt — a snippet's newlines are the code's own line structure, and joining them would invent text that appears nowhere on the page. Three regression tests cover it.

## C14, the rule the bug was hiding

There are two flavours and they do not contain the same code: `storeImplementation` pulls in Google Play Billing, and `BillingProcessorImpl` is a `@Single` whose `init` calls `connectToBilling()` while `MainScreen` injects it as a default parameter — so the Play build opens a connection to Google **when the app opens**, not when you go looking for the support option.

That makes "no third-party SDK" and "exactly one network request" false for half the readers unless the sentence names a scope, and false for exactly the half who did not choose their build for privacy reasons. C14 fires on the bare absolute and stays quiet when the sentence says which build, or says "Thor's own code". The fix is never to weaken the claim — "the FOSS build has no third-party SDK" is stronger than the unqualified form, because a reader can act on it.

## Content correctness pass

Every page was verified line-by-line against the app's source rather than against its documentation. Roughly ninety claims were corrected. The ones worth naming:

- **Signing is a build-type property, not a flavour property.** `signingConfigs.create("release")` is applied by the release build type and neither `productFlavors` block mentions signing, so both release APKs carry the same key. What differs is the channel: CI uploads an app bundle with `skip_upload_apk`, so Play generates and signs its own device APKs. "Google re-signs it" describes a step that does not exist, and the site was telling readers to verify a signature themselves — a reader who believed the flavour picked the signer would have compared the wrong two things and concluded the build was tampered with.
- **`AutomationExtension` still ships** in thor-extension-api 3.0.0, in both jars, and the template's entry class still implements it. What Thor retired is the dispatch, not the type. Telling an author it is gone is actionable in the wrong direction: they delete a class their template still implements, and the build failure looks unrelated.
- **`pm uninstall -k` does not preserve the whole `PackageUserState`.** It keeps the data directories; `installed` still goes false, which is why every query in the freeze path passes `MATCH_UNINSTALLED_PACKAGES`. The pages that hedged around this were written against a `FreezePolicy.kt` predating #318 and have been settled against the merged source.
- **No `ACTION_SEND` intent-filter exists** in any source set, so the FAQ's "share it to Thor instead" workaround could not have worked.
- **Indus Appstore dropped**, and the FAQ's "four channels" now matches the three it actually lists.

## Self-review, since CodeRabbit skipped this PR

CodeRabbit declined it — *"Review skipped: 135 files exceed the limit of 100"* — so nothing had reviewed the branch. I ran an adversarial pass over the load-bearing parts instead. Six findings survived being refuted; every one of them is fixed in `7da71243`, and every fix ships with the gate that would have caught it. Each gate was then mutation-tested — the original bug re-injected, the gate watched go red — because a gate that cannot fail is worse than no gate: it is trusted.

Three of the six were live bugs on the deployed site:

- **The footer linked the wrong Telegram.** `t.me/thor_app_updates`, a handle that appears nowhere else in this project's history, on all eight pages — while three of the site's own pages linked `t.me/thorAppDev` in prose. The same deployment contradicted itself and every gate was green, because `check-links` skips other origins by design, `check-claims` never reads an `href`, and `t.me` answers 200 with a generic page for an unknown handle, so even the scheduled `lychee` sweep would have called it healthy. `site.test.ts` now derives the footer's outbound links from the repository: both in-app links, the README, the issue chooser, `shizu_store.json`, and all five entries of `FUNDING.yml`.
- **Every code sample rendered in one flat colour.** Shiki's `css-variables` theme writes `color:var(--astro-code-token-function)` inline with **no fallback**, and Astro declares those names only inside its dev error overlay — 121 references, none resolvable. An undeclared custom property is not a CSS error; it is invalid at computed-value time, so `color` falls back to `inherit` and the page renders. `tokens.css` maps all six to existing palette tokens, `contrast.test.ts` pins each above 4.5:1 in dark, light and AMOLED, and `check-markup` gained a whole-build assertion that every `var()` in `dist` resolves. `token-usage.test.ts` walks `src/` and structurally cannot see a property that only exists after a build.
- **`ThemeToggle` removed `data-theme` to mean "system"** — the one operation `ThemeScript` spends a paragraph forbidding, because AMOLED is keyed on `[data-theme='dark'][data-amoled='true']` and that cannot match without a concrete value. Latent rather than live, since nothing on the site writes `thor-amoled` yet. It now resolves the OS preference itself, which is what makes it responsible for the `change` listener too — the CSS media query no longer gets the chance, and without the listener system mode would freeze at whatever the OS said on load. `theme-runtime.test.ts` pins the contract between the two scripts, which until now was held together by a comment.

The other three were gates that could not fail, or documentation asserting something untrue:

- **`check-sitemap` never compared the sitemap to the build.** All four of its assertions were about the sitemap in isolation, and all of them hold on a sitemap describing a site that does not exist — deleting `/faq` from it left the gate reporting `OK — 6 page URLs`. The root cause was in the fixtures: they contained no HTML at all, sitemap files floating in a void. Both directions are checked now, and the fixtures model builds.
- **The claims gate could not read a meta description.** `extractText` started at `document.body`, so the sentence Google prints under the search result and Telegram prints in the unfurl card was unreachable by all 15 rules — despite being read by more people than the page body is.
- **`fonts.css` described a preload block `BaseLayout` did not ship**, and prescribed weight 900, which `tokens.css` declares to mirror `Type.kt` but which no rule on the site references — preloading it would have spent bandwidth on a file no page requests. `BaseLayout` now preloads the three faces genuinely above the fold, and the hashes are confirmed identical to the ones the `@font-face` rules request, so it removes a round trip rather than doubling a download.

Twelve further candidates were refuted and are not in the commit.

One repo change outside `web/`: I created the **`area: website`** label. `labeler.yml` has referenced it since this branch opened and it did not exist, so PR Labeler would have failed on every future `web/**` PR.

## The second review

You posted a review from OpenAI GPT-5 (Codex): five findings, two P1 and three P2. Every one of them was real. **Three of the three P2s had already been fixed** by the self-review above, independently and with the same correction — the `matchMedia` listener and the recomputed `data-amoled` in `ThemeToggle`, and the both-directions sitemap cross-check with `404.html` and disallowed routes exempt. Agreement between two reviewers that did not see each other is the strongest signal either of them produced.

The two P1s were new, and both are fixed here:

- **An unscoped permission claim on the home page.** "…permissions all stay" was standing on one runtime permission, measured once, on one platform build, with app-ops never measured at all. The mechanic underneath it (`pm uninstall -k` → `DELETE_KEEP_DATA`) makes no promise about the permission table. The sentence now says what was measured and names what was not. This is C14's failure mode again — a true-in-the-case-we-checked claim written as an absolute — so it gets C14's treatment: **rule C15** forbids the unscoped phrasing on every page and stands aside for the scoped one. Mutation-tested by putting the shipped sentence back: `[C15] forbidden claim: "permissions all stay"`. Writing the rule also caught `claims/C1/pass` making the same unscoped claim while demonstrating a different rule — a `pass` fixture asserts zero violations from *any* rule, which is precisely how.

- **Ten placeholder screenshot frames.** I could not capture these at the time — they need a device, the app installed, and your authorisation to change device state — so I gated them instead. (They are all filled now; see below.) The placeholders are deliberate (`<Image>` resolves at build time, so requiring a capture would have broken `astro build` from the day each page was written) and `docs/launch-checklist.md` already carried the grep. The problem is that the checklist has no moment to run in: **Vercel's production branch is `dev`, so merging this PR is the production deploy.** A manual gate in front of an automatic action is a gate in front of an open door. `check:screenshots` now runs on every build and counts, but fails only when `VERCEL_ENV=production` — placeholders stay green locally, in CI, and on preview deploys, where seeing the unfilled slots on a real URL is the point. `CI` and `NODE_ENV` are both the wrong switch and the tests pin that. Mutation-tested in both directions: forcing the env check false reddens three tests, and rewording `DeviceFrame`'s aria-label reddens the guard that ties the regex to the string the component actually emits.

## The screenshots

All ten frames are filled. `check:screenshots` now reports `0 placeholders` under `VERCEL_ENV=production`, so the gate above has gone from "will refuse the deploy" to "permits it".

**Six captures, not the five the checklist named.** It had no entry for the Extension Manager frame on `privacy.mdx` — the one doing evidentiary work for *"this screen, and only this screen, makes a network request"*.

Five came off your phone. Slot 4 came off the emulator with Shizuku uninstalled (pulled first, reinstalled after), because its alt text asserts *"on a device with no privilege mode enabled"* and the phone's 79 frozen apps badge every list row — a badge is proof of privilege, so the phone cannot photograph that claim without contradicting it. That is the only frame not from the physical device.

**Two alts were wrong before a pixel existed**, and got corrected rather than weakened:

- The hero promised "active, frozen and **suspended**" counts. `SummaryStatRow` drops the Suspended tile when `suspendedCount == 0`, and the phone had none — so the frame would have shown two tiles under an alt claiming three. Suspended three disposable packages, shot 202 / 79 / 3, unsuspended them, verified `dumpsys package | grep -c 'suspended=true'` back to `0`.
- `features.mdx` promised a bento grid *"of device and app counts"*. The grid holds quick actions; the counts are a separate row above it, and the install-source chart is below. Rewritten to the order the screen renders.

Slot 4's alt now names the **Camera** chip it is filtered by. An unfiltered list on a bare emulator sorts alphabetically into RRO overlay packages — `2 Button Navigation Bar`, `android.auto_generated_rro_vendor__` — which shows the feature without demonstrating it.

`web/docs/screenshot-checklist.md` is rewritten against what the session actually found. The SysUI demo-mode step is gone — your call, and the right one; a real clock and battery read as a phone somebody uses. Each slot now carries the trap that nearly broke it: the Freezer list is only the watchlist and `toggleManaged(add = true)` freezes *before* it adds, so the active/frozen contrast has to be made by unfreezing something already on the list; Work Mode is an inline button group, not a dialog, and a tap persists immediately; `AppRiskDialog`'s BLOCKED branch emits no confirm button at all, while `com.android.settings.intelligence` is Expert-rated and shows a live "Freeze Anyway" — the destructive look-alike that is easy to shoot by mistake.

Both devices are restored and the restore verified: Shizuku reinstalled, apps unsuspended, the unfrozen app back to `enabled=2`, the phone's list filters put back.

## Owner decisions

- **Indus Appstore — deferred, on your call.** `README.md` keeps the badge; the site continues not to mention it. The two disagree on purpose for now.
- **Play tracks — you've asked for `internal` → `alpha` and `alpha` → `beta`.** That is being handled separately from this PR, since it changes `fastlane/Fastfile` and the release workflows rather than anything under `web/`.

## Checks

`npm run build` (types → build → links → claims → markup → sitemap → screenshots), `npm test` (**291 tests, 16 files**) and `check:a11y` (8 pages, 38 axe rules, 0 failing nodes) all pass. `astro check` is at 0 errors, 0 warnings, 0 hints. `dev` is merged in; the merge was clean.

Gate output after both rounds of review fixes:

```
check-links:       OK — 8 pages, 182 internal links, 62 fragments, 119 external skipped
check-claims:      OK — 16 rules, 8 pages
check-markup:      OK — 8 pages, 78 custom properties resolved
check-sitemap:     OK — 2 sitemap files, 7 page URLs, 7 built pages cross-checked
check-a11y:        OK — 8 pages, 38 axe rules, 0 failing nodes
check-screenshots: OK — 8 pages, 0 placeholders (strict: production)
```

That last line is the one that was blocking: it read `10 placeholders (advisory)` and would have refused the production deploy. The run above is with `VERCEL_ENV=production` set, which is the switch Vercel will set on merge — so the deploy this PR triggers is one that has already been asserted green.


web/src/pages/index.astro
```diff
@@ -0,0 +1,164 @@
...
+    <p>
+      Thor freezes them: they stop running, stop starting at boot, and leave the launcher. Unfreeze
+      whenever you want. It works through root — or through Shizuku or Dhizuku, neither of which
+      needs root. <strong>You will need one of the three.</strong>
+    </p>
+    <p class="source-line">
+      Source: <a href={SITE.repo}>github.com/trinadhthatakula/Thor</a>
+    </p>
+
+    <DeviceFrame slot="media" alt="Thor's home screen, showing the active, frozen and suspended app counts above the action tiles." />
```
trinadhthatakula: **[P1] Replace the production screenshot placeholders before merge**

Every production `DeviceFrame` call currently omits `src`, so this branch renders ten visible `Screenshot pending` frames across the homepage, Features, and Privacy pages. The launch checklist itself identifies five required captures and explicitly says the green build does not mean this step is finished. Because `dev` is documented as Vercel's production branch, merging in this state can deploy an obviously incomplete landing page. Add and import the five captures, then pass them into every corresponding frame.

web/src/pages/index.astro
```diff
@@ -0,0 +1,164 @@
...
+    </p>
+    <p>
+      And a home-screen shortcut Thor pinned for an app doesn't vanish when you freeze it — the icon
+      turns grey, and tapping it unfreezes the app and opens it.
+    </p>
+    <p>
+      Your data stays where it is. Freezing switches an app off; it doesn't take anything away. On
+      the handful of devices that refuse to let a preinstalled app be switched off, Shizuku mode
+      removes it for your Android user instead, with the keep-data flag — the app's files, settings
+      and granted permissions all stay, so unfreezing brings it back the way you left it. Root has
```
trinadhthatakula: **[P1] Scope the permission-preservation claim to the evidence**

`granted permissions all stay` is a blanket guarantee that the repository evidence explicitly does not establish. `FreezePolicy` and the corrected Features copy say only that one shell-granted runtime permission survived on one API 36 build; user-granted permissions and app-ops were not measured. This headline copy should use that same scoped wording rather than promising that every grant survives on every supported platform.

web/src/components/ThemeToggle.astro
```diff
@@ -0,0 +1,83 @@
...
+        stored = localStorage.getItem('thor-theme')
+      } catch (_) {
+        // Private browsing and some embedded webviews throw on any access.
+        // Falling back to "system" is correct: the OS preference still applies.
+      }
+      return order.indexOf(stored) === -1 ? 'system' : stored
+    }
+
+    function paint(mode) {
+      if (mode === 'system') root.removeAttribute('data-theme')
```
trinadhthatakula: **[P2] Preserve the resolved theme attribute in system mode**

The head script deliberately resolves system mode to a concrete `data-theme`, because AMOLED styling requires `[data-theme='dark'][data-amoled='true']`. This line immediately removes that attribute once the body loads, disabling AMOLED in system-dark mode and causing the documented black-to-near-black flash. `paint()` also never recalculates `data-amoled` while cycling modes. Keep storage empty for system mode, but retain a resolved DOM theme, update it on `matchMedia` changes, and recompute AMOLED whenever the resolved theme changes.

web/src/lib/site.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface ExternalLink {
+  href: string
+  label: string
+}
+
+export const PROJECT_LINKS: ExternalLink[] = [
+  { href: SITE.repo, label: 'Source on GitHub' },
+  { href: `${SITE.repo}/issues`, label: 'Issue tracker' },
+  { href: `${SITE.repo}/releases`, label: 'Releases' },
+  { href: 'https://t.me/thor_app_updates', label: 'Telegram' },
```
trinadhthatakula: **[P2] Point the footer at the project's actual Telegram channel**

`thor_app_updates` currently resolves to Telegram's generic contact page, while the README, `shizu_store.json`, both in-app links, and the site's own Privacy/FAQ/Download pages consistently use `https://t.me/thorAppDev` (whose page title is `Thor Dev`). The scheduled external checker cannot catch this because the wrong handle still returns HTTP 200. Use the canonical `thorAppDev` URL here.

web/scripts/check-sitemap.mjs
```diff
@@ -0,0 +1,162 @@
...
+          `names ${loc}, which is not in the build`,
+          'A sitemap index that points at a missing child is the same silent failure as a ' +
+            'missing index: the crawler gets a 404 and moves on.',
+        )
+        continue
+      }
+      pending.push(childRel)
+    }
+  }
+  counts.urls = pageUrls.length
```
trinadhthatakula, line 169: **[P2] Cross-check sitemap URLs against the built page set**

The gate only requires a non-zero page count and checks `Disallow` conflicts; it never proves that sitemap URLs resolve in `dist` or that every built page is listed. Deleting `/faq` from the generated sitemap still reports success with six URLs, and adding an orphan `/handbook` URL also passes. Compare both sets in both directions (excluding `404.html` and intentionally disallowed routes) so the checker enforces the indexing guarantee described by this module.

trinadhthatakula (commented):
## Code review

Reviewed the PR head against `dev` (`015f083..2d36073`) in an isolated checkout.

**Review model:** OpenAI GPT-5 (Codex)

I found **5 actionable issues**: 2 P1 release blockers and 3 P2 correctness gaps. The inline comments contain the evidence and recommended correction. This is submitted as a comment-only review and does not alter the PR approval state.

Verification performed: `npm test` (254 tests), full Astro build and content gates, accessibility check, GitHub Actions status, and CodeQL.

PR #320, opened 2026-08-01
chore(ci): move both Play uploads up one track (internal→alpha, alpha→beta)

trinadhthatakula:
Moves both Play uploads up one rung, per your call: `internal` → `alpha`, `alpha` → `beta`.

| Workflow | Branch | Lane | Track before | Track after |
|---|---|---|---|---|
| `dev-check.yml` | `master` | `distribute_dev` | `internal` | **`alpha`** (Closed testing) |
| `production-deploy.yml` | `production` | `distribute_production` | `alpha` | **`beta`** (Open testing) |

Neither lane has ever written `production`, and neither does now — promoting a build to the Production track stays a manual Play Console action. That manual step is your "check each stable build" gate, and this PR does not touch it.

## What else changed, and why it isn't just a rename

The two-string swap would have left one thing lying. `dev-check.yml`'s Telegram caption hardcoded `Track: Internal Testing` — a sentence about the upload that the upload could not affect. Editing those two words re-arms the same trap on the next track change, and there is no actionlint, yamllint or shellcheck over `.github/workflows`, so nothing in the repo can notice a caption drifting from the lane it describes.

So fastlane now writes `track.txt` next to the `version_name.txt` / `version_code.txt` it already writes, and the workflow derives the label from it:

```sh
TRACK=$(cat track.txt 2>/dev/null || echo "")
case "$TRACK" in
  internal)   TRACK_LABEL="Internal Testing" ;;
  alpha)      TRACK_LABEL="Closed Testing" ;;
  beta)       TRACK_LABEL="Open Testing" ;;
  production) TRACK_LABEL="Production" ;;
  "")         TRACK_LABEL="unknown" ;;
  *)          TRACK_LABEL="$TRACK" ;;
esac
```

The default arm prints the raw key rather than inventing a friendly name, so an unrecognised track shows up as itself instead of as a plausible lie. The step is gated on `steps.ver.outputs.release == 'true'` — the same condition under which `distribute_dev` ran — so the empty arm is defence, not a live path.

`prepare_release_artifacts` also loses its `track:` default. It defaulted to `'internal'`, which meant `build_release_candidates` — a lane that uploads nothing — logged `Track: internal` for a build that went nowhere. It is now `nil` plus a fail-closed allowlist:

```ruby
if upload_to_store
  allowed_tracks = %w[internal alpha beta]
  track = track.to_s.strip
  unless allowed_tracks.include?(track)
    UI.user_error!("❌ upload_to_store: true needs track: one of #{allowed_tracks.join(', ')} — got #{track.inspect}")
  end
end
```

`production` is deliberately off the list. Promoting to production is a manual Console action and that manual step is the release gate, so a lane able to reach it from CI would silently remove the gate — the comment saying "nothing here ever writes production" is now enforced rather than asserted. A typo lands on the wrong side of the check too, which matters because `upload_to_play_store` would otherwise *create* a track by that name instead of failing.

Inert on both current upload paths (each passes a valid track explicitly). Exercised over all nine input shapes — three valid tracks, `nil`, `''`, `'  '`, `'production'`, `'aplha'`, and the non-upload lane.

One more line went with it, found in review: the caption also said `Uploaded to Play Store & GitHub`, but `Create GitHub Pre-Release` is a **later step in the same job** — so it announced a release that had not been attempted, and would have said the same had it failed. It now reads `Uploaded to Play Store — GitHub pre-release next`.

## What this does not change

- **The two-branch version-code collision** (`docs/follow-ups/two-branches-one-play-version-code.md`) stays `Status: OPEN`. Play's version-code uniqueness is **per app, not per track**, and it constrains *uploads*, not track membership — so moving both tracks is exactly neutral to it. The doc's table, its three options and its acceptance criterion are re-pointed at `alpha`/`beta`; the problem itself is untouched.
- **`release_status: 'completed'`** — unchanged and still correct. `draft` breaks the shared helper; `rollout` needs `inProgress`.
- **The site.** `web/download.mdx` says CI "uploads an Android App Bundle to Play's closed-testing track" — still true after this change, since the master lane now goes to `alpha` = closed testing.
- **The Telegram and GitHub pre-release channels**, which never went through Play.

## Verification

- `ruby -c fastlane/Fastfile` → Syntax OK
- Both workflow YAMLs parse
- `bash -n` on the extracted Telegram `run:` block → OK
- Track guard exercised over all nine input shapes
- CI green on the head SHA: `build-and-test`, `Analyze (java-kotlin)`, `CodeQL`, `shizu-manifest`, `label`
- No app code touched, no `versionCode` change

## Play Console actions this PR cannot do

Code can point at a track; it can't configure one. Before the next `chore(release)` lands:

1. **Open testing needs countries/regions configured**, or the `beta` upload succeeds and reaches nobody.
2. **Existing internal testers stop receiving builds.** Nothing publishes to `internal` any more — move them onto the closed-testing tester list.
3. **Confirm the closed-testing tester list exists** and has you on it, since that is now where every `master` release lands.
4. **Keep doing the manual Production promotion** — that is unchanged and remains the release gate.


trinadhthatakula:
Took two of three.

**dev-check.yml:225 — caption over-claims GitHub.** Valid and in scope: `Create GitHub Pre-Release` is a *later* step in the same job, so the caption announced a release that had not been attempted and would have said the same had it failed. Fixed to `Uploaded to Play Store — GitHub pre-release next`, which is what is actually true at send time. This is the same defect class the PR exists to fix, one line further down.

**Fastfile — validate the track, not just its presence.** Valid, and it makes the PR's own claim enforceable. The guard only rejected `nil`, so `''`, `'  '` and a typo'd `'aplha'` all reached `upload_to_play_store`, which *creates* a track by that name rather than failing. Now an allowlist of `internal|alpha|beta` against the stripped value, with `production` deliberately excluded — promoting to production is a manual Console action and that manual step is the release gate, so a lane able to reach it from CI would silently remove the gate. The comment saying "nothing here ever writes production" is now enforced instead of asserted. Exercised over all nine input shapes (three valid tracks, nil, empty, whitespace, production, typo, and the non-upload lane).

**Follow-up doc — rollout date asserted as past.** Declined as stated: the date was not in the future, the review clock was a day behind it. But the underlying worry is fair — a date in a doc drifts from the change it describes — so the parenthetical now cites PR #320 instead of a date, which cannot go stale.

`ruby -c`, both workflow YAMLs, and `bash -n` on the extracted Telegram `run:` block all pass.

PR #322, opened 2026-08-02
ci(web): publish from master, stage from dev — and correct what #321 said about Vercel

trinadhthatakula:
Two changes to the deploy pipeline shipped in #321, in the same paragraphs.

1. **The site's production branch moves from `dev` to `master`**, so day-to-day site work lands on `dev` like everything else and the live site changes only on a deliberate release merge.
2. **Corrections to what #321's docs asserted about Vercel**, several of which were wrong in ways that would have cost the owner a broken first deploy.

Companion PR **#323** puts `web/` on `master`, which the first change requires. **Merge this one first** — #323 carries byte-identical copies of two `paths-ignore` blocks introduced here, and they only stay identical in that order.

---

## 1. `master` publishes, `dev` stages

| Trigger | Before | After |
|---|---|---|
| push to `master` | *(no trigger)* | **production** |
| push to `dev` | production | preview |
| PR into `master` | preview | preview, production-strict |
| PR into `dev` | preview | preview |
| `workflow_dispatch` from `master` | — | production |

`dev` is kept as a push trigger on purpose rather than dropped. A per-PR preview only ever shows one branch; the `dev` preview is the only URL that shows several merged site PRs *together*, which is the state the release merge is about to publish.

### This closes a real gap

The workflow now sets `REQUIRE_SCREENSHOTS=1` when `github.base_ref` is `master`. `check:screenshots` refuses a placeholder frame when `VERCEL_ENV=production` **or** `REQUIRE_SCREENSHOTS=1` (`web/scripts/check-screenshots.mjs:48-50`), so the release PR's preview is held to the production standard *before* the merge that publishes.

Previously the first build ever held to the production standard was the one already serving on the live domain. A placeholder would have been found by a visitor, not by a check.

### And opens one, stated plainly

If the Vercel project ever gets connected to the repository, a `master` push now starts **two production deploys racing for the same alias** — this workflow's gated artifact, and a Git-sourced Vercel build from the Gradle repository root, which has no `package.json`. Whichever finishes last wins.

Before this PR it was an *uncontested overwrite* — worse in outcome, but at least deterministic. Either way the answer is the same and it is the one the docs now lead with: do not connect the repository. There is no Production Branch value that makes a connected project safe; Vercel's default is `master`, which is precisely the dangerous one.

All six trigger shapes were exercised against the extracted step:

```
EVENT             REF                 BASE     OUTPUTS
push              refs/heads/master   —        env=production flag=--prod strict=1
push              refs/heads/dev      —        env=preview    flag=       strict=
pull_request      refs/pull/9/merge   master   env=preview    flag=       strict=1
pull_request      refs/pull/9/merge   dev      env=preview    flag=       strict=
workflow_dispatch refs/heads/master   —        env=production flag=--prod strict=1
workflow_dispatch refs/heads/feat/x   —        env=preview    flag=       strict=
```

### `paths-ignore` on the Android workflows

`web/**` was added to `dev-check.yml` and `production-deploy.yml` on the reasoning that the site adds no Gradle module and cannot change a byte of the APK. That reasoning covers three paths the glob does not: `/vercel.json`, `.github/workflows/web-*.yml` and `.github/labeler.yml`.

Not hypothetical — #323 carries all three, so without this the merge that first puts the site on `master` would have run the full Android build.

---

## 2. Corrections to #321's Vercel documentation

Researched against Vercel's live documentation and the `vercel@58.4.4` CLI source, with adversarial verification. Five things were wrong or missing.

**The kill switch was oversold.** `web/vercel.json`'s `"git": { "deploymentEnabled": false }` was documented as *the* guarantee against double deploys. Vercel does not document which `vercel.json` its Git integration reads for that flag — it is evaluated at webhook time, before any build container exists, so the CLI's `join(cwd, rootDirectory)` rule verified in source does not transfer. `vercel/vercel#11176` ("deploymentEnabled is ignored") has been open since November 2024 with independent confirmations.

Fixed by adding `/vercel.json` at the repository root so the flag exists in both plausible locations, and by rewriting the docs around the guarantee that *is* structural: **do not connect the repository**. The CLI is unaffected either way — it reads `join(cwd=web, '.')/vercel.json` and never the root copy.

**How the project gets created was never written down**, and it is the single most likely way a first deploy fails. Importing the repo from `vercel.com/new` sets Root Directory to `web`, installs the webhook, and writes `framework: astro` into the project settings. This workflow already runs the CLI from inside `web/`, so Root Directory `web` resolves to `web/web` (`packages/cli/src/commands/build/index.ts`; `vercel/vercel#16749`). Worse, it does not fail the way the repo advertises — with `framework: astro` pulled it dies at `Command "astro build" exited with 127`, which names nothing about paths, *before* reaching the "Gate the staged artifact" tripwire written to catch exactly this. `cd web && vercel link` leaves Root Directory unset and connects nothing.

**"Production Branch — irrelevant, setting it does no harm" was backwards.** Corrected twice: first to name `dev` as the dangerous value, and now — since production moved — to state that the *default* is the dangerous one and no safe value exists.

**`ignoreCommand` was called inert. It is not provably inert.** `vercel deploy` sends `projectSettings.commandForIgnoringBuildStep` on the same payload as `--prebuilt`, and Vercel has shipped then rolled back a build where the Ignored Build Step cancelled prebuilt deployments. It fails loudly — `vercel deploy` prints "The deployment has been canceled." and exits 1 under `set -euo pipefail`. A comment on the deploy step now says **never add `--no-wait`**, which is the one flag that would convert that from loud to silently green.

**The domain guidance coached the owner through an error that cannot occur.** A Cloudflare 526 requires Cloudflare in the request path; the `thor` record is grey-clouded. What the owner will actually see, verified live:

```console
$ openssl s_client -connect thor.trinadhthatakula.com:443 -servername thor.trinadhthatakula.com \
    </dev/null 2>/dev/null | openssl x509 -noout -subject -dates
subject=CN=*.trinadhthatakula.com
notBefore=Feb 10 12:08:05 2026 GMT
notAfter=May 11 12:08:04 2026 GMT        # expired

$ curl -sSI -k https://thor.trinadhthatakula.com | head -3
HTTP/2 404
server: Vercel
x-vercel-error: DEPLOYMENT_NOT_FOUND
```

Vercel's edge already answers for the hostname and falls back to a `*.trinadhthatakula.com` wildcard that expired 83 days ago. It cannot renew a wildcard without DNS-01 and the zone is on Cloudflare nameservers, so it will not self-heal. Two years of HSTS and a 308 on plain HTTP mean there is no way to click through in a browser. `ERR_CERT_DATE_INVALID` here means "no project has claimed this hostname yet" — **not** that the DNS record is wrong, which is the diagnosis that would break the working record. The checklist now gives a browser-free verification order: `domains inspect` → `certs ls` → `openssl`, expecting `CN=thor.trinadhthatakula.com` rather than the wildcard.

Also corrected: the first credentialed run should be a `workflow_dispatch` from `master`, not a test PR — Vercel makes the first deployment of a new project production regardless of `--prod`, so whatever runs first is what the domain attaches to. The natural instinct (add secrets, push a test PR to see if it works) is the one that puts a preview-environment build on the production alias.

---

## Verification

291 tests / 16 files pass. `npm run build` clean — 8 pages, 16 claim rules, 78 custom properties, 182 internal links, 0 failing axe nodes. `REQUIRE_SCREENSHOTS=1 npm run check:screenshots` reports `strict: production`, 0 placeholders. The workflow YAML parses and all 13 `run:` blocks pass `bash -n`.

No site content changes — this is the workflow, three markdown docs, and the two Android `paths-ignore` lists.

### DIFF
--- a/.github/workflows/dev-check.yml
+++ b/.github/workflows/dev-check.yml
@@ -20,7 +20,15 @@ on:
       # web-ci.yml already gates it. Its own build reads gradle.properties and
       # gradle/libs.versions.toml, which are deliberately NOT ignored here, so a
       # chore(release) bump still runs everything.
+      #
+      # The site's plumbing is listed alongside it because 'web/**' alone does not
+      # cover it, and a push carrying only site plumbing would therefore still run
+      # this whole workflow. That is not hypothetical: the merge that first puts the
+      # landing page on master carries all four of these.
       - 'web/**'
+      - 'vercel.json'
+      - '.github/workflows/web-*.yml'
+      - '.github/labeler.yml'
       - 'LICENSE'
       - '.gitignore'
       - '.github/**/*.md'
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -17,7 +17,14 @@ on:
       # Gradle module, and cannot reach the APK or the Play upload. It has its own
       # gate in web-ci.yml and its own deploy in Vercel. gradle.properties stays out
       # of this list, so a release push still builds.
+      #
+      # The three entries after 'web/**' are the site's plumbing, which that glob
+      # does not cover — without them a push carrying only site plumbing would run
+      # a full build and a Play upload to prove nothing.
       - 'web/**'
+      - 'vercel.json'
+      - '.github/workflows/web-*.yml'
+      - '.github/labeler.yml'
       - 'LICENSE'
       - '.gitignore'
       - '.github/**/*.md'
--- a/.github/workflows/web-deploy.yml
+++ b/.github/workflows/web-deploy.yml
@@ -13,13 +13,20 @@ name: Web Deploy
 # in git. See web/docs/deploy.md.
 #
 # ────────────────────────────────────────────────────────────────────────────
-# THE ONE THING THAT MUST STAY TRUE: `web/vercel.json` sets
-#   "git": { "deploymentEnabled": false }
-# If that is ever removed while this workflow exists, EVERY web commit deploys
-# twice — once by Vercel's Git integration and once by this workflow — and the
-# two race for the production alias. The loser is whichever finishes first, so
-# the symptom is the site intermittently serving the older of two commits.
-# Nothing reports this. The two settings are one decision.
+# THE ONE THING THAT MUST STAY TRUE: the Vercel project is NOT connected to the
+# GitHub repository. Create it with `cd web && vercel link`, never by importing
+# the repo — importing also sets Root Directory to `web`, which resolves to
+# `web/web` because this workflow already runs the CLI from inside web/.
+#
+# `"git": { "deploymentEnabled": false }` is defence in depth, and it is in TWO
+# files — web/vercel.json and /vercel.json — because Vercel does not document
+# which one its Git integration reads. Delete neither.
+#
+# If the repo does get connected: Vercel's production branch here is `master`,
+# so a dev push is a Git-sourced *preview* and there is no alias race. The
+# hazard is the first dev→master release merge, when a master push becomes a
+# real Vercel production deploy built from the Gradle repo root, with no run of
+# this workflow competing. Not a race — an uncontested overwrite.
 # ────────────────────────────────────────────────────────────────────────────
 #
 # Path filter: identical to web-ci.yml's, and for the same reason. The site
@@ -248,7 +255,7 @@ jobs:
           out=".vercel/output/static"
 
           if [ ! -f "$out/index.html" ]; then
-            echo "::error::$out/index.html is missing, so the artifact about to be deployed is not the site. The usual cause is a Root Directory mismatch — this job already runs inside web/, so the Vercel project's Root Directory must not also be 'web'. See web/docs/deploy.md."
+            echo "::error::$out/index.html is missing, so the artifact about to be deployed is not the site. The usual cause is a Root Directory mismatch — this job already runs inside web/, so the Vercel project's Root Directory must be UNSET, not 'web'. Note that this check is not the first thing a Root Directory mismatch trips: if the project was created by importing the repo, its pulled settings carry framework=astro and the Build step dies earlier with 'Command \"astro build\" exited with 127', which names nothing about paths. See web/docs/deploy.md."
             ls -la .vercel/output 2>/dev/null || echo "(no .vercel/output at all)"
             exit 1
           fi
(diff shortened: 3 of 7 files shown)


trinadhthatakula:
Thanks — three of four were right and are fixed in `f95492e4`. One declined.

**Fixed: fork PRs get no preview.** Correct, and it matters more than "Major" suggests. `web-deploy.yml`'s deploy job skips itself when `github.event.pull_request.head.repo.full_name != github.repository`, so a fork PR gets no preview URL, no comment, and — since this PR makes a PR into `master` the place production strictness is asserted — no strict screenshot check either. The trigger table now has a fork row, both PR rows say "from a branch in this repository", and the docs state the backstop: the `master` push re-runs every gate strictly *before* it uploads, so a placeholder fails the deploy red with the previous deployment still serving. Caught after the merge rather than before it.

**Fixed: `#11176` date.** `gh api repos/vercel/vercel/issues/11176` → `created_at: 2024-02-19T16:00:53Z`. Now reads "19 February 2024".

**Fixed: the certificate check, though not the way suggested.** You are right that the identity lives in `subjectAltName` and that reading the subject is wrong. But "accept either the explicit hostname or a matching `*.trinadhthatakula.com` wildcard" would invert the check: the wildcard *is* the failure case here. It is the stale cert Vercel's edge already falls back to, it expired on 11 May 2026, and it does match `thor.trinadhthatakula.com` under RFC 6125 — which is exactly why the old CN note was fragile. The step now reads the SAN and requires the host listed **explicitly**, with the note that dates alone are also insufficient, since a renewed wildcard would pass a date check and still not be this project's certificate. Verified against the live host:

```console
$ openssl s_client -connect thor.trinadhthatakula.com:443 -servername thor.trinadhthatakula.com \
    </dev/null 2>/dev/null | openssl x509 -noout -ext subjectAltName -dates
X509v3 Subject Alternative Name:
    DNS:*.trinadhthatakula.com, DNS:trinadhthatakula.com
notBefore=Feb 10 12:08:05 2026 GMT
notAfter=May 11 12:08:04 2026 GMT
```

**Fixed: "skips invisibly".** It emits a notice naming the dashboard path. Reworded.

---

**Declined: pinning every documented CLI call to `vercel@58.4.4`.**

The major pin is deliberate and `web-deploy.yml` says why at the install step: patch releases within a Vercel CLI major are usually upload fixes and are wanted; a major bump is not something to discover during a release. Also `npm view vercel@58 version` → `58.4.4` today, so an exact pin changes nothing now and freezes a stale patch later.

The legitimate half of the concern is reproducibility of the behavioural claim, and that is addressed directly. The Root Directory section now names the version the source was read from and states both halves of the condition — the `web/web` path needs the CLI running from `web/` **and** a Root Directory of `web`; either alone is fine.

On the related web-search result about `VERCEL_RESOLVE_ROOT_DIRECTORY` being enabled by default: not in the version this pipeline runs. `vercel/vercel#16749` is `state: closed`, `merged_at: null`, and the string does not appear anywhere in the published `vercel-58.4.4.tgz`. `join(cwd, project.settings.rootDirectory || '.')` in `packages/cli/src/commands/build/index.ts` stands. The docs now record that, and note that if a later CLI does land the re-anchoring fix, the mismatch stops being a build failure and starts being silently fine — which changes nothing about the advice and everything about what the tripwire catches.

PR #327, opened 2026-08-02
fix(web): let Vercel's Git integration actually deploy, and defer the Actions path

trinadhthatakula:
## The thing to read first

**Both `vercel.json` files told Vercel not to auto-deploy.** `"git": { "deploymentEnabled": false }` was committed at the repository root *and* in `web/`, and Vercel documents a bare `false` as the way to stop every branch triggering a deployment upon commits. As committed, this repo asked for exactly the opposite of what you set up.

The live site does not contradict that, and the timeline is airtight rather than inferred:

| | |
|---|---|
| last push to any branch | `refs/heads/master` → `42630d54`, **18:13:01Z** |
| first Vercel commit status anywhere in the repo | **18:26:48Z** ("Vercel is deploying your app") |
| Vercel commit statuses on any *earlier* commit | **zero** |

The Vercel project did not exist when that push happened, and no push webhook fired at 18:26 — a webhook posts "pending" within seconds, not after 13 minutes. The deployment was started from the dashboard at import. It is a fully Git-sourced deployment (`vercel[bot]` wrote a GitHub Deployment record, and the `thor-git-master-*` branch alias exists), just not a push-triggered one.

**Nothing has been pushed to any branch since.** So "the flag is honoured" and "the flag is ignored" are observationally identical right now — both predict exactly one deployment. This PR removes the flag, which is correct under either hypothesis and costs nothing under the one where it was already being ignored.

## What changed

- **`web/vercel.json`** — the `git` block removed. Everything else untouched.
- **`/vercel.json`** — deleted. Its entire content was that flag. With Root Directory set to `web`, `web/vercel.json` is the copy Vercel reads; the second copy only existed because the Actions design left Root Directory unset.
- **`ignoreCommand` kept, and its status inverted.** It was documented as inert; it is now the live path filter. Exit 0 means *skip*. Its `:/` prefixes are load-bearing — the command runs inside the Root Directory, so a bare `gradle.properties` would resolve to `web/gradle.properties`, match nothing, exit 0 for every commit, and report "Build skipped", correctly, forever.

## What the docs now say

`web/docs/deploy.md` is rewritten around the model that is actually running. `web/README.md`, `web/docs/launch-checklist.md` (§5 and §6 become done-state records) and `web-deploy.yml`'s banner follow.

The four dashboard settings survive — not as an argument against the Git integration, but as the risk inventory, because they are now load-bearing and none of them is visible in a diff:

| Setting | Correct value | How it fails |
|---|---|---|
| Root Directory | `web` | unset → no `package.json`, build fails. **The exact inverse of what the Actions path needs.** |
| Include files outside Root Directory | on | `repo-facts` cannot read the two Gradle files — fails *loudly* |
| Build / Install / Output override | none | a dashboard Build Command replaces `npm run build` and all six checkers stop running, silently |
| System Environment Variables | enabled | `VERCEL_ENV` empty → `check:screenshots` downgrades from gate to advisory **on the production build** |

The honest mitigation is stated too, because it narrows the original argument for the Actions path: `web-ci.yml` runs the same gate chain from a file in git on every web push and PR, so a dashboard edit cannot make the gates vanish. It can only ship an artifact CI never gated.

**One property did not survive, and the docs say so rather than claiming it.** Nothing sets `REQUIRE_SCREENSHOTS=1` on a PR into `master` any more — that was the Actions pipeline. The production build is now the first strict one. A failed build leaves the previous deployment serving, so the failure mode is a stale site and a red deployment, not a placeholder on the live domain.

## The follow-up

`docs/follow-ups/vercel-actions-deploy.md`, filed at Tier 2 with a row in the index. It carries the reactivation order and the hard interlock:

> Turning the Actions path on **requires** turning Vercel's Git integration off in the same change. Both production branches are `master`, so leaving both on races two production deploys at one alias — with no error anywhere and no symptom except the live site intermittently serving the older build.

It also records that Vercel resolved this project's production branch to `master` by its *second* rule (there is no `main`), not because `master` is the default branch. The coincidence invites the wrong inference.

## Verification

- 291 tests pass (16 files).
- `npm run build` green through all six gates: `check:types`, `astro build`, `check:links`, `check:claims`, `check:markup`, `check:sitemap`, `check:screenshots`.
- `web-deploy.yml` parses; `web/vercel.json` parses.
- Live cert re-checked: SAN is exactly `thor.trinadhthatakula.com`, Let's Encrypt, valid to 31 Oct 2026, `HTTP/2 200`.

## What is still on you

**Two things.**

1. **This needs a second PR to reach production.** `master` is what Vercel publishes from, and `dev` and `master` currently have identical trees — so a `dev` → `master` release PR after this merges would carry *only* this change. `versionCode` is unchanged at 1931, so no Play release fires.

2. **The first real push is still untested.** No commit has reached any branch since the project existed, so the first `master` push touching `web/` is the first time the trigger path runs at all. Watch the Deployments list: **exactly one row** is correct.

   Test with a commit that **touches `web/`**. An Android-only commit is masked twice over — by `ignoreCommand`, and by Vercel's "Skipping unaffected projects" setting — and would read as a false confirmation. And if a row appears in state `Canceled`, that is `ignoreCommand` doing its job, not the old flag: `deploymentEnabled: false` creates no row at all. That distinction is the whole diagnostic.

One dashboard read worth doing while you are there: Settings → Build & Deployment → Root Directory should say `web`.

## Summary by CodeRabbit

* **Deployment**
  * Vercel Git Integration is now the documented deployment path.
  * The `master` branch publishes production releases; other branches generate previews.
  * Deployment guidance now covers build filtering, project settings, root directory configuration, rollbacks, and failure visibility.
* **Documentation**
  * Updated launch and deployment checklists with current verification steps, screenshot enforcement, certificate checks, and troubleshooting guidance.
  * Added follow-up documentation for a potential future migration to GitHub Actions.
* **Maintenance**
  * Removed obsolete deployment configuration and updated related deployment references.


trinadhthatakula:
**Update — the discriminating test ran, and it passed.**

Pushing this branch *was* the test the PR body said had never been run. Vercel built it and posted `Vercel — success — Deployment has completed`, aliased at `thor-git-fix-vercel-git-autodeploy-…vercel.app` (302, i.e. behind Deployment Protection — a nonexistent alias 404s, so the 302 proves the alias exists).

Three things fall out of that one build succeeding, with no dashboard read needed:

- **Root Directory is `web`.** A build from the repo root finds no `package.json` and fails.
- **"Include files outside the Root Directory" is on.** `/download` renders the `minSdk` fact, which `repo-facts` derives from `gradle/libs.versions.toml` — a file above `web/`.
- **`ignoreCommand` returned non-zero** and let the build proceed, which is correct for a commit touching `web/`.

Production is unaffected: `thor.trinadhthatakula.com` still 200.

Pushed `3eaaf3ad` to record it. What remains genuinely unobserved is now narrower — the first *production* push. Everything above was a preview.

trinadhthatakula:
Reviewed all ten. **Nine applied in `eaee35b9`, one declined.**

### The Major one was right, and it was mine

`web-deploy.yml:35` — the banner told the reader to clear Root Directory to empty *and* to restore `git.deploymentEnabled` to `web/vercel.json` because "with Root Directory `web` that is the copy Vercel reads". Those contradict: which `vercel.json` Vercel reads follows Root Directory, so once it is cleared the file a reconnected integration reads is `/vercel.json` — the copy this PR deleted. Anyone following the instruction would have put the kill switch in a file nothing reads, and it would have looked set.

Taken the second of your two options: **disconnecting the repository is now the only documented interlock**, in all three places that discussed the flag (`web-deploy.yml`, the follow-up doc, `deploy.md`). The flag survives as defence in depth with its correct location stated as "whichever file matches the Root Directory then in force".

### Applied

| # | File | What changed |
|---|---|---|
| 1 | `web-deploy.yml:13` | "never run" → "has never deployed" |
| 2 | `web-deploy.yml:35` | above |
| 3 | follow-up:72 | "same commit" → "same activation change", noting the dashboard half lands first |
| 4 | follow-up:94 | `#11176` reframed as an unresolved compatibility risk — Vercel's docs and the issue can both be true if the bug is conditional — with a no-op-commit observation to run before relying on it |
| 5 | follow-up:99, 162-166 | acceptance is now path-scoped, specifies a real one-line change under `web/` instead of an empty commit, and defines the expected out-of-scope result (no deployment, or one `CANCELED`) |
| 7 | `deploy.md:126` | success and failure observed; *skipped* is now stated as unobserved, consistent with the required-checks section below it |
| 8 | `launch-checklist.md:88` | "enforced, not just listed" → **unresolved** until the System Environment Variables box is inspected. It was a claim about a setting nobody has looked at |
| 9 | `launch-checklist.md:180` | `ERR_CERT_DATE_INVALID` now means what it means — date validation failed — with provisioning failure and DNS mismatch listed alongside the unclaimed-hostname case, and SAN/dates/issuer/DNS/domain-state checks before acting. The "do not touch DNS" advice survives, attached to its real reason |
| 10 | `README.md:81` | both preview claims qualified by the deploy paths; a commit touching none of them is `CANCELED`, which is most Android commits |

### Declined: #6, the `ignoreCommand` multi-commit window

The gap is real and I am not disputing it — `deploy.md` already documented it before this review, at the bullet directly above the line you flagged. What I disagree with is the remedy.

- **The command is `-- .`, not `-- web`.** It runs inside the Root Directory, so `.` *is* `web/`. Minor, but the quoted form suggests a repo-root evaluation that would behave differently.
- **The complete push range is not available.** Vercel does not expose the push base to `ignoreCommand`, and the clone is `--depth=10`. There is no correct range to substitute.
- **"Build every push" is the failure this filter exists to prevent** — it rebuilds the site on every Android commit, which is nearly all of them.
- **The blast radius is a missing preview, never a stale production site.** Nothing pushes to `master` except merges, and a merge commit's `HEAD^` is the previous `master` tip, so the diff spans everything merged.

`deploy.md` now states that bound explicitly and names the condition that would end it: a direct multi-commit push to `master`.

Web gates re-run after the edits: 291 tests pass, full `npm run build` chain green.

PR #328, opened 2026-08-02
docs(web): the screenshot gate is enforced — its setting was confirmed on

trinadhthatakula:
You enabled **"Enable access to System Environment Variables"** in Vercel today. That was the one §5 item #327 could not resolve, and the reason §4 had to hedge its own gate as *possibly* advisory. This unhedges it.

### What changes

`check:screenshots` is a gate again rather than a hopeful one. `VERCEL_ENV` is now injected, `isProductionDeploy()` returns true on the production build, and a placeholder frame fails it instead of shipping.

- `web/docs/launch-checklist.md` §4 — "currently unconfirmed" → **enforced**
- `web/docs/launch-checklist.md` §5 — the checkbox is ticked, dated, with the reason it needed checking rather than assuming (Vercel doesn't document the default for a portal import)
- `web/docs/deploy.md` — the same correction in the project-settings list

### What I did not claim

The **setting** is confirmed. The **behaviour** is not, and the docs now say which is which.

No production build has run since the box was turned on, so nothing has observed `VERCEL_ENV=production` actually reaching `check-screenshots.mjs`. The first release merge proves it — and proves it *by passing*, which is indistinguishable from the downgraded case unless a placeholder happens to be present at that moment. That is a weak proof and is labelled as one, because "it went green" is exactly what the failure mode also looks like.

`REQUIRE_SCREENSHOTS=1 npm run build` locally remains the only way to see the strict verdict on demand.

### Verification

291 tests pass, full `npm run build` gate chain green. Docs-only otherwise.

## Summary by CodeRabbit

* **Documentation**
  * Updated deployment guidance to reflect the verified system environment variable setting.
  * Added UTC/date interpretation guidance and instructions for checking `VERCEL_ENV` in the first production build log.
  * Clarified that a successful build alone does not confirm production propagation, which remains unobserved.
  * Documented the screenshot gate’s strict and advisory behaviors in the launch checklist.

### DIFF
--- a/web/docs/deploy.md
+++ b/web/docs/deploy.md
@@ -114,8 +114,10 @@ can only report on it afterwards.
   System Environment Variables". `check:screenshots` is strict only when `VERCEL_ENV=production`, and
   Vercel injects that variable. If the box is off, `VERCEL_ENV` is empty, `isProductionDeploy()`
   returns false, and the screenshot gate silently downgrades from gate to advisory **on the
-  production build**. Vercel does not document whether it is on by default for a portal import.
-  Verify it once, in one click.
+  production build**. Vercel does not document whether it is on by default for a portal import, so it
+  was checked rather than assumed: **confirmed on, 2026-08-03**. Note what that does and does not
+  establish — the setting is on; no production build has run since, so `VERCEL_ENV` reaching the
+  build has not been observed.
 - **Deployment Protection.** A protected preview URL returns 401 with no notice explaining why. Every
   branch and every PR now gets a preview URL posted by the Vercel bot, so this is more visible than
   it was.
--- a/web/docs/deploy.md
+++ b/web/docs/deploy.md
@@ -115,9 +115,12 @@ can only report on it afterwards.
   Vercel injects that variable. If the box is off, `VERCEL_ENV` is empty, `isProductionDeploy()`
   returns false, and the screenshot gate silently downgrades from gate to advisory **on the
   production build**. Vercel does not document whether it is on by default for a portal import, so it
-  was checked rather than assumed: **confirmed on, 2026-08-03**. Note what that does and does not
-  establish — the setting is on; no production build has run since, so `VERCEL_ENV` reaching the
-  build has not been observed.
+  was checked rather than assumed: **confirmed on, 2026-08-02** (UTC). Note what that does and does
+  not establish — the setting is on; no production build has run since, so `VERCEL_ENV` reaching the
+  build has not been observed. A passing production build will not establish it either: with no
+  placeholders left there is nothing for the strict path to fail on, so it exits 0 either way. The
+  evidence is the `check-screenshots` line in the build log — `0 placeholders (strict: production)`
+  means the variable arrived, `0 placeholders (advisory)` means it did not and the gate downgraded.
 - **Deployment Protection.** A protected preview URL returns 401 with no notice explaining why. Every
   branch and every PR now gets a preview URL posted by the Vercel bot, so this is more visible than
   it was.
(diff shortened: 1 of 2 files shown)


trinadhthatakula:
Both findings addressed in 557eed20. Neither was declined, but the first one's *diagnosis* was wrong even though its *fix* was right, and the second turned out to be more actionable than the finding suggested. Details below.

---

### 1. "Use a valid confirmation date in both documents" — applied, reasoning corrected

**Applied.** All three sites now read `2026-08-02`.

The stated reason does not hold, though, and it is worth recording why so this does not get re-litigated. The date was not future-dated. The repo owner is on IST (UTC+5:30), and the commit that introduced it is authored `2026-08-03 02:37 +0530` — i.e. `2026-08-02 21:07 UTC`. On the owner's wall clock it genuinely was 3 August when the Vercel setting was ticked, so `2026-08-03` was a true statement about a past event, not a claim about the future.

The actual defect is a **convention mismatch inside one document**, which is what made it look future-dated:

- §6 quotes the certificate verbatim: `notBefore=Aug  2 17:29:49 2026 GMT` — UTC.
- The push-deploy verification was recorded as `2026-08-02` in 3eaaf3ad, authored `2026-08-03 01:36 +0530`. Same local-midnight-crossing situation, resolved the other way.

So two events **one hour apart in the same sitting** were labelled a day apart. That reads as a gap in the timeline that never existed. `2026-08-02` is correct under the convention the document already uses everywhere else, so both occurrences move, and the intro now states the convention explicitly:

> Dates here are **UTC**, anchored to the certificate timestamps quoted in §6. The launch work ran past local midnight in the owner's timezone, so a `git log` timestamp for the commit that recorded an item can read a day later than the item's date. That is the timezone, not a gap in the timeline.

That way the next reader who cross-checks a date against `git log` and sees a `+0530` timestamp a day later does not conclude the doc is wrong.

---

### 2. "Do not treat a passing production build as proof of propagation" — applied, and the signal you hypothesised already exists

**Applied, and this one was a real error, not just a hedging problem.** The old text read:

> The first release merge is what proves it, and it proves it by passing — which is indistinguishable from the downgraded case unless a placeholder is present.

Two separate things wrong with that.

**A pass proves nothing, and the trailing hedge does not rescue it.** All six screenshots have shipped, so `counts.placeholders` is `0`. The strict branch maps an empty `pending` array to an empty `failures` array and exits 0. The "unless a placeholder is present" escape hatch can never fire at release time, because there will not be a placeholder. The exit code is a constant.

**But the two cases are *not* indistinguishable** — which is where the sentence was factually incorrect rather than merely over-claiming. Your comment says strictness is only demonstrated "unless the build produces an observable signal". It does. `check-screenshots.mjs` labels its counts line off the same `strict` flag the gate branches on:

```js
[strict ? 'placeholders (strict: production)' : 'placeholders (advisory)', counts.placeholders]
```

and `report()` writes that line to stdout on the **success** path, not just on failure. So a zero-placeholder run still announces its mode. Verified locally against the built `dist`:

```console
$ VERCEL_ENV=production node scripts/check-screenshots.mjs dist
check-screenshots: OK — 8 pages checked, 0 placeholders (strict: production)
exit=0

$ VERCEL_ENV=preview node scripts/check-screenshots.mjs dist
check-screenshots: OK — 8 pages checked, 0 placeholders (advisory)
exit=0
```

Same exit code, different line. And on Vercel the reading is unambiguous in one direction: `REQUIRE_SCREENSHOTS` is the only other input to `isProductionDeploy()`, and nothing sets it in the Vercel environment, so `strict: production` appearing in a Vercel build log can only have come from `VERCEL_ENV`.

Both documents now say to read that line in the first `master` deploy's build log, and say plainly that a green checkmark is not the observation. The load-bearing nuance the previous commit was protecting — *setting confirmed, behaviour not* — is preserved, and is now actionable instead of being declared unobservable.

---

### Gates

Run from `web/`:

- `npm run build` — exit 0. `check-links` 8 pages / 182 internal links / 62 fragments; `check-claims` 16 rules over 8 pages; `check-markup` 8 pages / 78 custom properties; `check-sitemap` 2 sitemap files / 7 URLs; `check-screenshots` 8 pages / 0 placeholders.
- `npm test` — exit 0. **16 test files, 291 tests passed.**

PR #329, opened 2026-08-03
fix(apps): keep the app list intact when a ROM revokes package visibility in the background

trinadhthatakula:
## The bug

Users on Chinese-market ROMs (MIUI/HyperOS, and the xiaomi.eu builds that blend the China and Global variants) report that after leaving Thor backgrounded — switching to a RAM-hungry app, or just locking the phone for a while — the app list comes back showing **one entry: Thor itself**. It stays that way until Thor is restarted.

## Why

`com.android.permission.GET_INSTALLED_APPS` — a **non-AOSP** runtime permission from the Chinese association standard **T/TAF 108-2022**, implemented by MIUI/HyperOS, ColorOS, OriginOS and MagicOS. On those ROMs package visibility is a **three-state** permission (allow / while in use / deny) that applies *in addition to* `QUERY_ALL_PACKAGES`.

These ROMs hand out **"while in use"** by default. In that state `getInstalledPackages()` returns a near-empty list the moment Thor loses foreground. The background scan then looks exactly like *every app on the device was uninstalled at once* — and the cache prune faithfully deletes them all.

`QUERY_ALL_PACKAGES` is **not** the problem and is **not** touched here. It is still required on every Android version. The manifest diff is a pure addition; both existing elements are byte-identical.

### 1. Declare the permission
```xml
<uses-permission android:name="com.android.permission.GET_INSTALLED_APPS" />
```
Silent no-op on AOSP, where the permission simply does not exist.

### 2. Refuse to prune against a scan that can't be believed

`scanVerdict()` (pure function, `domain/model/InstalledAppsVisibility.kt`) inspects the scanned package **names** and returns `Accept` or `Retain(reason)`. On `Retain`, rows are kept, updates still applied, and the cached entries are re-emitted so the UI keeps showing them.

Three suspect shapes:

| Reason | Test |
|---|---|
| `EmptyScan` | scan returned nothing |
| `PlatformPackageMissing` | `"android"` not in the results |
| `Collapsed` | scan is under half the cache |

The `"android"` check is **identity-based, not count-based**, on purpose. Google's automatic-visibility guarantees mean the platform package is visible to *every* app on *every* version regardless of filtering — so its absence is **proof** about the scan, not a heuristic about it. A device with 120 of 175 packages visible passes a count threshold and still fails this one.

Escape hatches, in this order:
- `cachedCount == 0` → always `Accept` (first run must be able to populate).
- Permission **`Denied`** → `Retain` **indefinitely** (checked *before* the tolerance counter — the whole point is that this device's scans stay untrustworthy).
- Otherwise, **two** consecutive suspect scans are tolerated, then `Accept` — so a genuinely small device, or a real mass uninstall, still converges.

### 3. Offer the grant

A banner appears **only** while the permission is `Denied`, requested through the standard `ActivityResultContracts.RequestPermission()`.

Availability is probed with `getPermissionInfo()` **before** any of this, and there is deliberately **no rationale check and no "open Settings" fallback**:

> `shouldShowRequestPermissionRationale()` returns a hard `false` for a permission the platform has never defined ([`PermissionManagerServiceImpl`](https://cs.android.com/android/platform/superproject/+/main:frameworks/base/services/core/java/com/android/server/pm/permission/PermissionManagerServiceImpl.java): `if (permission == null) { return false; }`). That makes every Pixel read as `denied && !shouldShowRationale` — which the standard recipe misinterprets as *"permanently denied → send them to Settings"*, and it nags forever, pointing at a toggle that isn't there.

The availability probe is what makes the rationale question **unnecessary** rather than merely skipped. `AppListScreen.kt:224-236` documents this at the call site so it doesn't get "fixed" back in later.

The launcher's result boolean is **ignored** and the state re-read instead — a "while in use" grant answers `true` and then stops being true. Re-read on every `ON_RESUME` too, since a grant made in system Settings never comes back through the callback.

## Architecture

Binder calls stay in `data/`, decisions are pure functions in `domain/model/`. `AppListViewModel` takes an `InstalledAppsPermissionGate` port (`domain/repository/`) so `AppListViewModelTest` builds on the JVM with a hand-written fake — matching how every other dependency in that ViewModel is wired, and consistent with the no-Robolectric / no-mocking-library policy.

The gate is **non-nullable with no default**, on purpose: a nullable-with-default param would let a Koin misconfiguration present as *"no device supports this"* — a silent, permanent, invisible failure. `koinCompiler`'s safety checks turn a missing binding into a build error instead.

`declared` is cached (a ROM's permission table can't change under a running process); the **grant** is re-read every time (that's the part that changes).

## Verification

- `:app:compileFossDebugKotlin :app:compileFossDebugUnitTestKotlin` → **BUILD SUCCESSFUL**, no new warnings
- `:app:testFossDebugUnitTest --rerun-tasks` → **443 tests, 0 failures, 0 errors, 0 skipped** (45 suites; counts parsed from the JUnit XML, not inferred from `BUILD SUCCESSFUL`)
- `:app:lintFossDebug` → **BUILD SUCCESSFUL** — all five locale files carry the three new strings, so `MissingTranslation` stays quiet under `warningsAsErrors = true`
- `git diff app/src/main/AndroidManifest.xml` → pure addition, `QUERY_ALL_PACKAGES` byte-unchanged

**14 new tests**: 11 in `InstalledAppsVisibilityTest` (every verdict, reason precedence, the tolerance boundary, the `Denied`-before-tolerance ordering, `cachedCount == 0`), 3 in `AppListViewModelTest` (a device that doesn't define the permission never reaches the banner state; a denied grant surfaces; a re-read after the dialog picks up the new answer).

## Not verifiable here

No Chinese ROM in hand. The Pixel path (`Unsupported` → banner never shown, nothing requested) is fully covered by tests and by the 443-test suite passing. The MIUI path rests on the T/TAF 108-2022 semantics above; worth a device check by anyone who has one.

## Summary by CodeRabbit

* **New Features**
  * Added installed-app visibility detection across supported devices.
  * Added a warning banner with an action to grant full installed-app access.
  * Added localized messages in Arabic, Chinese, English, French, and Spanish.

* **Bug Fixes**
  * Prevented incomplete or unexpectedly reduced scans from removing previously discovered apps.
  * App visibility status now refreshes when returning to the app.
  * Improved handling of denied or unavailable access across different device environments.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,6 +22,15 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="ReservedSystemPermission" />
 
+    <!--
+      Non-AOSP. Declared by Chinese-market ROMs (MIUI/HyperOS, ColorOS, OriginOS, MagicOS)
+      per T/TAF 108-2022, where package visibility is a three-state runtime permission
+      (allow / while-in-use / deny) *in addition to* QUERY_ALL_PACKAGES above. On a ROM that
+      grants it "while in use", getInstalledPackages() collapses once Thor is backgrounded.
+      Absent on AOSP, where requesting it is a silent no-op — never assume it exists.
+    -->
+    <uses-permission android:name="com.android.permission.GET_INSTALLED_APPS" />
+
     <uses-permission android:name="android.permission.REQUEST_DELETE_PACKAGES" />
 
     <uses-permission android:name="shizuku.permission.API_V23" />
--- a/app/src/main/java/com/valhalla/thor/data/permission/InstalledAppsPermissionChecker.kt
+++ b/app/src/main/java/com/valhalla/thor/data/permission/InstalledAppsPermissionChecker.kt
@@ -33,12 +33,30 @@ class InstalledAppsPermissionChecker(
 ) : InstalledAppsPermissionGate {
 
     /**
-     * Cached, because the ROM's permission table is fixed for the life of the OS image — whether
-     * `com.android.permission.GET_INSTALLED_APPS` exists cannot change under a running process, and
-     * re-asking on every scan would put a binder call on the app-list hot path for an answer that is
-     * a build-time property of the device.
+     * Cached once resolved, because the ROM's permission table is fixed for the life of the OS
+     * image — whether `com.android.permission.GET_INSTALLED_APPS` exists cannot change under a
+     * running process, and re-asking on every scan would put a binder call on the app-list hot path
+     * for an answer that is a build-time property of the device.
+     *
+     * Deliberately *not* `by lazy`, because only a **definitive** answer may be cached. A caught
+     * `NameNotFoundException` is definitive — the device does not define the permission — and so is
+     * a successful lookup. Anything else is a failed question rather than an answer, and memoizing
+     * it would be self-defeating: as [declaredPermission] notes, the ROMs that define this
+     * permission are the same ones that throw unexpected things out of the package manager, so one
+     * unlucky probe would pin [InstalledAppsPermission.Unsupported] for the rest of the process and
+     * silently disable the banner, the prompt, and the unconditional-retain path in `scanVerdict` —
+     * on exactly the devices this exists for. A transient failure is left uncached and re-asked on
+     * the next [state] call instead.
+     *
+     * `@Volatile` for publication, not mutual exclusion: two threads racing the first probe may both
+     * make the binder call, which is wasteful for one round trip and otherwise harmless, since the
+     * answer they write is the same.
      */
-    private val declared: DeclaredPermission? by lazy { declaredPermission() }
+    @Volatile
+    private var resolved: Resolution? = null
+
+    /** A definitive answer about this device. [permission] is null when the device does not define it. */
+    private class Resolution(val permission: DeclaredPermission?)
 
     /**
      * The current state, re-reading the *grant* every time.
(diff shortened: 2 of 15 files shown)


trinadhthatakula:
Both review findings were real and are fixed in `1128ed84`.

**1. Transient `getPermissionInfo()` failures were memoized as "unsupported"** — confirmed, and worse than it looks. Only `NameNotFoundException` means "this device does not define the permission"; the broad catch treated every other throwable the same way. The code's own comment concedes the ROMs that *define* this permission are the ones most likely to throw something unexpected out of the package manager — so one unlucky probe would pin `Unsupported` for the life of the process and silently switch off the banner, the prompt, **and** the unconditional-retain path in `scanVerdict`, on exactly the devices the fix exists for. Now only a definitive answer is cached (`NameNotFoundException` → absent, or a successful lookup); anything else is re-asked on the next read. `by lazy` gave way to a `@Volatile` field since the caching is now conditional — volatile for publication only, as a racing double-probe costs one redundant binder call and writes the same answer.

**2. Superseded permission read could publish late** — confirmed. The result callback and `ON_RESUME` fire within milliseconds of each other when the user answers the dialog, so two reads really are in flight. An older `Denied` landing after the newer `Granted` puts the banner back over a permission the user just granted, and it stays until the next resume. The previous job is now cancelled before a new one starts, and `ensureActive()` sits between the binder call and the state write.

Regression test added: two refreshes queued before the dispatcher runs either, asserting the superseded read never reaches the package manager. The narrower interleaving `ensureActive()` covers — older read already holding its answer, descheduled just short of the state write — needs two real threads to stage and cannot be provoked on a single-threaded test dispatcher, so it is documented in the test rather than faked into a green assertion.

444 tests, 0 failures. Lint clean.

PR #330, opened 2026-08-03
fix(suspend): stop reporting success when an unsuspend removed nothing

trinadhthatakula:
## The bug

A user switched from Shizuku to root and found **every app they had suspended under Shizuku stuck suspended**. Thor reported each unsuspend as successful. Nothing in the logs said otherwise.

## Why

Android records **who** suspended a package, and from **API 30** only that identity can lift it — [`PackageSettingBase.removeSuspension(callingPackage)`](https://cs.android.com/android/platform/superproject/+/android-11.0.0_r1:frameworks/base/services/core/java/com/android/server/pm/PackageSettingBase.java;l=443-452) removes the caller's own entry and nothing else.

Thor's suspender identity is **not stable across privilege modes**:

| Privilege | Recorded as |
|---|---|
| Root | `com.valhalla.thor` |
| Shizuku / Dhizuku | `com.android.shell` |
| Device owner (DPM) | `android` |
| Pre-GH#239 builds, API < 28 root path | `root` |

So a suspension applied under one privilege is simply unliftable under another.

**And the failure is silent by construction.** Unsuspending under a non-owning identity finds no entry to remove, so `oldSuspendParams == newSuspendParams == null` → `changed` stays `false` → PackageManagerService logs *"No change is needed"* → the package is **never added to `unmodifiablePackages`** → the API returns an **empty failure array**. Every caller reads that as total success. Thor did too.

### `parseSuspendingPackages()` — `domain/model/SuspenderReadback.kt`

Reads the suspender set out of `dumpsys package <pkg>`. Four line shapes, because this changed three times:

| API | Shape |
|---|---|
| 28 | inline, `suspendingPackage=<pkg> dialogMessage=` |
| 29 | inline, with ` dialogInfo=` |
| 30–34 | `Suspend params:` block, `suspendingPackage=` entries |
| 35–37 | same block, but the key is `UserPackage.toString()` = `"<" + userId + ">" + packageName` |

That last one is the trap: **a regex written for 30–34 silently captures `<0>com.android.shell`** — a name that matches nothing. Asserted three ways in the tests (equals the bare name, does *not* contain the prefixed form, nothing starts with `<`).

Two more hardening decisions:
- **Free text is cut before any key is read.** `dialogMessage=` / `dialogInfo=` print attacker-influenced text *after* the name, so a crafted message containing `suspended=false suspendingPackage=com.evil.app` would otherwise be parsed. Tested.
- **Indentation-insensitive.** `dumpPackageLPr` takes its prefix from the call site — the `Hidden system packages:` section dumps deeper — so columns are not a reliable anchor. Content is.

### `canLiftSuspension()` — encode the version split, don't assume the modern rule

Below API 30, `setSuspended(false)` blows away the single slot **regardless of who set it** ([android-9.0.0_r1 :399-407](https://cs.android.com/android/platform/superproject/+/android-9.0.0_r1:frameworks/base/services/core/java/com/android/server/pm/PackageSettingBase.java;l=399-407)) — any identity can lift any suspension. From API 30, only the recorder can. Getting this wrong in either direction produces a wrong refusal or a wrong success.

### Unsuspend now reads, acts, and reads back

Read the record → issue one removal **per recorded owner, under that owner's name** → read back to confirm.

Root can name another package because [`enforceCanSetPackagesSuspendedAsUser`](https://cs.android.com/android/platform/superproject/+/android-17.0.0_r1:frameworks/base/services/core/java/com/android/server/pm/PackageManagerService.java;l=3354-3358) **unconditionally early-returns for `Process.ROOT_UID` before any suspender-name validation**. Shell cannot — `allowedShell = callingUid == SHELL_UID && isCallerSameApp(...)`.

Where the platform genuinely will not allow it — a shell-uid Shizuku facing a root-recorded suspension on API 30+ — **Thor now says so and names the owner** rather than returning success.

### An unreadable dump is never "nothing to do"

`readSuspenders()` returns `null` unless the output actually contained a `Package [<pkg>]` block, and `null` fails closed everywhere it is consumed.

This is load-bearing: `dumpsys` requires `android.permission.DUMP`, so a permission-denied read parses as **the empty set** — which is the exact same silent-success bug one level down. `parseSuspendingPackages` documents in bold that empty means *unknown*, never *not suspended*.

## AIDL compatibility

`setAppSuspendedAs` and `dumpPackage` are **appended, not inserted**. Transaction codes are derived from declaration order (`FIRST_CALL_TRANSACTION + index`) and a root daemon **outlives an app update** — inserting would make a stale daemon dispatch `clearAppData()` to whatever moved into that index. Verified with the `aidl` binary that `setAppSuspended` stays +0 and `clearAppData` stays +1.

A stale daemon asked for a method it doesn't have reads back an **empty reply parcel** — `null` for a String — which `readSuspenders()` already treats as unknown and fails closed on. (Note: the generated proxy ignores `transact()`'s return value, so this degrades honestly rather than throwing.)

## Incidental fix

The `SuspendDialogInfo` builder **has never worked**: the `setTitle(CharSequence)` / `setMessage(CharSequence)` overloads don't exist (only the `@StringRes int` ones), and `setNeutralButtonAction(2)` is out of range — only `0` and `1` are defined. The *"managed by Thor"* line users actually see is **system-generated from the suspending package name**, which is why it kept appearing regardless. Now the builder is correct too.

## Verification

- `:app:compileFossDebugKotlin :app:compileFossDebugUnitTestKotlin` → **BUILD SUCCESSFUL**, no new warnings
- `:app:testFossDebugUnitTest --rerun-tasks` → **446 tests, 0 failures, 0 errors, 0 skipped** (counts parsed from JUnit XML, not inferred from `BUILD SUCCESSFUL`)
- `:app:lintFossDebug` → **BUILD SUCCESSFUL** (all five locales carry the new string)

**17 new tests**: all four dump shapes, the `<0>` prefix trap (three assertions), multi-suspender dumps, work-profile sectioning in both directions, `<10>`-inside-`User 0:` filtering, `suspended=false` with a stale block, free-text forgery, deeper indentation, unattributable blocks, empty/whitespace/permission-denied/truncated/garbage input, and `canLiftSuspension` across the API 29→30 boundary from both sides.

## Not verifiable here

**No device testing has been done.** Everything above rests on AOSP source reading and unit tests over captured dump shapes. The parser is pinned against fixtures written from the documented formats, not against output from a real device — worth a check on hardware before release, particularly the API 35+ `<0>`-keyed shape and the cross-privilege refusal path.

### DIFF
--- a/app/src/main/aidl/com/valhalla/thor/rootservice/IThorRootService.aidl
+++ b/app/src/main/aidl/com/valhalla/thor/rootservice/IThorRootService.aidl
@@ -1,6 +1,62 @@
 package com.valhalla.thor.rootservice;
 
+/**
+ * The privileged surface the :root daemon exposes to the app process.
+ *
+ * <p><b>Only ever append to this interface, and never change an existing method's arity.</b> AIDL
+ * derives each transaction code from declaration order (FIRST_CALL_TRANSACTION + index), and a root
+ * daemon can outlive the app that started it -- it is a separate app_process, so an app update
+ * leaves the old one running and bound. Inserting a method renumbers everything below it, and the
+ * stale daemon keeps the old numbering: a clearAppData() call would be dispatched to whatever now
+ * sits at that index, wiping the wrong thing. Appending is safe because an unknown transaction code
+ * falls through to Binder's default onTransact, which returns false, which the generated proxy reads
+ * back as an empty reply parcel -- false for a boolean, null for a String. A stale daemon therefore
+ * degrades into an honest failure rather than a mis-dispatch.
+ *
+ * <p>That is why setAppSuspended below keeps its two-argument shape even though setAppSuspendedAs
+ * supersedes it.
+ */
 interface IThorRootService {
     boolean setAppSuspended(String packageName, boolean suspended);
     boolean clearAppData(String packageName);
+
+    /**
+     * Sets {@code packageName}'s suspended state while recording {@code suspendingPackage} as the
+     * suspending identity, and reports whether the platform's own record agrees afterwards.
+     *
+     * <p>Android keys a suspension on the suspending package name captured at suspend time, and from
+     * API 30 a caller may only lift its own entry (PackageSettingBase.removeSuspension(callingPackage),
+     * android-11.0.0_r1 PackageSettingBase.java:443-452). Naming a suspender you do not own is not an
+     * error either: it leaves oldSuspendParams == null == newSuspendParams, so changed == false, so
+     * the package is left *out* of the returned failure array and the call looks like a success. The
+     * identity therefore has to be passed in from a readback rather than guessed, which is what this
+     * overload exists for.
+     *
+     * <p>Root, and only root, may name an arbitrary identity:
+     * PackageManagerService.enforceCanSetPackagesSuspendedAsUser unconditionally early-returns for
+     * Process.ROOT_UID before any suspender-name validation (android-17.0.0_r1
+     * PackageManagerService.java:3354-3358), unchanged from API 28 to main. Shell is not exempt, so
+     * this rescue path exists in the root daemon and nowhere else.
+     *
+     * @param suspendingPackage the identity to act as. When null the daemon falls back to its own
+     *   historical behaviour: for a suspend, try com.valhalla.thor then com.android.shell then
+     *   android; for an unsuspend, clear every identity the readback reports.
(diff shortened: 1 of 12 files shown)


trinadhthatakula:
@coderabbitai review

trinadhthatakula:
@coderabbitai review

PR #333, opened 2026-08-03
fix(privilege): make Dhizuku usable on device — user id, re-init, Refresh

trinadhthatakula:
Three defects found while device-testing #332's freeze rungs on an Android 17 emulator with Dhizuku as device owner. All three are Dhizuku-facing; one of them (Refresh) also affects root.

## 1 — `am get-current-user` fails under Dhizuku's identity

`DhizukuAPI.newProcess` spawns `pm`/`am` **inside the device-owner app**, not at shell uid, and that identity holds no `INTERACT_ACROSS_USERS`. So `am get-current-user` returned `SecurityException … exit 255` and every `--user`-scoped command died before it ran — including the `pm uninstall -k` fallback #332 had just added.

`removeForUser` and `reinstallApp` now read `Process.myUserHandle().hashCode()` in process (`thorUserId`): no shell, no permission, no binder.

The Shizuku helper moved to the same symbol. Nothing was broken there — the call is permitted at shell uid — but it answered a *different question* from the one the disable rungs ask. Those target `Packages.myUserId`, so on a work-profile device (Thor in profile 10, parent 0 in the foreground) rung 1 disabled for 10 while the fallback removed for 0: a fallback acting on a package the rung it covers for never touched. `RootSystemGateway` deliberately keeps its own shell-based lookup — root runs at uid 0, where the question and the permission both hold.

`reinstallApp` under Dhizuku also swallowed non-zero exits silently; it now logs what `pm` said.

## 2 — the Privilege Check dialog never named Dhizuku

`privilege_check_desc` said "Root or Shizuku" in all five locales, while `privilege_required_warning` right beside it already listed three. Each locale follows its own existing phrasing for the three-way list.

## 3 — Refresh did not refresh privileges

The dialog says *"grant access in your manager app and click Refresh"*. The button called `viewModel.loadDashboardData()`, which reloads the app list and nothing else — `PrivilegeManager.refresh()` had **no call site** outside its own Shizuku listeners.

Shizuku recovered anyway, which is why this went unnoticed: `PrivilegeManager` owns Shizuku's binder and permission listeners and refreshes itself from them. **Root and Dhizuku publish no such callback**, so a grant made while Thor was running stayed invisible until the process was killed and relaunched.

And once the probe *did* re-run, `DhizukuHelper` still latched the `Dhizuku.init` result from `ThorApplication.onCreate` — which on a first run is *before* the user has authorised anything — so every later probe asked an unbound client whether it had permission and was told `false`. Dhizuku 2.6.0 publishes no connection callback to register, so the retry has to be *pulled* from the probe: `probeDhizuku` re-binds whenever the client is not already bound, keeps a successful bind, and re-reads the grant on every probe so a revocation is visible on the very next one.

## Measured

**emulator-5558 — Android 17, Dhizuku device owner, `active=DHIZUKU`**

Freeze `com.android.egg`:

| | message |
|---|---|
| before | *"Android reported: Failed to determine cur…"* — rung 2 threw at `am get-current-user` |
| after | rung 1 refused, rung 2 issued `pm uninstall -k --user 0` and got `Failure [only root can delete system app for a particular user]` → the Root-mode message |

```
E/DhizukuHelper: setAppDisabled(com.android.egg, disabled=true): all rungs ran
  (pm shell, IPackageManager reflection, unprivileged PackageManager) and the
  state did not change — the platform REFUSED (SecurityException)
W/DhizukuSystemGateway: freeze(com.android.egg): this device refuses to let
  Dhizuku's device-owner identity disable system packages; falling back to
  `pm uninstall -k --user N`, which keeps the app's data
W/DhizukuHelper: `pm uninstall -k --user 0 com.android.egg` exited 1:
  Failure [only root can delete system app for a particular user]
```

The `--user 0` in that last line is `thorUserId` resolving in process on real hardware. This is also the hardware answer #332 left open — *"the same on Android 17 where rung 2 should produce the Root-mode message"*: yes.

`com.android.egg` unchanged throughout — `installed=true enabled=0 ceDataInode=549088`, before and after.

Refresh, on the same process (no force-stop, pid 7453 both times):

```
17:24:17  probe total=90ms root=87ms/false shizuku=4ms/false dhizuku=8ms/false
17:24:17  ready sinceProcessStart=564ms active=NONE      <- launched with access revoked
          … access granted in Dhizuku while Thor stays in the background …
17:25:45  probe total=34ms root=3ms/false shizuku=0ms/false dhizuku=32ms/true
```

Chip red → green. The 32 ms is the re-bind doing its job; steady-state was 1–8 ms. Returning to the app did *not* re-probe on its own — the second line appears only after Refresh, which is what makes it evidence.

**emulator-5556 — Android 17, `active=SHIZUKU`** — Fix Store on a user app ran `pm install -r -d -i "com.android.vending" --user $thorUserId …` and moved `installerPackageName` `null` → `com.android.vending`, `versionCode` unchanged. Restored to `null` afterwards.

## Tests

7 new tests pin `probeDhizuku`'s decision table. There is no Robolectric/mockk here and `unitTests.returnDefaultValues` is off, so the two `DhizukuAPI` statics are passed in as lambdas — that seam is the only reason any of this is reachable from a plain JVM test. They cover the two properties the fix rests on: a failed bind stays retryable, and a successful bind is not repeated.

- `:app:testFossDebugUnitTest --rerun-tasks` — **49 classes / 497 tests / 0 failures** (dev baseline 48 / 490)
- `:app:lintFossDebug` — 5 issues, all pre-existing `VectorPath` hints, 0 errors, 0 warnings

Follow-up to #332 (GH#316). No issue closed by this PR.

## Summary by CodeRabbit

- **New Features**
  - Added Dhizuku support for privilege detection and package management.
  - Privilege checks can now be refreshed directly from the home screen.
  - Improved app operations across the active user profile.

- **Bug Fixes**
  - Dhizuku initialization now retries after failures and preserves successful connections.
  - Improved error reporting for uninstall and reinstall operations.
  - System-app freezing continues to preserve app data.

- **Documentation**
  - Updated Arabic, Chinese, English, French, and Spanish privilege messages to mention Dhizuku.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/ThorApplication.kt
+++ b/app/src/main/java/com/valhalla/thor/ThorApplication.kt
@@ -12,6 +12,7 @@ import com.rosan.dhizuku.api.Dhizuku
 import com.valhalla.bypass.Bypass
 import com.valhalla.thor.core.ThorShellConfig
 import com.valhalla.thor.data.service.AutoFreezeManager
+import com.valhalla.thor.data.source.local.dhizuku.DhizukuHelper
 import com.valhalla.thor.domain.repository.PreferenceRepository
 import com.valhalla.thor.presentation.settings.BillingProcessor
 import com.valhalla.thor.presentation.utils.AppIconFetcher
@@ -86,10 +87,16 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
         Bypass.prepareThor()
         ThorShellConfig.init()
 
+        // Hand the outcome to DhizukuHelper rather than dropping it. This init runs once, at process
+        // start — which for a first-run user is *before* they have authorised Thor in Dhizuku, so it
+        // returns false and there is nothing here to retry it. Recording the answer is what lets the
+        // later probe tell "already bound, just ask about permission" apart from "never bound, bind
+        // now" instead of asking an unbound client whether it has permission and always hearing no.
         try {
-            Dhizuku.init(this)
+            DhizukuHelper.markClientInitialised(Dhizuku.init(this))
         } catch (e: Exception) {
             Logger.e("ThorApp", "Dhizuku init failed", e)
+            DhizukuHelper.markClientInitialised(false)
         }
 
         autoFreezeManager.startObserving()
--- a/app/src/main/java/com/valhalla/thor/data/gateway/AndroidUserIds.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/AndroidUserIds.kt
@@ -16,5 +16,9 @@ private const val PER_USER_RANGE = 100_000
  * Shared by all three gateways rather than repeated in each: they must agree on which user a package
  * lives in, or the same `pm grant` would land on a different user depending on which privilege mode
  * happened to be active. One copy is also one place to correct if the platform ever moves the range.
+ *
+ * This answers *which user does this uid belong to*. For *which user is Thor itself*, see
+ * `thorUserId` in `data/source/local/ThorUser.kt` — that one is also read by the two privilege
+ * helpers, so it sits in the package they share rather than here.
  */
 internal fun userIdOf(uid: Int): Int = uid / PER_USER_RANGE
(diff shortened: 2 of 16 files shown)


trinadhthatakula:
Taken — `thorUserId` was in `shizuku/` only because that was the first call site.

Moved to `data/source/local/ThorUser.kt` rather than next to `userIdOf` in `data/gateway/AndroidUserIds.kt`: `data/gateway` imports from `data/source/local` today and nothing imports back, so putting a value the two helpers read into `gateway` would have inverted that. The common parent package keeps every edge pointing the same way. Both files now cross-reference each other — `userIdOf` maps some *other* uid to its user, `thorUserId` is Thor's own.

The move also turned up two copies of the definition already inlined in `Shizuku.kt` (`clearCache`, `clearAppData`, both spelling out `android.os.Process.myUserHandle().hashCode()`); folded those in too. Same expression, no behaviour change.

`0bbf9e96`. 497 tests / 0 failures, lint unchanged.

PR #334, opened 2026-08-03
chore(release): bump to 1932 and write the v1.93.2 release notes

trinadhthatakula:
## What

Bumps `versionCode` 1931 → 1932 (**v1.93.2**) and writes the release notes for all five channels.
Another pre-release dev build; the current **stable** is still v1.93.0.

## Range

Baselined on the tag **`v1.93.1-dev-103`**, not on `master` — `master` runs ahead of its own release
tag, so `master..dev` undercounts. **100 commits, 28 merged pull requests, #304–#333.** Roughly two
thirds of that is the new `web/` site, documentation and CI, so the user-visible surface is much
smaller than the commit count suggests.

## Five channels, three files

`playstore.txt` is the single source for Play, F-Droid and the Shizu Store, so it is written once
and propagated:

```bash
cp release-notes/v1.93.2/playstore.txt fastlane/metadata/android/en-US/changelogs/1932.txt
.github/scripts/sync-shizu-changelog.sh
```

| Channel | File | Status |
|---|---|---|
| Google Play "What's new" | `release-notes/v1.93.2/playstore.txt` | 482 chars / 486 UTF-16 (limit 500) |
| F-Droid changelog | `fastlane/…/changelogs/1932.txt` | byte-identical ✅ |
| Shizu CoreFetch | `shizu_store.json` → `.changelog` | matches (manifest holds it without the trailing newline, by design) ✅ |
| Telegram broadcast | `release-notes/v1.93.2/telegram.md` | 826 UTF-16 units (budget ~870) |
| GitHub Release body | `release-notes/v1.93.2/github.md` | present |

The *writer* is the manual half: no workflow runs `sync-shizu-changelog.sh`, because the `master`
ruleset needs a PR plus a status check and a `GITHUB_TOKEN`-authored PR does not trigger
`pull_request` workflows — so no bot can land that commit. Without this one, Shizu CoreFetch would
have gone on advertising v1.93.1's feature list ("New: Freeze Profiles", "export as .xapk") under
v1.93.2.

The *checker* is the automated half: `pr-ci.yml:82` runs `check-shizu-manifest.sh` on every PR,
deliberately un-path-filtered, so forgetting the sync is a red `shizu-manifest` check rather than a
silent store regression. The green `shizu-manifest` check on this PR **is** that gate agreeing the
three copies match — including *"changelog matches release-notes/v1.93.2/playstore.txt (versionCode
1932 -> v1.93.2)"*.

`github.md` exists for the same class of reason: `dev-check.yml:166` falls back to a raw
`git log --oneline` dump when it is missing, which for this range would have published 100
unfiltered commit subjects — mostly docs and Dependabot bumps — as the release body, while CI went
green.

## Verification

```
playstore.txt   482 chars / 486 UTF-16 units      (Play rejects >= 500)
telegram.md     826 UTF-16 units                  (budget ~870)
  → dev-check.yml final caption      974 units    (cap 1024, 50 spare)
  → telegram-release.yml caption     966 units    (cap 1024, 58 spare)
closing keywords in release-notes/v1.93.2/*       clean
blank line between every bullet                   both files
diff playstore.txt vs fastlane/…/1932.txt         identical
check-shizu-manifest.sh                           all checks passed
every commit hash cited in github.md              50/50 resolve, subjects match
@Test methods under app/src/test                  497 (matches the notes' claim)
```

The telegram budget is the one that fails **silently**: the notes go out as a `sendDocument`
*caption*, Telegram caps captions at 1024 UTF-16 units and **rejects** an oversized one outright
(no truncation), and the `curl` has no `--fail` — so the step goes green having broadcast nothing.
The file is only one of the caption's inputs, so the numbers above are the **assembled** captions,
with the wrapper measured at 149 units (dev) and 141 (release) rather than estimated.

## Claims were checked against the code, not the commit subjects

Corrections made while verifying the first draft:

* **The data-safety headline was mis-scoped.** "Thor disables it instead" is *false* on exactly the
  devices that motivated the change — HyperOS's vendor `PackageManagerServiceImpl` refuses the
  disable outright, and the fix for those users is that the fallback now carries `-k`. Both consumer
  files now say the data survives either way.
* **Dropped the OPPO and Honor brand names.** Only Xiaomi/HyperOS was ever observed; the rest of the
  ROM list is inference from T/TAF 108-2022. A defect claim about named third-party hardware,
  published verbatim to three stores with no surrounding context, needs evidence behind it.
* **Dropped "Root can rescue apps suspended under Shizuku" from the broadcast.** It is implemented
  and unit-tested but has **not** run on a device — `docs/follow-ups/cross-privilege-suspend-ownership.md`
  says so explicitly. The half a readback does guarantee is kept.
* **Softened the permission-banner claim.** The banner is gated on `InstalledAppsPermission.Denied`,
  and a "while in use" grant reads as `Granted` — i.e. the users most likely to hit the bug will not
  see it.
* **Added the behaviour change nothing mentioned:** a system-app freeze can now *fail* where it used
  to appear to succeed, because root no longer escalates to an uninstall and the fallback does not
  exist at shell uid on API 37. That is the likeliest source of "1932 broke system-app freezing"
  reports.
* Vocabulary: "unpause" → "unsuspend" (the app's own string is `action_unsuspend`), 🔥 → ❄️ on a thaw
  bullet, and dropped the self-deprecating "really".

Every commit hash cited in `github.md` was verified to exist in the range; the test counts
(376 → 497) and the PR count (28, not the 32 raw merges — four are branch syncs) were measured.

## Note

No GitHub closing keywords in any of the three notes files — `master` is the default branch, so a
`dev` → `master` PR body triggers auto-close. `release-notes/README.md` is rewritten in this commit
to document the pipeline, the four traps and the propagation/verification steps.

Part of GH#316 and GH#310, which their own PRs already handled.

### DIFF
--- a/docs/follow-ups/telegram-caption-length-guard.md
+++ b/docs/follow-ups/telegram-caption-length-guard.md
@@ -1,0 +1,75 @@
+# Follow-up: assert the Telegram caption length before `sendDocument`
+
+**Status:** open. Raised in review of PR #334 (the v1.93.2 release notes). Deliberately **not**
+fixed there — see "Why not in #334".
+
+## The failure it prevents
+
+Release notes reach Telegram as a `sendDocument` **caption**, not as a message
+(`.github/workflows/telegram-release.yml:146`, `.github/workflows/dev-check.yml:249`). Telegram caps
+captions at **1024 UTF-16 code units** and **rejects** an oversized one outright — it does not
+truncate. The `curl` carries no `--fail` and its output goes to `/dev/null`, so the step exits 0
+having broadcast nothing. Every downstream signal — the job, the check, the release — stays green.
+
+`release-notes/README.md` documents a ~870-unit budget for `telegram.md`, but that is a *manual*
+gate applied to only one input. The value actually sent is `CAPTION`, which is `telegram.md` plus a
+header the workflow builds and, on the release path, a GitHub link footer.
+
+## Measured, for v1.93.2
+
+| | UTF-16 units | headroom to 1024 |
+|---|---|---|
+| `telegram.md` alone | 825 | — |
+| `dev-check.yml` final caption | 974 | 50 |
+| `telegram-release.yml` final caption | 966 | 58 |
+
+Wrapper cost: **149** units on the dev path, **141** on the release path. It is not a constant — the
+dev header interpolates `github.ref_name`, `github.actor` and a track label, so a longer branch name
+or actor pushes it up with nothing to notice.
+
+v1.93.2 ships with 50 units of headroom, and only after two bullets were trimmed to buy it back: a
+one-line accuracy fix in review took the dev caption to 1006 units — 18 from silence — and nothing
+but a hand-run `python3` one-liner said so. v1.93.0's notes were 1008 units *before* any wrapper,
+i.e. already over.
+
+## What to add
+
+After `CAPTION` is fully assembled and before the `curl`, in **both** workflows:
+
+```bash
+CAPTION_UNITS=$(printf '%s' "$CAPTION" | python3 -c \
+  'import sys; t=sys.stdin.read(); print(sum(2 if ord(c) > 0xFFFF else 1 for c in t))')
+if [ "$CAPTION_UNITS" -ge 1024 ]; then
+  echo "::error::Telegram caption is $CAPTION_UNITS UTF-16 units; the cap is 1024. \
+Trim release-notes/v\$VERSION_NAME/telegram.md."
+  exit 1
+fi
+```
+
+## Why not in #334
+
+Two reasons, and the second is the design question that makes this its own change.
+
+1. #334 is a release PR that ships a 988-unit caption on the path the guard would run on. A guard
+   with an off-by-one in its own arithmetic would redden the release it is meant to protect, and the
+   only way to find out is to publish.
+
+2. **Placement is the actual decision, and a naive fix gets it wrong.** In `dev-check.yml` the
+   Telegram step runs *after* fastlane has uploaded to Play and *before* "Create GitHub
+   Pre-Release". A guard there fails loudly — but it fails a release that is already half-published:
+   on Play, absent from GitHub, silent on Telegram. That is arguably worse than the silent skip it
+   replaces.
+
+   The gate belongs **before the build**, as a pre-flight step that reads `versionCode`, derives the
+   version name, reconstructs the caption the later step would send, and fails while nothing has
+   shipped. That step does not exist yet, and inventing it inside a release PR is how a release PR
+   stops being one.
+
+A pre-flight step is also the natural home for the other unchecked size — `playstore.txt` must stay
+under 500 characters and nothing in CI verifies that either. Both belong in the same script.
+
+## See also
+
+* `release-notes/README.md` — trap 1, and Step 6, which is the manual gate this would replace.
+* `.github/scripts/check-shizu-manifest.sh` — the precedent: a script whose every assertion exists
+  to make one silent failure loud, wired into `pr-ci.yml` on every PR.
--- a/fastlane/metadata/android/en-US/changelogs/1932.txt
+++ b/fastlane/metadata/android/en-US/changelogs/1932.txt
@@ -1,0 +1,11 @@
+• 🧊 Freezing a preinstalled app no longer wipes its data — Thor disables it, and keeps the data if it must remove it.
+
+• ❄️ Removing an app from the Freezer unfreezes it first, on every screen — and says so if it can't.
+
+• 📱 Fixed the app list collapsing to one entry on Chinese-market ROMs.
+
+• ⏸️ Thor no longer claims success when it couldn't unsuspend an app.
+
+• 🔄 Privilege Check's Refresh re-checks Root, Shizuku and Dhizuku.
+
+• 💖 New: more ways to support Thor's development.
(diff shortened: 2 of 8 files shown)


trinadhthatakula:
## Review pass — 4 fixes in `626031d0`, 1 declined with a follow-up

I reviewed this PR myself as well, and the first finding is one of mine.

### 1. My own commit message was wrong about CI

I wrote that `check-shizu-manifest.sh` "is wired into no workflow". It is —
[`pr-ci.yml:82`](.github/workflows/pr-ci.yml#L82) runs it on **every** PR, deliberately
un-path-filtered because `on.pull_request.paths` would gate the whole workflow including the
required `build-and-test` check. The green `shizu-manifest` check on this PR *is* that gate agreeing
the three copies match.

What no workflow runs is **`sync-shizu-changelog.sh`**, the *writer*. Conflating the writer with the
checker turned a documented division of labour into "nothing verifies the manifest". Commit message
amended, PR body rewritten, and `release-notes/README.md` now states both halves.

### 2. Found while checking that: the `fastlane/` copy is audited by nothing

`check-shizu-manifest.sh` compares `shizu_store.json`'s `.changelog` against `playstore.txt` and
stops there. `fastlane/metadata/android/en-US/changelogs/<versionCode>.txt` — the copy **F-Droid
actually ships** — is checked by no script and no workflow. The README claimed the audit "compares
them", plural, which reads as all three. Corrected, and Step 6's `diff` is now labelled as that
copy's only gate.

### 3. `telegram.md` freeze bullet — taking the finding, not the wording

You're right that the line promised more than the code delivers. "If your phone refuses, it removes
the app but keeps the data" reads as *removal always follows*, and it does not:
`uninstallFreezeFallbackAllowed` returns `false` for `PrivilegeMode.ROOT`
([`FreezePolicy.kt:234`](app/src/main/java/com/valhalla/thor/domain/model/FreezePolicy.kt#L234)),
and the shell fallback is refused outright on API 37 — both leave the package untouched and report
failure. Now reads *"where your phone refuses that, it **can** remove the app and keep the data
instead."*

Not taking the suggested diff, which puts `` `-k` `` in front of a Telegram audience.
`release-notes/README.md` bars internal vocabulary from the consumer channels, and `-k` is a
`pm uninstall` flag.

### 4. …which nearly caused the exact failure this PR documents

That 19-unit fix took the **assembled** dev-check caption to **1006** of Telegram's 1024 — 18 units
from posting nothing, silently, with the job still green. Trimmed two bullets to buy it back:

| | UTF-16 units | headroom |
|---|---|---|
| `telegram.md` | 826 | (budget ~870) |
| `dev-check.yml` caption | 974 | **50** |
| `telegram-release.yml` caption | 966 | **58** |

The wrapper is now measured — **149** units on the dev path, **141** on the release path — instead of
the ~140 the README estimated, and it is not a constant: it interpolates `github.ref_name`,
`github.actor` and the track label.

### 5. Declined here: the caption-length assertion — `docs/follow-ups/telegram-caption-length-guard.md`

You're right that one should exist. Not in this PR, for two reasons, and the second is why it needs
its own:

1. This PR ships a 974-unit caption on the very path the guard runs on. A guard with an off-by-one
   in its own arithmetic reddens the release it exists to protect, and the only way to find out is
   to publish.

2. **Placement is the real decision.** In `dev-check.yml` the Telegram step runs *after* fastlane has
   uploaded to Play and *before* "Create GitHub Pre-Release". A guard bolted on there fails loudly —
   but it fails a release that is already half-published: live on Play, absent from GitHub, silent on
   Telegram. That is arguably worse than the silent skip it replaces. The gate belongs in a
   **pre-flight** step that reconstructs the caption before anything ships — and that step is also
   the natural home for the `playstore.txt` 500-char limit, which CI does not check either.

Written up with the measurements, the exact snippet and that reasoning.

---

### Also verified this pass

* All **50** commit hashes cited in `github.md` resolve, and their subjects match what the notes say
  they did.
* `497` `@Test` methods under `app/src/test` — the count the notes claim.
* `uninstallFreezeFallbackAllowed`, `thorUserId`, `SUSPECT_SCAN_TOLERANCE`,
  `disabledAppsNotInFreezer`, the `GET_INSTALLED_APPS` manifest entry, and
  `freeze_system_app_requires_root` / `removed_from_freezer_partial_failure` in all five locales all
  exist as described; `cd_uninstalled` is gone from all five, as claimed.
* No GitHub closing keywords in any notes file — `master` is the default branch, so a `dev` → `master`
  body auto-closes.

trinadhthatakula:
### Third finding — taken, in `bfb124f0`

Right, and it is the kind of overstatement this particular file cannot afford. `github.md` is the open-source accountability record, and the whole freeze rework rests on *"we can show the data survives"*.

What `cf739fc0` measured is that `ceDataInode` and `deDataInode` were **unchanged** across `pm uninstall -k` + `pm install-existing` — so the CE and DE data directories were never recreated, which is exactly what `DELETE_KEEP_DATA` promises. That is sufficient evidence for the claim being made. It is not a byte comparison of file contents, and calling two integers "byte-identical" invited exactly the reading you gave it. Reworded to say what was measured and, explicitly, what was not.

**Note for a separate change:** the same phrasing is in the three KDocs it was inherited from — `Shizuku.kt:1044`, `Dhizuku.kt:411`, `FreezePolicy.kt:82`. Deliberately not touched here; a `chore(release)` branch is the wrong place to edit app source, even comments only. Worth its own small PR — `cf739fc0`'s own message calls stale prose on a changed path a defect, and this is that.

Issue #310, opened 2026-07-31
Freezer: bulk removal swallows restore failures and always reports success


trinadhthatakula:
Surfaced during review of #308, which fixed the same class of bug on the two single-app removal paths (`AppListViewModel.toggleFreezerMembership`, `AppInfoDetailsViewModel.addOrRemoveFromFreezer`). The bulk path on the Freezer screen was left alone because it was out of scope for that PR — but it is the worse of the three.

`FreezerViewModel.removeFromFreezer` (`app/src/main/java/com/valhalla/thor/presentation/freezer/FreezerViewModel.kt:194`):

```kotlin
packageNames.forEach { pkg ->
    freezerRepository.remove(pkg)
    val app = appByPkg[pkg]
    if (app != null) manageAppUseCase.restoreApp(pkg, app.enabled, app.isSuspended)
    else manageAppUseCase.forceUnfreeze(pkg)
    freezerShortcutManager.disableAppShortcut(pkg)
}
_uiState.update { it.copy(multiSelection = emptySet()) }
emitToast(UiText.PluralsResource(R.plurals.removed_from_freezer_success, packageNames.size))
```

Two problems, and they compound:

**1. The `Result` is discarded.** `restoreApp` and `forceUnfreeze` both return a `Result`; neither return value is bound or inspected. A shell failure — no privilege, gateway fallback exhausted, `pm` returning non-zero — is silently swallowed.

**2. The toast reports `packageNames.size` unconditionally.** Select twelve apps, have every restore fail, and the user is told "Removed 12 apps from the Freezer" while twelve apps sit frozen.

**3. Ordering, same as #308.** `freezerRepository.remove(pkg)` is a durable Room delete and runs *before* the fallible call. So a failed restore leaves the app frozen with no watchlist entry — the Freezer screen no longer lists it, Unfreeze-all can't reach it, and the only route back is the import-already-disabled flow.

### Suggested fix

Mirror what #308 landed on the single-app paths, adapted for a bulk operation:

- restore first, drop the row + grey out the shortcut only on success;
- accumulate failures per package rather than aborting the loop — one app failing shouldn't strand the other eleven;
[...]

PR #336, opened 2026-08-03
docs(follow-ups): resweep the index, which had drifted from its own docs

trinadhthatakula:
Docs only. No source, no build files, no workflows.

`docs/follow-ups/README.md` is the file you read before picking up work. It had drifted from the docs
it links, which is the one failure mode it exists to prevent — so this is a resweep, not a tidy-up.

## The drift, worst first

**#22 contradicted its own doc.** Two rows still read *"2 of 8 configurations, and root-granted comes
back **bad**"*. `privilege-manager-cold-start.md` line 3 has said `✅ CLOSED as of 2026-08-01. All 8
configurations are measured` since PR #306 landed.

That gap matters more than a date being wrong. The interim verdict was **refuted, not superseded**:

* the bimodal slow `root=` mode is **CPU contention, not an emulator artifact** — reproducible on
  demand by saturating a physical device's cores (idle 45–60 ms unimodal → loaded 71–91 / 157–205 ms),
* the probe-vs-first-frame race is **self-balancing**, because contention slows the first frame too —
  worst case **−1 ms** at 2× oversubscription,
* so the projected *"~579 ms of visible spinner on ~60% of rooted cold starts"* never existed.

Anyone reading the index would have concluded Thor has a known startup regression. It does not. Both
rows corrected; the doc stays on disk as the **measurement record**, not as open work.

**Three rows overtaken by their own PRs.**

| Row | Said | Actually |
|---|---|---|
| #55a freeze profiles | "scoped 2026-07-30, unstarted" | PR #295 merged 2026-07-30, shipped v1.93.1 |
| #285 permission filter | "in review; issue still open — close it on merge" | PR #294 merged; issue closed |
| #164 export formats | "can be closed *with its scope stated*" | done — closed 2026-08-03, scope stated |

**One doc was never indexed at all.** `telegram-caption-length-guard.md`, filed 2026-08-03, had no
row. That is precisely how a follow-up stops being tracked — it now has one, carrying the measured
977-of-1024 margin and the reason the guard belongs in a pre-flight step rather than next to the
`curl`.

## Deletions, and one deliberate non-deletion

This file's own rule is *"when one ships, delete the row **and** the doc."* Applied to two:

* `shortcut-match-flags-system-apps.md` — shipped, referenced only by this index.
* `export-share-defects-found-during-30-recon.md` — all three defects resolved (two fixed, one
  decided). **Its surviving decision was folded forward first**: the OBB-field call (*option 2 — the
  field stays for phase 2*) now lives in the #51 phase-2 row, so deleting the file loses nothing.

`viewmodel-behavior-tests.md` shipped too and is **kept**. The open `BulkFreezeRunner` row cites it as
its worked example — it sets the mutation-checking bar that row has to meet — and
`app/build.gradle.kts:317` points at it by path. Deleting it would trade a tidy index for context an
open item depends on. The rule in the header now carries that exception explicitly, so the next sweep
does not "fix" it.

## GitHub reconciled in the same pass

The same drift had left shipped work looking open on the issue tracker.

**Closed 3**, each with the scope of what shipped stated rather than implied:

* **#310** — bulk removal swallowing restore failures. Fixed by #331, shipped v1.93.2.
* **#164** — export formats. 3 of 4 shipped; the raw split *folder* is declined, and the closing
  comment says so rather than letting the close imply it all landed.
* **#210** — the request split in two. The launcher half shipped (Suspend mode in v1.93.0, plus the
  bridge that unfreezes on tap in v1.93.1); auto-refreeze-on-Recents is declined, no public API.

**Commented, left open 3:** #55 (profiles shipped, process manager deferred), #51 (export half
shipped and rides Nearby Share; the data half is what keeps it open), #161 (posted the diagnosis —
`android:host="*"` makes the 35 `pathPattern` entries mandatory, so adding extensions cannot fix it).

Where a row outlives its now-closed issue, the row says so. The freezer escape-hatch row is what
GH#310 left behind, and **nothing on GitHub tracks it** — that is worth stating in the one place
someone will look.

## Checks

* No dead links: every `(*.md)` target in the index resolves.
* No orphans: every doc on disk is linked from the index.
* Tables well-formed: verified by unescaped-pipe count per row against each header (the two rows that
  look wrong to a naive counter carry escaped `\|` inside cell text).
* Sweep claim re-verified: **zero** `TODO`/`FIXME`/`HACK`/`XXX` markers across `app/src/main/java`
  and `bypass`.
* `ACTION_SEND` claim in the #161 row re-verified against the manifest: zero matches.


trinadhthatakula:
Both findings were right, and both are the same failure this PR set out to fix — the index asserting something its own contents contradict. Fixed in 1e5ab647.

**Dates — right, and worse than "future-dated".** I wrote them off a +05:30 clock. GitHub records the closures at **`#310 2026-08-03T21:59:37Z`, `#164 …:39Z`, `#210 …:40Z`** — so all eight `2026-08-04` claims were a day ahead of the record they exist to be checked against, not just `Last swept`. Anyone verifying a row against the issue tracker would have found it off by one, every time. All eight now read `2026-08-03`, and the header states the convention:

> Dates here are **UTC**, matching the GitHub timestamps they can be checked against. A sweep run late in the evening IST lands on the previous UTC day.

**Retention rule — right, and it undersold the file by two cases.** The rule claimed one exception, but the file was already leaning on three:

1. an open row cites the doc as a worked example — `viewmodel-behavior-tests.md`, for the `BulkFreezeRunner` row;
2. the doc is the evidence under a conclusion — `privilege-manager-cold-start.md`, which the #22 row itself calls *"retained as the measurement record"*;
3. seven closed Tier 0/1 rows whose only remaining job is to stop the work being restarted.

A rule narrower than the practice is an invitation for the next sweep to delete (2) and (3) as cleanup, which would drop the evidence under #22's verdict. All three are written down now, each with the example that motivates it.

Re-verified after the change: zero dead links, zero orphaned docs, no `2026-08-04` left anywhere under `docs/follow-ups/`.

PR #340, opened 2026-08-04
fix(security): close the app lock's cold-start gap, and stop leaking the app list to Recents

trinadhthatakula:
Four defects on the path between `onCreate` and the first frame, all of them variants of one mistake: **"not yet known" encoded as a safe-looking default.**

## 1. The app lock reported itself open before the preference had been read

`_biometricEnabled` seeded `false`, and `!enabled -> NotRequired` was the *first* branch of the `when`. So "the preference has not arrived from DataStore yet" and "the user turned the lock off" were the same value, and DataStore answers a beat *after* `setContent` returns. Every cold start therefore reported `NotRequired` until the disk read landed — long enough to compose the entire app behind a lock that had not finished switching on.

Fixed with a tri-state rather than by reseeding: `StateFlow<Boolean?>` seeded `null`, and an explicit `AuthState.Loading` as the first branch.

Reseeding to `Locked` fails the other way. `BiometricScreen` fires the prompt from a `LaunchedEffect(Unit)` on composition, so every user who never armed the lock would get a fingerprint dialog on every launch. The `stateIn` seed moves to `Loading` for the same reason: `Eagerly` starts the collector but does not make `combine` emit inline, so the seed is what `HomeActivity` actually reads at `setContent`, and leaving it `Locked` would just relocate the bug into the seed.

## 2. Restored navigation state was consumed and dropped

`MainScreen`'s first composition reads the restored state, which Compose's `SaveableStateRegistry` then discards. A user who authenticated a moment later landed on the start destination with their place gone.

Closed by (1) plus holding the splash: `core-splashscreen`'s `setKeepOnScreenCondition` installs an `OnPreDrawListener` that cancels the frame, so during `Loading` nothing composes and nothing calls `consumeRestored()`. The restored map has no timeout, so waiting costs nothing.

## 3. The Recents card showed the app list straight through the lock

Android snapshots whatever was on screen when Thor backgrounds — typically the full installed-app list, including which apps are frozen or hidden — and shows it to anyone who picks the device up. No prompt involved. `FLAG_SECURE` excludes the window from that capture, and from screenshots and screen recorders with it.

Keyed on `authState`, **not** on `prefs.biometricLockEnabled`. `prefs` is seeded with `UserPreferences()` until DataStore answers and its default is `false`, so keying on the preference would leave the window capturable for exactly the window this PR exists to close — the same defect one screen later. The condition is inverted to fail closed: every state is secure *except* `NotRequired`, so an `AuthState` added later is protected by default rather than by someone remembering to amend that line. A user who never armed the lock keeps their screenshots.

## 4. AP-C02 in `BiometricScreen`

Both animated reads moved into `graphicsLayer` lambdas.

## Tests

`SecurityViewModelColdStartTest`, with a preference that answers only after 50 ms of virtual time on a `StandardTestDispatcher` — the condition `SecurityViewModelTest` cannot reach, since a `MutableStateFlow` already has its value. Assertions are on the whole *sequence* of states, because "the gate was never open" is a claim about every emission and sampling `.value` afterwards cannot make it.

Its collector is a foreground `launch`, cancelled once the clock runs out, and that is load-bearing: **work in `backgroundScope` is not guaranteed to have been dispatched when `advanceUntilIdle()` returns**, so a collector there records the initial value and misses every update. The first draft did exactly that and asserted `[Loading]` against an `authState` that had already reached `Locked`. Measured four ways before rewriting.

## Reconciliation with #339

Both branches rewrote the same `combine`, because they fix two halves of one defect: this one closes the window where the preference had not been read *yet*, #339 closes the window where it could not be read *at all*. The merge commit keeps both conditions:

```kotlin
val required = enabled == true || (settingsLost && capable)
when {
    enabled == null -> AuthState.Loading
    !required -> AuthState.NotRequired
    ...
```

`enabled == true` rather than leaning on a smart cast from the `Loading` branch above it, so that reordering the `when` cannot silently turn "not read yet" back into "not armed".

## Verification

517 unit tests, 0 failed, 0 skipped. `lintFossRelease` 0 errors, 0 warnings.

**Not covered:** `HomeActivity` has no test — there is no Robolectric or instrumentation harness in this suite — so the splash guard, the `Loading` branch and the `FLAG_SECURE` toggle are unverified by any automated check. Defect 3 in particular wants a device: arm the lock, open the app list, background it, look at the Recents card.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup security by keeping the splash screen visible while authentication settings load.
  * Prevented the app from briefly displaying unlocked content before biometric-lock preferences are resolved.
  * Applied screen-capture protection whenever authentication is required.
  * Added an opaque themed loading backdrop for a smoother, more secure startup experience.

* **Tests**
  * Added coverage for delayed preference loading and cold-start authentication transitions.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/HomeActivity.kt
+++ b/app/src/main/java/com/valhalla/thor/HomeActivity.kt
@@ -8,12 +8,19 @@ import android.content.Intent
 import android.os.Build
 import android.os.Bundle
 import android.provider.Settings
+import android.view.WindowManager
 import android.widget.Toast
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
+import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.ui.Modifier
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.lifecycleScope
@@ -62,7 +69,14 @@ class HomeActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        installSplashScreen()
+        val splashScreen = installSplashScreen()
+        // Hold the frame until the app lock has answered. The preference comes from DataStore, which
+        // answers a beat after `setContent` returns, so without this the branch the user sees first
+        // is decided by a race rather than by the preference — and the losing outcome is the app,
+        // fully composed, behind a lock that had not finished switching on.
+        splashScreen.setKeepOnScreenCondition {
+            securityViewModel.authState.value == AuthState.Loading
+        }
         enableEdgeToEdge()
         shizukuHandler.register()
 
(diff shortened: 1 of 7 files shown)


trinadhthatakula:
@coderabbitai review

trinadhthatakula:
@coderabbitai full review

PR #342, opened 2026-08-04
fix(billing): let a resume rebuild a connection the ladder gave up on

trinadhthatakula:
Three defects in one lifecycle, all in the `store` flavour, all costing subscription revenue on the devices least able to afford a flaky Play Store.

### The reconnect ladder ended after one step, not five

`BillingClientImpl.zzbu` answers `BILLING_UNAVAILABLE` **synchronously, on the calling thread**, when the Play Store service cannot be resolved — the normal path on a device with a degraded or restricted Play Store, not an edge case. So the failure callback re-entered the ladder from inside the very coroutine servicing the queued retry. The old guard asked *"is that coroutine still alive?"*, which answered **yes to its own caller**, and dropped the step.

The guard is now an explicit flag cleared immediately before `startConnection`, so a synchronous failure is read as what it is.

### Nothing re-armed a spent ladder

`enableAutoServiceReconnection()` is **lazy, not proactive**: `zzaI` is reached only from `zzbw`/`zzbx` at the head of an API callable, so a client whose every call site is `isReady`-guarded never reconnects at all. Past the fifth attempt the support sheet rendered the *"Rate on Play Store"* fallback for the rest of the process's life and no subscription could be sold.

`refreshPurchases` now re-arms on resume — the one input the ladder structurally cannot have: evidence about the outside world, rather than about the last 31 seconds. Bounded by a **30 s floor on the rate**, not by a count. Any fixed budget is a promise that Play becomes reachable within it, and that is not a promise this code can keep.

### And the counting is now testable

`app/src/store` is not on `testFossDebugUnitTest`'s classpath, so the money-losing half of this file could not be asserted at all. `BillingReconnectLadder` and its step type move to `app/src/main`, **taking a clock instead of calling one**; everything naming `BillingClient` stays on the other side of that line. 398 lines of new tests cover the ladder, the re-arm cooldown, and the delay clamp.

## Riding along

- **The one untranslated string in the flavour.** The subscription tier rendered `"$price / month"` from a hardcoded English literal — and a claim about the billing period that nothing checked. Play's own `billingPeriod` now picks a translated template, and an unrecognised period renders the price alone rather than guessing.
- **A pending purchase says so** instead of looking like a failure.
- **`MainScreen` stops resolving the processor in a default argument**, which put the Play billing bind on the first-frame main thread purely to read one flag.

## Verification

- `:app:testFossDebugUnitTest --rerun-tasks` — **542 tests, 0 failures** (was 525 on the pre-merge base)
- `lintFossRelease` — clean, 5 Hints, no new warnings
- `:app:compileStoreDebugKotlin` — the flavour that actually contains this code compiles

Reviewed by a 5-lens adversarial workflow with 3-refuter verification; the two findings that survived refutation are the two defects fixed above. Three further findings were refuted on the bytecode (`enableAutoServiceReconnection` short-circuiting `isReady`, the acknowledge-token race, the Koin-in-`onResume` cost) and are not addressed here because they are not real.

Part of the 2026-08-04 dev-branch audit remediation.

## Summary by CodeRabbit

* **New Features**
  * Billing status and purchases now refresh automatically when returning to the app.
  * Improved handling of pending, acknowledged, and restored purchases.
  * Subscription prices now display the correct weekly, monthly, quarterly, semiannual, or annual billing period.
  * Added localized billing messages and price labels for Arabic, Spanish, French, and Simplified Chinese.

* **Bug Fixes**
  * Improved billing reconnection and purchase acknowledgement reliability, including safe, bounded retry handling.


trinadhthatakula:
Follow-up review found that both fixes in this PR were inert on device, and pushed a correction (`472dc88a`).

`.enableAutoServiceReconnection()` does not only enable lazy reconnection — it also short-circuits `isReady`. Verified with `javap` against the artifact Gradle resolves (`com.android.billingclient:billing:9.1.0`):

```
BillingClient$Builder.enableAutoServiceReconnection()  ->  putfield zza:Z (constant 1)
BillingClientImpl.<init>  (all four overloads)         ->  this.zzH = builder.zza
BillingClientImpl.isReady()                            ->  if (zzH) return true; else zzby()
```

So `isReady` answers `true` from construction onwards, on a client with no binding at all. Every guard in this file that meant *"the binding is up"* was therefore always taken:

- `scheduleReconnect`s `isReady -> reset()` arm always won, making `else -> connectToBilling()` unreachable — the ladder issued exactly **one** bind, never the documented 1/2/4/8/16 s ladder.
- `refreshPurchases` `isReady` branch always won, so `reconnect.onResume()` — the entire point of the re-arm — was dead code.

`getConnectionState()` is unaffected (raw `zzb` under the client monitor), so the four sites now read a new `isConnected`. The flag is **kept**: it genuinely covers the two deliberately unguarded call sites, `queryProducts` and `acknowledgePurchase` (the money path). But it reconnects with an *internal* listener and never calls this client's `onBillingSetupFinished`, so it cannot repair `_isBillingAvailable`/`_products` — which is why the ladder is not redundant. The KDoc asserted the inverse and is rewritten.

The same review also found the stuck-support-sheet defect had a second door, now swept: binding is local IPC and needs no network, so a first launch in airplane mode gets `OK` from `onBillingSetupFinished` and `SERVICE_UNAVAILABLE` from `queryProductDetails`. The connection never breaks, so nothing re-runs the catalogue — `queryProducts` only call site is the OK branch that needs a *reconnect*. `refreshPurchases` now re-runs it when the catalogue is empty.

Test count is unchanged at 542, deliberately: this file is not on `testFossDebugUnitTest`'s classpath, which is exactly how it shipped. 12 of the 14 findings raised in that review were refuted with full quorum.

PR #343, opened 2026-08-04
fix(installer): make the confirmed identity and the installed bytes one set

trinadhthatakula:
An archive handed to Thor by another app is untrusted input. Two of its strings used to decide different things — the entry the confirmation sheet parsed to name the app, and the entries `pm` was actually given — and nothing held those two to the same answer.

### The identity and the install set are now one set, by construction

An entry named `good\base.apk` could identify the install on screen and then be silently dropped from it: the sheet describing one APK while another was written. `isSafeEntryFileName` was enforced on both *write* paths but not on `extractEntryTo`, the read the **analyzer** uses to pick the identity, nor on plan resolution.

Rather than add the guard at each consumer, the untrusted names are filtered **once**, at plan resolution. `resolveBundleInstallSet` drops any entry whose leaf fails the check, and `resolveBundlePlan` derives `identityCandidates` *from* the surviving `installSet` — so identity is a subset of the install set by construction, and divergence is unrepresentable rather than merely absent. The manifest's own split list inherits the same guarantee for free: it is validated against leaves that already passed, so a list naming an unusable file reads as stale and is discarded **whole**, never shortened.

`extractEntryTo` gets the guard as a second lock. Every writer already refuses such a name; an identity must not be readable from one either.

**`entryNames` itself stays deliberately unfiltered**, and that is the interesting part. The classifiers must see what the archive actually holds — hide a backslash entry from them and the archive classifies as a *monolithic APK* and gets streamed whole, which is the same divergence one layer up. Verified instead that both sides converge on the whole file, which is the honest answer for an archive whose only APK entries are unusable.

### "Nothing was written" no longer means "this input is monolithic"

A resolved install set that selected zero entries fell through to the monolithic branch and streamed the **outer container** as `base.apk` — a file that by construction is not the one the identity was read from. `filesWritten = toWrite.isNotEmpty()` conflated *absent* with *everything was rejected*; empty meant unknown, not fine.

The fallthrough is now gated on the **absence** of an install set — `resolveInstallSetFromFile`'s own monolithic verdict, and the exact condition `AppAnalyzerImpl` checks before letting the whole file identify itself. A non-empty set that yields no writable entry refuses.

### Refusing instead of truncating

`extractEntries` now refuses on the terms its own KDoc already argued for the budget case: *half a bundle is not a smaller install, it is a different one.* Three refusal modes share one verdict — budget overrun, an unsafe wanted leaf, and a wanted name the archive does not hold — each deleting what was already written.

Two ordering details that are load-bearing: wantedness is checked **before** safety, so unrelated junk in an archive is still skipped quietly rather than failing the install; and the completeness check compares against the internally lowercased set, because the caller builds `wanted` case-sensitively and comparing raw sizes would spuriously refuse an archive holding both `base.apk` and `BASE.APK`.

## Riding along

- **A cancelled parse no longer strands its staged copy.** Ownership transfers on `analyze()`'s return, and a cancelled `withContext` *throws* instead of returning, so `analyzed` was never assigned and nothing could discard it — while the copy, a blocking non-cooperative loop, ran to completion anyway. Swiping the sheet away mid-parse is the ordinary way to leave.
- **Staging failures no longer strand their temp directory.** Cleanup is a wrapper rather than a move inside the rung's `try`: that `try` ends in a `catch (e: Exception)` which would fold `InstallRefusedException` into a generic error *and* stop it propagating — and propagation is what makes the ladder halt instead of dropping to a rung that would read the same refused bytes. Two of the three rungs stage into `externalCacheDir`, where the residue is user-visible.

## Verification

- `:app:testFossDebugUnitTest --rerun-tasks` — **578 tests, 0 failures, 0 skipped**
- `lintFossRelease` — clean, 5 hints, no new warnings
- 11 tests added, covering each fix: an unsafe leaf is in *neither* list; no plan of any shape offers a name a writer would refuse (all three routes in one archive); a manifest split list naming an unusable entry is discarded whole; a resolved set yielding no writable entry refuses instead of falling through; a bundle missing one split refuses rather than installing the rest; two spellings of one name are not counted as two files.

Reviewed by a 32-agent, 5-lens adversarial workflow with 3-refuter verification, run against the **post-fix** state. All 7 confirmed findings were regressions or gaps introduced by this branch's own first attempt, not pre-existing bugs — the guard had reached two of its three consumers.

Part of the 2026-08-04 dev-branch audit remediation.

## Summary by CodeRabbit

* **New Features**
  * Installation now uses a staged copy of the selected package for consistent results.
  * Added safer bundle planning and integrity verification.
* **Bug Fixes**
  * Improved cancellation, incomplete package, metadata, temporary-file, and installation failure handling.
  * Added cleanup for partial files and abandoned installation sessions.
* **Security**
  * Added protections against oversized archives, decompression bombs, unsafe filenames, path traversal, and invalid sidecar references.
* **Tests**
  * Expanded coverage for archive limits, cleanup, integrity checks, package selection, and failures.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
+++ b/app/src/main/java/com/valhalla/thor/data/repository/AppAnalyzerImpl.kt
@@ -14,122 +14,231 @@ import android.graphics.drawable.Drawable
 import android.net.Uri
 import android.os.Build
 import androidx.core.graphics.createBitmap
+import com.valhalla.thor.domain.model.AnalyzedPackage
 import com.valhalla.thor.domain.model.AppMetadata
+import com.valhalla.thor.domain.model.StagedPackage
 import com.valhalla.thor.domain.repository.AppAnalyzer
 import com.valhalla.thor.util.getDisplayName
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.withContext
 import org.koin.core.annotation.Single
 import java.io.File
 import java.io.FileOutputStream
+import java.security.MessageDigest
 import java.util.UUID
+
+/** Sub-directory of cacheDir holding staged installer inputs. */
+private const val STAGING_DIR_NAME = "staged_installs"
 
 @Single(binds = [AppAnalyzer::class])
 class AppAnalyzerImpl(private val context: Context) : AppAnalyzer {
 
-    override suspend fun analyze(uri: Uri): Result<AppMetadata> = withContext(Dispatchers.IO) {
+    override suspend fun analyze(uri: Uri): Result<AnalyzedPackage> = withContext(Dispatchers.IO) {
         val displayName = uri.getDisplayName(context)
         // Random, unpredictable temp names (CWE-377): avoids collisions between
         // concurrent analyses and predictable cache paths.
         val token = UUID.randomUUID()
+        val stagingDir = File(context.cacheDir, STAGING_DIR_NAME).apply { mkdirs() }
+        // The staged file outlives analyze() now, so a process death between the analysis and
+        // the install (or the dismissal) would strand a full copy of the input — hundreds of MB
+        // for an XAPK. Nothing else ever revisits this directory, so the sweep happens on the
+        // way in.
+        sweepStaleStagedPackages(stagingDir, System.currentTimeMillis())
         // The whole input is copied to disk once so it can be read with ZipFile
         // (random access via the central directory). ZipInputStream cannot handle
         // APKPure's STORED-with-data-descriptor entries (zero-size local headers) and
         // mis-reads the archive; ZipFile reads the central directory like `unzip`.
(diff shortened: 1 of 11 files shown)


trinadhthatakula:
Fixed in `318316a8`.

Confirmed on the code and it is the same defect class this branch already fixed once, applied to the teardown path and not swept to its sibling — which is this audit's dominant theme.

`parsePackage` released only the copy that had already been *handed over*. A parse still in flight owns a copy the ViewModel cannot see: `analyzed` is assigned when `analyze()` returns, so a discard landing mid-parse finds `null`, reclaims nothing, and the older parse then completes and overwrites the newer one's `analyzed`.

Cancelling releases it, and the machinery was already here: `AppAnalyzerImpl:110` does `if (!isActive) bundleFile.delete()`, and a cancelled coroutine never reaches `analyzed = analysis`.

No test — the suite has no main-dispatcher rule, so no ViewModel is reachable from it. That is why this branch's tests are all pure-logic. 578 tests / 0 failures, `lintFossRelease` clean.

PR #346, opened 2026-08-04
fix(gateway): name the user on every privileged command, per that command's own AOSP default

trinadhthatakula:
Audit item 2 of the 2026-08-04 dev-branch sweep.

## The bug

Thor ran most of its `pm` / `am` / `appops` lines with no `--user`, so the user each one acted on
was whatever the AOSP shell command happens to seed. Those seeds are **not the same**, and none of
them is "Thor's user":

| command | seed |
|---|---|
| `pm clear` / `enable` / `disable` / `path` | `USER_SYSTEM` |
| `pm install` / `uninstall` | `USER_ALL`, widened to `INSTALL_ALL_USERS` / `DELETE_ALL_USERS` |
| `am force-stop` | `USER_ALL` |
| `appops` | `USER_CURRENT`, resolved by `ActivityManager.getCurrentUser()` **inside system_server** — the globally foreground user, not the caller's |

On a single-user device these coincide and nothing is visibly wrong. On a work profile they diverge
silently: `UsageAccessManager` wrote the op through `appops` against the foreground user, then
confirmed it by reading `AppOpsManager` for `Process.myUid()`. The write and the read were asking
about two different users and never met.

## The fix

Every command now names `thorUserId` explicitly. The builders live in one new file,
`data/source/local/PerUserCommands.kt`, and each carries KDoc for **its own** seed rather than
inheriting a neighbour's — the seeds differ, so one shared comment would be wrong for most of them.

Things that look mergeable and are not, documented at each site:

- **`appops` needs care beyond the flag.** `parseUserPackageOp` consumes options before positionals,
  so a `--user` written *after* the package name parses as the **op**, and the command fails with
  "Unknown operation string" rather than a usage error. The builder places it first.
- **Two `MATCH_ARCHIVED_PACKAGES` version guards.** The constant is a `long` (`1L shl 32`,
  `since="35"`), so it cannot ride the `getPackageInfo(String, Int)` overload and needs
  `PackageInfoFlags` — API 33+, a different floor from the API 35 the constant itself requires.
- **AIDL methods are appended, never inserted or reordered.** Binder transaction codes are
  positional, so a stale daemon returns `false` rather than misinterpreting a call. Superseded forms
  are retained deliberately for that.
- **`pm set-installer` is left userless, verified:** it takes no `--user` on any supported release.
  Session sub-commands inherit their user from `install-create --user` at the point the session is
  made.

## Review follow-ups (last two commits)

A 43-agent adversarial review over 7 lenses returned no confirmed findings against the main change.
A second, tighter run upheld one, and I found one more myself:

- **Deleted `USER_ID_REGEX` from both gateways.** Declared beside the package-name regex that *is*
  used, and never referenced. The user id reaching those classes is an `Int` and cannot carry a
  shell metacharacter, so a validator that is never called only reads as coverage the callers do
  not have.
- **The reflective sweep compared the wrong thing.** It enumerated every boolean combination of a
  builder, then *joined* the renderings before comparing user 10 against user 11 — which hands back
  the hole the enumeration exists to close. A builder naming the user on only one branch still
  renders two different joins (the varying branch carries the whole string) and still contains the
  `--user 10` token, so both assertions pass. For the mirrored builder, joining is strictly *weaker*
  than the single-value render it replaced.

  Measured, not argued. A builder with exactly that shape, added temporarily:

  ```kotlin
  if (withUser) "pm probe --user $userId $pkg" else "pm probe $pkg"
  ```

  - against the joined sweep: **BUILD SUCCESSFUL**, 669 tests, blind to it
  - against this version: **fails**, naming the branch — `... emits the same thing for user 10 and
    user 11 on the branch [false]`

  The probe was reverted; `PerUserCommands.kt` is untouched by that commit.

## Verification

- `:app:testFossDebugUnitTest --rerun-tasks` → **669 tests, 0 failures, 0 skipped**
  (`PerUserCommandsTest` contributes 32)
- `:app:lintFossDebug` → **0 issues**
- Not device-tested. The work-profile divergence this fixes needs a second user to observe.

## Summary by CodeRabbit

* **New Features**
  * Added explicit user targeting for app installation, uninstallation, suspension, data clearing, cache management, and usage-access permissions.
  * Added support for managing apps across user profiles with improved state verification.
  * Improved visibility of uninstalled and archived apps in package information.

* **Bug Fixes**
  * Prevented package operations from unintentionally affecting other users.
  * Improved service connection cleanup and command failure reporting.
  * Increased reliability when freezing, unfreezing, suspending, and reinstalling apps.

* **Tests**
  * Added comprehensive coverage for user-scoped commands and permission queries.

### DIFF
--- a/app/src/main/aidl/com/valhalla/thor/rootservice/IThorRootService.aidl
+++ b/app/src/main/aidl/com/valhalla/thor/rootservice/IThorRootService.aidl
@@ -13,8 +13,8 @@ package com.valhalla.thor.rootservice;
  * back as an empty reply parcel -- false for a boolean, null for a String. A stale daemon therefore
  * degrades into an honest failure rather than a mis-dispatch.
  *
- * <p>That is why setAppSuspended below keeps its two-argument shape even though setAppSuspendedAs
- * supersedes it.
+ * <p>That is why setAppSuspended below keeps its two-argument shape, and setAppSuspendedAs its
+ * three-argument one, even though setAppSuspendedAsForUser supersedes both.
  */
 interface IThorRootService {
     boolean setAppSuspended(String packageName, boolean suspended);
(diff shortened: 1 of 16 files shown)


trinadhthatakula:
Both CodeRabbit comments reviewed. One applied, one deferred with a reason.

**1. `Dhizuku.kt` overload order — applied** (d1272b77).

Correct, and it belongs in *this* PR rather than a follow-up: the branch's whole thesis is that the
user id must actually reach the command, and this was a branch where it provably could not. Worse,
`Shizuku.clearCache` was reordered **on this same branch**, with a KDoc spelling out the reasoning —
and the sibling was left in the old order. That is the audit's dominant theme reproduced inside the
fix for it: the right argument already written down, next to an unfixed instance of the same defect.

One correction to the comment's framing, for the record: an earlier adversarial review pass had
*refuted* this finding, on the grounds that the whole reflective rung is transport-dead — `asInterface`
double-wraps the binder (Dhizuku's wrapper, then `ShizukuBinderWrapper` on top), so on a Dhizuku-only
device the call dies in a transport belonging to a privilege mode the user has not set up. That
refutation was about **impact**, not **correctness**, and I had wrongly let it stand in for both. A
rung that is dead today is still the rung a transport fix would wake up. So the reorder lands, with
the reachability caveat written into the KDoc rather than used to justify inaction.

The fallback stays gated on `NoSuchMethodException` alone and not `Exception`, deliberately: a
*refusal* of the per-user call must not fall through to one that clears a different user's cache and
reports it as this user's success.

**2. `ThorRootService.kt` null `IPackageDataObserver` — correct, and deliberately not in this PR.**

This is real and already tracked as the next branch (`fix/privileged-result-verification`), which is
in design as of today. It is not a one-line change and it is not about user ids:

- A `java.lang.reflect.Proxy` cannot be marshalled over binder, so receiving `onRemoveCompleted`
  needs a genuine `IPackageDataObserver.Stub` subclass — which means vendoring the AOSP AIDL into the
  module and depending on the boot classpath shadowing the generated class at runtime.
- That has to be assessed for hidden-API enforcement separately in the app process (Shizuku path) and
  the root process, for R8 keep rules, and for the FOSS flavour's build reproducibility requirement.
- It also needs a timeout policy, since a callback that never arrives must read as **unverified** —
  never as success, never as a hang.

It is one instance of a class of ~13 sites where a privileged operation reports success from a call
that cannot fail (`am force-stop` ends in an unconditional `return 0`; `runSuspend` returns 0 whenever
`setPackagesSuspendedAsUser` didn't throw and discards the failure array). Folding one of them into a
user-id PR would land the least-testable part of that change without its siblings or its test plan.

PR #347, opened 2026-08-04
fix(privilege): judge privileged actions by a readback, not by an exit code

trinadhthatakula:
Item 3a of the 2026-08-04 dev-branch audit remediation. Eight sites across the three gateways reported success on evidence that could not fail — and in most cases the honest verifier was **already written in the same function**, sitting unreachable behind the short-circuit that always won.

## Why the old evidence was worthless

| Fact | Consequence |
|---|---|
| `ActivityManagerShellCommand.runForceStop` ends in an unconditional `return 0` | `am force-stop` exits 0 for a package nothing is running for, and for one not installed for `--user N` at all |
| `PackageManagerShellCommand.runSuspend` returns 0 whenever `setPackagesSuspendedAsUser` did not throw | the failure array naming packages it declined to suspend is discarded; an exempt package is a clean exit |
| `IActivityManager.forceStopPackage` and `IAppOpsService.setMode` return `void` | a reflective call can only report that the binder did not throw |

## What changed

**Shizuku** — `forceStopApp` rungs 1 and 2 answer to `isAppStopped`; `setAppRestricted` gained a `checkOperation` readback, with the op code and uid resolved once and shared.
**Dhizuku** — the same for `forceStopApp`, plus a background-mode readback; `clearCache`/`clearAppData`'s reflection rungs now report failure rather than claiming a wipe they cannot see.
**Root** — `forceStopApp` gained a shared `isStoppedNow()` (three hand-written copies of one verifier is how one of them stops matching the others); the API 28 suspend rung is judged by `readSuspendedFlag`; `clearAppData` now names *which* of three ways the daemon produced nothing instead of folding them into "AIDL failed".

### Two readback policies, deliberately not unified

Where the rung's own error reporting is honest — `appops set` exits non-zero for an unknown package or op, `setMode` throws on denial — an unreadable state **fails open**, so a restriction that really was applied is not reported as a failure. Where "did not throw" is the whole of the evidence, and on Dhizuku's double-wrapped transport is not even a statement about the target service, it **fails closed**. Every readback is tri-state: `null` means "could not tell" and never stands in for success.

## Deleted as unfixable rather than left unfixed

- `SystemRepository.aggressiveCleanup` — force-stop then clear-cache, **both** `Result`s discarded, then unconditional `Result.success(Unit)`. Zero production callers; its only references were the declaration and three interface-forced test overrides, none of which exercised it.
- `Packages.forceStopApp` / `setAppDisabled` / `setAppRestricted` — unprivileged `Bypass` reflection from Thor's own uid, wrapped in `runCatching { ...; true }`. Deleted as one edit because they only ever called each other: `setAppDisabled` was `forceStopApp`'s only caller and itself had none.

Both are locked by surface tests. **The anti-vacuity guard on the repository one paid for itself on its first run**: methods returning `Result` are JVM-mangled to `forceStopApp-gIAlu-s`, and `aggressiveCleanup` returned `Result<Unit>` too — so the absence assertion was green whether or not the method was there, until the de-mangler learned to strip the inline-value-class suffix as well as the `internal` one.

## Correction to my own earlier reasoning

An adversarial review of the audit killed the Dhizuku overload-order finding as "transport-dead, therefore unreachable". That is an argument about **impact**, not about **correctness** — a rung that is dead today is still the rung a future transport fix would wake up. The order is fixed here, with the transport argument recorded next to it so the next reader gets both halves.

## Verification

- `:app:testFossDebugUnitTest --rerun-tasks` — **673 pass** (+4 new)
- `:app:lintFossDebug` — clean

## Not covered here, and why

The two `IPackageDataObserver` sites (`Shizuku.kt` clear-cache/clear-data, `ThorRootService`) are held back for a separate PR. They need a vendored AIDL and a real `Stub` subclass, and if the callback never arrives on a device, clears that *did* work would newly report failure — so that change is merge-blocked on hardware testing that this one is not.

Residual, unchanged by this PR and stated rather than buried: `forceStopApp`'s reliance on `FLAG_STOPPED` being readable promptly after the call returns, and `checkOperation` reporting the mode *in force* rather than a record of the write, are both device-observable and unmeasured.

## Summary by CodeRabbit

- **Bug Fixes**
  - System actions now verify that force-stop, app suspension, restriction, cache clearing, and data clearing actually succeeded.
  - Improved error reporting distinguishes command failures, refusals, binding issues, and unconfirmed state changes.
  - Privileged operations use more reliable fallback behavior and state checks.

- **Refactor**
  - Removed the obsolete aggressive cleanup operation while retaining individual force-stop and cache-clearing actions.

- **Tests**
  - Added regression coverage to ensure removed operations stay unavailable and supported operations remain present.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -206,6 +206,10 @@ class RootSystemGateway(
         // caller here treats false as "not proven stopped" and does more work, ending at the
         // failure below; so "could not read" costs a wasted killBackgroundProcesses and a reported
         // failure, never a success that did not happen.
+        //
+        // What it does cost is a *sentence*. "FLAG_STOPPED is clear" and "the package could not be
+        // read" are the same `false` here, so the failure message must not speak for both — see the
+        // last read below, which keeps its `ApplicationInfo` for exactly that.
         fun isStoppedNow(): Boolean = getApplicationInfoCompat(packageName)?.run {
             (flags and android.content.pm.ApplicationInfo.FLAG_STOPPED) != 0
         } ?: false
@@ -241,21 +245,36 @@ class RootSystemGateway(
             am?.killBackgroundProcesses(packageName)
         }
 
-        if (isStoppedNow()) return Result.success(Unit)
+        // The last read is spelled out rather than asked for through [isStoppedNow], because the
+        // failure message below has to say *which* of that function's two `false`s this is, and
+        // re-reading to find out would describe a different moment than the one that decided.
+        val postKillInfo = getApplicationInfoCompat(packageName)
+        if (postKillInfo != null &&
+            (postKillInfo.flags and android.content.pm.ApplicationInfo.FLAG_STOPPED) != 0
+        ) {
+            return Result.success(Unit)
+        }
 
         // Two ways to arrive here now, and a bug report has to be able to tell them apart: the
         // shell command itself failed, or it exited 0 and the app kept running anyway. The old
         // message asserted the first unconditionally, which the guard above has just made false.
         val shellVerdict = if (shellResult.isSuccess) {
-            "`am force-stop` exited 0 but the app never entered the stopped state"
+            "`am force-stop` exited 0"
         } else {
             "the shell command failed"
         }
+        // The same defect one level down, and the reason this is not simply "it is still running":
+        // that claim rests on [isStoppedNow], where an unreadable `ApplicationInfo` and a genuinely
+        // clear FLAG_STOPPED are indistinguishable. A bug report generated from "Thor could not
+        // read the package" must not read as "the kill did not work" — they need different fixes.
+        val stateVerdict = if (postKillInfo == null) {
+            "the package's ApplicationInfo could not be read back after killBackgroundProcesses, " +
+                "so whether it is still running is unknown"
+        } else {
+            "FLAG_STOPPED is still clear after killBackgroundProcesses, so it is still running"
+        }
         return Result.failure(
-            Exception(
-                "Root force stop of $packageName failed: $shellVerdict, and it is still " +
-                    "running after killBackgroundProcesses."
-            )
+            Exception("Root force stop of $packageName failed: $shellVerdict, and $stateVerdict.")
         )
     }
 
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -196,19 +196,44 @@ class RootSystemGateway(
             return Result.failure(IllegalArgumentException("Invalid package name: $packageName"))
         }
         val escapedPackage = packageName.escapeForShell()
+
+        // What "stopped" means here, asked of the platform rather than of an exit code. All three
+        // rungs below ask this one question, so it is one function: three hand-written copies of a
+        // verifier are how one of them quietly stops matching the others.
+        //
+        // An unreadable package answers false, and that collapse is safe in the one direction that
+        // matters — the opposite of the `null` [readSuspendedFlag] is careful to preserve. Every
+        // caller here treats false as "not proven stopped" and does more work, ending at the
+        // failure below; so "could not read" costs a wasted killBackgroundProcesses and a reported
+        // failure, never a success that did not happen.
+        fun isStoppedNow(): Boolean = getApplicationInfoCompat(packageName)?.run {
+            (flags and android.content.pm.ApplicationInfo.FLAG_STOPPED) != 0
+        } ?: false
+
         // The two halves of this function have to name the same user. Everything below the shell
         // rung is in-process — killBackgroundProcesses and the FLAG_STOPPED reads both answer for
         // Thor's own user — while a bare `am force-stop` reads USER_ALL and kills the package
         // everywhere. Naming [thorUserId] is what makes the post-check evidence about the process
         // the command was aimed at.
         val shellResult = runCommand(forceStopCommand(escapedPackage, thorUserId))
-        if (shellResult.isSuccess) return shellResult
+        // The exit code alone decided this, and it cannot say no:
+        // `ActivityManagerShellCommand.runForceStop` ends in an unconditional `return 0`, so it
+        // reports that the command parsed, never that a process died. `shellResult.isSuccess` was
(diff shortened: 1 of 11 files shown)


trinadhthatakula:
### Residual findings from this sweep, traced but deliberately not fixed here

Recording these so they are not lost with the branch. None is a correctness defect; all are dead surface found while auditing the eight sites above, and each would widen this PR past the one thing it is about.

**Unreferenced members** (no callers anywhere in the repo — verified with a control pattern, since an unquoted glob under zsh returns empty with exit 0 and looks exactly like a clean result):

- `Packages.packageUri`
- `Packages.getInstalledApplications`
- `Packages.getUnhiddenPackageInfoOrNull`
- `Packages.isPrivilegedApp`
- `ShizukuReflector.isPrivilegedApp` — a near-copy of the `Packages` one, so both would go together

`Packages.packageUid` was checked as part of the same sweep and **is** live: `Shizuku.kt`, `Dhizuku.kt` (two sites) and `ShizukuReflector.kt` (two sites) all call it. It stays.

**House-rule deviation:** `Shizuku.forceStopApp` builds `"am force-stop --user $userId $packageName"` by hand, using neither `forceStopCommand()` from `PerUserCommands.kt` nor `escapeForShell()` — the only shell line in that object that skips both. Harmless today (package names arrive from `PackageManager`), but it is the shape the rest of the file was standardised away from.

**Generic failure message:** `ShizukuSystemGateway`'s freeze path returns one failure string for several distinct causes, in the same way `RootSystemGateway.clearAppData` did before this PR split it into three. Same fix, different file.

**Stale doc pointers:** `ShizukuSystemGateway` says "Rungs 1 and 2 both live inside `Shizuku.setAppDisabled`", but `setAppDisabled` is a one-line delegate and the rungs are in `setAppDisabledDetailed`. A reader following that pointer lands somewhere the rungs are not. I fixed the identical stale pointer in `Dhizuku.kt` here because it sat on code this PR was already changing; this one does not, so it is listed rather than swept.

trinadhthatakula:
Revision pushed (`e95aa903`) after an adversarial review pass. The headline: **the review found this branch still shipping the exact defect it exists to delete**, one rung down from where it was fixed.

### The bug the branch had

`Shizuku.setAppRestricted`'s reflection fallback ended in `reflectionRan && opReadsBackAsExpected() != false`, and the comment defending that `!= false` said `setMode` throws when it is denied.

That claim is false everywhere in Thor's range, verified against AOSP release by release:

- **< API 30** — a package/uid mismatch meets `Slog.w("Bad call: specified package … but it is really …")` and a `null` return out of `getOpsRawLocked`. `setMode` changes nothing and returns normally.
- **API 30+** — the check *does* throw, but from inside `verifyAndGetBypass`, where `setMode` catches it: `Slog.e(TAG, "Cannot setMode", e); return;` (and via `logVerifyAndGetBypassFailure` on 35+). Current AOSP adds a *second* silent return ahead of it for `!isIncomingPackageValid(...)`.

The only refusal that ever reaches the caller is `enforceManageAppOpsModes`, which sits outside the try on every release checked. So on a denied write: `reflectionRan` is `true`, an unreadable `checkOperation` is `null`, and `true && (null != false)` reported **success with nothing restricted**. Now `== true`.

This is also @coderabbitai's nitpick on that line, reached independently — thank you, it was right and it was load-bearing.

### The rule, now stated once

> A rung may fail **open** on an unreadable read-back only if it has a self-report that is evidence *independent* of the read-back. With no such evidence it fails **closed**.

Rung 1 has that evidence — `appops set` exits non-zero for an unknown package or op — and keeps `!= false`. Rung 2 has none, because a `void` return is not a mode. The clear-data sites fall out of the same rule rather than being a separate convention.

`reflectionRan` stays in the expression as a **guard, not as evidence**: for `restricted = false` the expected mode is `MODE_ALLOWED`, which is also `RUN_ANY_IN_BACKGROUND`'s platform default — so an op nobody ever wrote reads back as expected, and without the conjunct a throwing `setMode` would be reported as success by the default mode alone.

### Three comments this branch got wrong, corrected

- **`verifyIncomingUid`** covers API 28–29 only. `checkOperation` dropped it in 30; current AOSP passes `shouldVerifyUid = false` there explicitly while `noteOperation`/`startOperation` pass `true`. Routing the read through the privileged binder is still right, but for reasons that actually span 28..37.
- **`DhizukuHelper.forceStopApp` rung 1 is not "alive"** — the claim the read-back was justified by. `execute` runs `am` *inside the Dhizuku app* via an AIDL call to `IDhizuku.remoteProcess`, at Dhizuku's ordinary app uid, and `ActivityManagerService.forceStopPackage` opens on `FORCE_STOP_PACKAGES` (`signature|privileged`), which device owner does not confer. The identical failure shape one subcommand over is already **measured on device** in that same file: `am get-current-user` denied to uid 10231. The read-back stays anyway — it costs nothing on a rung that short-circuits, and it guards the ROM that *does* hold the permission.
- **`buildSuspendDialogInfo`'s KDoc** had Shizuku's answer pasted into Dhizuku: it claimed a `@StringRes` id would resolve against `com.android.shell`. Dhizuku's only caller passes `BuildConfig.APPLICATION_ID`.

### Known blind spot, recorded rather than fixed

`checkOperationUnchecked` returns a **uid-level** mode before it ever consults the package entry. Both of Thor's writes are package-level, so a uid-level mode on `RUN_ANY_IN_BACKGROUND` would make this read report something unrelated to whether the write landed. `checkOperationRaw` is not the fix — `raw` drops the foreground evaluation, not the uid short-circuit. Mechanism is AOSP-verified; **no uid-level writer of that op was identified**, so it is documented as a blind spot, not claimed as a bug.

### Tests

675 unit tests, 0 failures; lint clean.

Both surface tests gained a **source-text sweep**, because reflection is structurally blind to an extension function, a companion member without `@JvmStatic`, and a `@JvmName` rename — all three leave the call site reading exactly as before. Each sweep runs its guards first (corpus size, anchor file non-empty, the declaration pattern proven against a declaration known to be present, the extension pattern proven against its own shape), because an absence assertion is the one test shape that passes best when it is looking at nothing.

Not taken on trust: I mutated a production file with both offending declarations and confirmed **both** new tests go red, then reverted. `surfaceNames()` also now unions `declaredMethods` with `methods`, so moving a member to a super-interface no longer takes it out of view.

### Still open, and deliberately not silently fixed

The force-stop read-back disclosed in the PR body stands as-is. `FLAG_STOPPED` is cleared again by five AOSP paths, so a negative read cannot distinguish "the platform refused" from "stopped, then restarted" — and retrying widens the window rather than closing it. Proving it either way needs pid-set evidence and a device. Documented, not papered over.

PR #348, opened 2026-08-04
fix(privilege): wait for the observer instead of reporting a wipe that was only dispatched

trinadhthatakula:
> [!NOTE]
> **Device-tested 2026-08-05 — the merge gate is discharged.** Measured at uid 2000 (Shizuku's own privilege level) on a live Android 16 / API 36 image, with a real `IPackageDataObserver.Stub` built from this PR's vendored aidl. `clearApplicationUserData` calls back in **96 ms** and the data is verifiably gone; the cache path never calls back **and the cache is verifiably still there**, with `PackageManager: Calling uid 2000 does not have android.permission.INTERNAL_DELETE_CACHE_FILES, silently ignoring` in logcat as the witness. The regression this was gated on — a working clear newly reporting failure — does not exist. See [the results comment](https://github.com/trinadhthatakula/Thor/pull/348#issuecomment-5185202498).
>
> One consequence worth knowing before the bug reports arrive: **Shizuku "clear cache" now reports failure whenever Shizuku was started over adb**, because it never worked. Root-started Shizuku (uid 0) is unaffected.

Item 3b of the 2026-08-04 dev audit, stacked on #347 (now merged).

## The defect

Three sites cleared user data or cache and reported success because a `void` binder call returned.

| Site | Call | What it returns |
|---|---|---|
| `Shizuku.clearAppData` | `IPackageManager.clearApplicationUserData` | `void` |
| `Shizuku.clearCache` | `IPackageManager.deleteApplicationCacheFilesAsUser` | a boolean that says nothing about the outcome |
| `ThorRootService.clearAppData` | `IPackageManager.clearApplicationUserData` | `void` |

In every case `PackageManagerService` posts the work to its own handler and reports the result on `IPackageDataObserver.onRemoveCompleted`. **All three passed `null` there.** At shell uid PMS additionally *accepts* cache clears it then declines — it logs that it is silently ignoring the request, deletes nothing, and tells the caller none of it.

So `true` meant "the call did not throw", for the most destructive operation Thor performs. The daemon site was worse than the other two: it has no `PackageManager` for the caller's user and no readback to compare against, so `runCatching { … }.isSuccess` over a void method was the *entire* verifier.

## The fix

`awaitDataObserver` builds a real observer, hands it to the privileged call, and waits once and briefly for the verdict.

Its invariant is structural rather than remembered: the result **starts** at `UNVERIFIED` and the callback is the only thing that ever writes it. `fire` throwing, the class being missing, the latch timing out, a verdict landing after the timeout — every one of those leaves it where it started, because none of them is a write. First verdict wins, so a late `true` cannot upgrade a refusal.

**A throw from `fire` maps to `UNVERIFIED`, never `REFUSED`.** From inside the function a `SecurityException` from PMS and a transport that died before PMS ever saw the call are the same object; calling the second one a refusal would claim knowledge of a conversation that never happened.

Both non-`CLEARED` values collapse to `false` at every caller. That direction is deliberate: clearing twice costs a user nothing, a false "done" costs them the chance to try a privilege mode that would have worked. The three outcomes are distinguished in the log rather than in the return type, because every consumer is a `Boolean`.

## Why a vendored aidl

A `java.lang.reflect.Proxy` cannot be marshalled over binder, so the callback has to be a genuine `IPackageDataObserver.Stub` subclass — and `Stub` is not in the public SDK. AOSP's aidl is vendored **purely to give the Kotlin compiler something to subclass**. At runtime `PathClassLoader` delegates parent-first, the boot classpath's framework class wins, and the generated copy is never loaded, so the object extends the *framework* `Stub` and the framework's own `onTransact` dispatches to it.

That is also why R8 must not rename `onRemoveCompleted`: nothing in the dex calls it, a binder transaction does. **Verified in the release mapping rather than assumed** —

```
com.valhalla.thor.data.source.local.PackageDataObserversKt$newDataObserver$1
  -> com.valhalla.thor.data.source.local.PackageDataObserversKt$newDataObserver$1
     void onRemoveCompleted(java.lang.String,boolean) -> onRemoveCompleted   ← kept

com.valhalla.thor.data.source.local.PackageDataObserversKt$newDataObserver$2 -> if2   ← JVM-only fallback, free to move
```

## Two things found along the way

**The aidl made lint see a call that was always there.** Vendoring it turned `ThorRootService`'s `getDeclaredMethod("clearApplicationUserData", …)` into a `SoonBlockedPrivateApi` error, on a line this PR does not touch. Lint builds a descriptor from the *argument classes*, and the second one is `Class.forName("android.content.pm.IPackageDataObserver")` — unresolvable before, so no descriptor and no report. Measured as a single variable: an `origin/dev` worktree lints clean; the same worktree with **only** that aidl file added reports the error on the identical line. Suppressed with the reason written down.

**The daemon needs no hidden-API exemption, and adding one would not be free.** A bare `app_process` is never zygote-specialised, so it never goes through `ActivityThread.handleBindApplication` and non-SDK enforcement is never switched on for it. The proof is already in that file: everything in it reaches `IPackageManager` and `ServiceManager` with plain reflection and no `Bypass` at all — including the `Class.forName` of this very observer. `Bypass.setHiddenApiExemptions` would mmap and parse core-oj on every daemon start, with no offset cache to hit, to guard against an enforcement that is off.

## Tests

693 unit tests (18 new), 0 failures. Lint clean on debug *and* release.

`DataClearOutcomeTest` drives the latch through every path, including the two that matter — a verdict arriving *after* the wait expired, and a second callback trying to overturn the first.

It reaches that logic through the non-binder fallback, which exists for exactly this reason: AGP's mockable `android.jar` rewrites constructors but makes every other method throw, and the generated `Stub()` constructor calls `Binder.attachInterface`. Without the fallback, `object : IPackageDataObserver.Stub()` cannot be constructed on the JVM and **every case would collapse to `UNVERIFIED` for the same uninteresting reason** — a green suite proving nothing.

`ObserverCallSitesTest` sweeps the source so a fourth clear site cannot be added with a `null` observer, and includes a test that the sweep itself can fail.

## What needed a device, and what came back

The JVM tests prove the latch. They could not prove the thing this PR bets on: **that `onRemoveCompleted` arrives, within 15 s, on a real ROM.** That has now been measured — [full results here](https://github.com/trinadhthatakula/Thor/pull/348#issuecomment-5185202498).

| # | Item | Result |
|---|---|---|
| 1 | Shizuku — clear data | ✅ `CLEARED` in **96 ms**, `files/` and `cache/` verified gone |
| 1 | Shizuku — clear cache | ✅ measured: no callback, cache untouched, PMS logs that it is ignoring the call. Failure is the true answer |
| 2 | Root daemon (uid 0) | ⚠️ not run — the available image is a Play build and refuses `adb root`. Same code path at a strictly higher privilege; argued in the comment rather than measured |
| 3 | A package PMS refuses | ✅ item 1's cache case *is* that, with PMS's own log line as the witness and the filesystem as the check |
| 4 | Timeout not load-bearing | ✅ 96 ms against a 15 000 ms bound |
| 5 | Release build | ✅ via `mapping.txt` above; not additionally run on device |

Review also turned up a real cost of the wait that no device could have shown: `MainViewModel.performLoggedMultiAction` clears a multi-selection **sequentially**, so a ROM that never calls back turned a fifty-app sweep into fifty consecutive fifteen-second waits. Whether the callback arrives is a property of the transport, not of the package, so the first full-length timeout is evidence rather than bad luck — `awaitDataObserver` now caps the *next* wait at 1500 ms, and any arriving verdict restores full patience. Shortened, never skipped: skipping would make `UNVERIFIED` structural and would permanently misreport a merely-slow ROM, which is the exact false negative this branch exists to avoid creating.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -51,3 +51,22 @@
 -keep class com.valhalla.thor.rootservice.** { *; }
 -keep interface com.valhalla.thor.rootservice.** { *; }
 
+# --- IPackageDataObserver (framework AIDL callback, shadowed at runtime) --------------------------
+# This file is the one the RELEASE build type wires in (`proguardFiles(..., "proguard-rules.pro")`),
+# so it covers both flavours -- `proguardFile("proguard-rules-foss.pro")` on the foss flavour ADDS to
+# that list rather than replacing it. The rule has to live here and not there: the foss file's only
+# blanket keep, `-keep class com.valhalla.thor.** { *; }`, is commented out, and it would not have
+# covered `android.content.pm.**` anyway.
+#
+# app/src/main/aidl/android/content/pm/IPackageDataObserver.aidl is an AOSP copy that exists only so
+# the Kotlin compiler has a Stub to subclass. At runtime PathClassLoader delegates parent-first and
+# the boot classpath's real framework class wins, so Thor's observer extends the FRAMEWORK Stub and
+# the FRAMEWORK's onTransact dispatches to it -- by the original method name and descriptor, which
+# R8 has no way to know about: nothing in the dex calls onRemoveCompleted, a binder transaction does.
+# Renaming the override would leave onTransact unable to find it, so the callback would never fire,
+# every clear would report "could not confirm" forever, and only in release, because debug is not
+# minified. Same failure shape as the rootservice block above.
+-keep class android.content.pm.IPackageDataObserver { *; }
+-keep class android.content.pm.IPackageDataObserver$Stub { *; }
+-keep class * extends android.content.pm.IPackageDataObserver$Stub { *; }
+
--- a/app/src/main/aidl/android/content/pm/IPackageDataObserver.aidl
+++ b/app/src/main/aidl/android/content/pm/IPackageDataObserver.aidl
@@ -1,0 +1,56 @@
+// SPDX-FileCopyrightText: 2007 The Android Open Source Project
+// SPDX-License-Identifier: Apache-2.0
+//
+// This file is AOSP's interface definition, not Thor's code, so it deliberately does NOT carry
+// Thor's usual `2025-2026 Trinadh Thatakula` / `GPL-3.0-or-later` pair. Apache-2.0 combined into a
+// GPL-3.0-or-later work is a compatible one-way combination; relabelling it would be a licence
+// claim over someone else's text. There is no CI licence-header gate, so a "fix the SPDX header"
+// reflex is the only thing that can get this wrong -- do not let it.
+//
+// Copied from frameworks/base/core/java/android/content/pm/IPackageDataObserver.aidl. The only
+// deviation from AOSP is the removal of `@UnsupportedAppUsage` on the method: that annotation lives
+// in `android.compat.annotation`, which is not on an application module's aidl import path, and it
+// is metadata for the platform's own greylist rather than part of the interface's shape.
+//
+// ---------------------------------------------------------------------------------------------
+// WHY A COPY EXISTS, AND WHY IT IS NEVER LOADED
+// ---------------------------------------------------------------------------------------------
+// `IPackageManager.clearApplicationUserData(String, IPackageDataObserver, int)` returns void and
+// `deleteApplicationCacheFiles(String, IPackageDataObserver)` returns a boolean that means nothing;
+// in both cases the real verdict arrives asynchronously on `onRemoveCompleted`. Receiving it needs
+// a genuine `IPackageDataObserver.Stub` subclass, because a `java.lang.reflect.Proxy` cannot be
+// marshalled over binder -- there is no way to hand the framework a callback object without one.
+// `Stub` is not in the public SDK, so the Kotlin compiler has nothing to subclass unless this file
+// generates one.
+//
+// At runtime the generated copy is dead weight and that is the point. `PathClassLoader` does not
+// override `ClassLoader.loadClass`, so resolution is parent-first: `BootClassLoader` supplies the
+// framework's real `android.content.pm.IPackageDataObserver$Stub` and the class generated from this
+// file is never loaded. Thor's observer therefore extends the FRAMEWORK Stub, and dispatch is done
+// by the framework's own `onTransact` using the framework's own transaction codes. That is the
+// safety property worth stating plainly: because we subclass rather than reimplement `onTransact`,
+// a change to the platform's transaction numbering cannot mis-dispatch us. It is the opposite
+// situation to `com/valhalla/thor/rootservice/IThorRootService.aidl`, whose append-only rule exists
+// precisely because those codes ARE ours to keep stable.
+//
+// Every way this can fail degrades to "could not confirm", never to a false success -- if the
+// framework class ever disappeared, the generated copy would load instead, the framework would
+// never call our object, `awaitDataObserver` would time out, and the clear would report failure.
+// See `data/source/local/PackageDataObservers.kt` for the receiving half and
+// `app/proguard-rules.pro` for why R8 must not rename the override.
+//
+// `oneway` and the parameter names/order are kept exactly as AOSP has them. `oneway` is irrelevant
+// to the receive path (the framework Stub does the dispatch), but a faithful copy is far easier to
+// re-verify against AOSP later than a tidied one.
+
+package android.content.pm;
+
+/**
+ * API for package data change related callbacks from the Package Manager.
+ * Some usage scenarios include deletion of cache directory, generate
+ * statistics related to code, data, cache usage(TODO)
+ * {@hide}
+ */
+oneway interface IPackageDataObserver {
+    void onRemoveCompleted(in String packageName, boolean succeeded);
+}
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -336,7 +336,7 @@ class RootSystemGateway(
                 return@withContext Result.success(Unit)
             }
             daemonVerdict = aidlCall.fold(
-                onSuccess = { "the root daemon refused the wipe" },
+                onSuccess = { "the root daemon answered no" },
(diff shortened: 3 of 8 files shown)


trinadhthatakula:
## Device results — the gate is discharged, and the answer is better than "no regression"

Run on a live Android 16 / API 36 image (`sdk_gphone64_arm64`) at **uid 2000**, which is exactly the uid Shizuku's service runs as when it is adb-started. That let the mechanism be measured at Shizuku's own privilege level without needing Shizuku itself in the loop.

The probe is a ~10 KB dex run through `app_process`: everything reflection except `IPackageDataObserver`, which is a real `Stub` subclass compiled against **this PR's vendored aidl**. So it exercises the same parent-first-delegation trick the PR argues for, rather than a paraphrase of it. Build recipe at the bottom.

### 1. `clearApplicationUserData` — the callback arrives, fast, and it is telling the truth

```
uid=2000 pkg=com.valhalla.thor.debug user=0
IPackageManager = android.content.pm.IPackageManager$Stub$Proxy
clearApplicationUserData:
  data: call returned null
  -> CLEARED (reported=com.valhalla.thor.debug) after 96ms
```

Paired against the filesystem rather than believed:

```
before:  files/probe.txt (24 bytes)   cache/probe.bin (512 KB)
after:   ls: files: No such file or directory
         ls: cache: No such file or directory
```

96 ms against a 15 000 ms bound. **Checklist items 1 and 4 answered for the data path**: the observer transport works, the framework `Stub` dispatches into our subclass, the echoed package name is correct, and the timeout is generous by two and a half orders of magnitude rather than load-bearing.

### 2. `deleteApplicationCacheFiles` — no callback, ever. And the cache is still there.

Both overloads, 15 s each:

```
deleteApplicationCacheFiles:        -> UNVERIFIED(timeout) after 15015ms
deleteApplicationCacheFilesAsUser:  -> UNVERIFIED(timeout) after 15004ms
```

This is the case the PR was gated on — *"if the callback does not arrive, a clear that worked newly reports failure"*. It does not arrive. So the question is whether the clear worked, and it is answerable:

```
before cache clear:  cache/probe.bin  512 KB
after  cache clear:  cache/probe.bin  512 KB   ← untouched
```

And `PackageManagerService` says why, in its own words:

```
W PackageManager: Calling uid 2000 does not have
                  android.permission.INTERNAL_DELETE_CACHE_FILES, silently ignoring
```

`INTERNAL_DELETE_CACHE_FILES` is `signature|privileged`. uid 2000 cannot hold it, on any build. **So the new `UNVERIFIED` is not a false failure — it is the first true answer this path has ever given.** The feared regression does not exist here because there was no working operation to regress.

### 3. The control reproduces the bug in one line

```
control: deleteApplicationCacheFiles(pkg, null) returned null and threw nothing
```

That is precisely what `main` does today, and precisely what it reports as success — while PMS logs that it is ignoring the request and the cache sits there untouched.

### 4. Rung 1 measured too, so the user-visible verdict is fully accounted for

Thor tries shell first. Issued verbatim as `Shizuku.clearCache` builds it:

```
$ rm -rf /data/user/0/<pkg>/cache /storage/emulated/0/Android/data/<pkg>/cache
rm: /data/user/0/<pkg>/cache: Permission denied
EXIT=1
```

External cache: **deleted**. Internal cache: **Permission denied**, exit 1, so rung 1 correctly reports failure and falls through to rung 2, which is §2 above.

Full user-visible truth for adb-started Shizuku, then: *clear cache removes the external cache and nothing else.* Before this PR Thor called that a success. After it, Thor calls it a failure. Neither label is perfect — something did happen — but the internal cache is the one users mean and the one that is left behind, so the failure is much the closer of the two.

### What this changes for users, stated plainly

**Shizuku "clear cache" will now report failure on every device where Shizuku was started over adb.** That will look like a regression in a bug report and is not one; it is a seven-line permission check in AOSP surfacing for the first time. Worth knowing before the reports arrive.

Root-started Shizuku runs at uid 0, where the permission check is bypassed outright, so that configuration is unaffected — and is the only one where clearing another app's internal cache is possible at all.

### Checklist

| # | Item | Status |
|---|---|---|
| 1 | Shizuku — clear data | ✅ `CLEARED` in 96 ms, data verified gone |
| 1 | Shizuku — clear cache | ✅ measured, and the honest answer is failure — see §2/§4 |
| 2 | Root daemon (uid 0) | ⚠️ **not run** — the available image is a Play build and refuses `adb root`. Argued below. |
| 3 | A package PMS refuses | ✅ §2 is that case, with PMS's own log line as the witness and the filesystem as the check |
| 4 | Timeout not load-bearing | ✅ 96 ms against 15 000 ms |
| 5 | Release build | ✅ via `mapping.txt` (in the PR body); not additionally run on device |

**On item 2, and why I am not treating it as still open.** The daemon uses the same `awaitDataObserver`, the same vendored-aidl `Stub`, and the same `clearApplicationUserData` — the one call proven above to deliver its callback in 96 ms. It differs only in running at uid 0, where every permission check the callback path could hit is bypassed. There is no mechanism by which root gets *less* than shell here. That is an argument rather than a measurement, and I would rather say so than dress it up; if it turns out wrong the symptom is loud and local (root clear-data reports failure while Shizuku's works).

<details>
<summary>Reproducing the probe</summary>

```bash
BT=$ANDROID_HOME/build-tools/37.0.0-rc2
$BT/aidl -Iaidl -ogen app/src/main/aidl/android/content/pm/IPackageDataObserver.aidl
javac --release 17 -cp $ANDROID_HOME/platforms/android-37.0/android.jar -d classes $(find gen src -name '*.java')
$BT/d8 --min-api 28 --output . $(find classes -name '*.class')
adb push classes.dex /data/local/tmp/obsprobe.dex
adb shell "CLASSPATH=/data/local/tmp/obsprobe.dex app_process /data/local/tmp probe.Probe <pkg> 0 all"
```

`Probe.main` takes `<pkg> [userId] [cache|cacheAsUser|data|control|all] [timeoutMs]`. The single-op forms exist because `clearApplicationUserData` wipes the cache dir too, which would make any before/after cache measurement meaningless if the survey ran in one pass.

A bare `app_process` is never zygote-specialised, so non-SDK enforcement is never switched on for it — the same reason the PR gives for `ThorRootService` needing no `Bypass` exemption, and the reason the plain reflection above works.

</details>

PR #362, opened 2026-08-06
feat(ci): three-rung release ladder — dev uploads, master and production promote

trinadhthatakula:
## What & why

Release notes authored on `dev` for an already-stable version, then merged to `master`, still
triggered the dev build. Yesterday's mitigation — check whether `versionCode` ends in `0` — is
arithmetic standing in for intent, and it fails whenever the digit and the intent disagree.

This replaces it with **three branches, three rungs, one Play upload**. A rung is identified by the
branch it runs on; there is no arithmetic on the version number anywhere in the routing.

| merge | workflow | Play | GitHub | Telegram |
|---|---|---|---|---|
| `<feature>` → `dev` | `1-dev-publish.yml` | **uploads** to `alpha` | pre-release `v<name>-dev-<run>` | yes |
| `dev` → `master` | `2-master-promote.yml` | promotes `alpha` → `beta` | pre-release `v<name>-beta-<run>` | no |
| `master` → `production` | `3-production-promote.yml` | promotes `beta` → `production` | **release** `v<name>` | yes |

Play allows one upload per version code **per app, not per track** — which is what produced
`Version code NNNN has already been used`. Exactly one rung uploads; the other two move that upload
up a track. The invariant is enforced by three independent barriers, not just by convention:

- `fastlane/Fastfile:78` appends `bundleStoreRelease` only when `upload_to_store`, so a promote lane
  never builds an AAB and `GRADLE_AAB_OUTPUT_PATH` is never populated for `upload_to_play_store` to
  back-fill from;
- `Fastfile:196-197` sets **both** `skip_upload_apk` and `skip_upload_aab` (the latter defaults
  `false`, so a promote lane needs both);
- `fastlane/lib/thor_release.rb:19` limits `UPLOAD_TRACKS` to `internal`/`alpha` and raises otherwise.

`upload_to_store: true` occurs exactly once in the repo, and only `1-dev-publish.yml` names the
`distribute_dev` lane.

**There is no hotfix bypass, deliberately.** A mid-cycle fix lands on `dev` with a new version code
and re-enters at the bottom rung, superseding the previous candidate on `alpha`. A fast lane
straight to `beta` would be a second uploader — the exact problem the ladder removes.

Every rung still **builds** its APKs from its own commit; only the Play artifact is promoted rather
than rebuilt. Consequence worth knowing: the three GitHub releases for one version carry different
APK bytes, because AGP embeds the build commit in `META-INF/version-control-info.textproto`. That is
expected, not a signing or reproducibility defect — IzzyOnDroid rebuilds at the tag, and each tag is
now pinned to its own commit via `target_commitish`.

Implements Phase 1 + Phase 2 (Tasks 1–13) of
`docs/superpowers/plans/2026-08-06-release-ladder-and-foss-distribution.md`. Phase 3 (the FOSS store
fronts) is blocked by design until a `v<name>` release minted by `3-production-promote.yml` exists,
since every store front resolves `/releases/latest/` or clones at a tag.

## Changes

**Routing**
- `release-rung.yml` — one reusable `workflow_call` implementation; `1-dev-publish.yml`,
  `2-master-promote.yml`, `3-production-promote.yml` are thin callers differing only in inputs.
  Deletes `dev-check.yml` and `production-deploy.yml`.
- `fastlane/lib/thor_release.rb` — the pure logic (track routing, tag/name shape, guard rails),
  extracted so it is testable without invoking Fastlane. Adds `promote_beta` / `promote_production`.
- `detect-version-bump.sh` — version-bump detection lifted out of inline YAML and given tests.

**Gates that fail before anything is published**
- `check-notes-budget.sh` — pre-flight size gate. The Telegram cap is **1024 UTF-16 units and
  Telegram rejects rather than truncates**, so this counts UTF-16 units, not bytes (one emoji = 1
  character, 2 units, 4 bytes). `require_notes` now requires the actual **files**, via a repeatable
  `--require <file>`, not merely the directory — the previous form let a production run fall through
  to a raw `git log` body and broadcast the generic filler bullet, and **a Telegram send is not
  retractable**. The required set is per rung: beta consumes neither `telegram.md` nor
  `playstore.txt`.
- The caption wrapper is now **measured per run** from the real ref/actor/prefix/status rather than
  assumed. The old constant `160` was a one-character margin: the header contains the branch name,
  so 152 is the max across `dev`/`master`/`production` only — a dispatch from this branch measures
  176. 160 is now a floor.
- `actionlint` + `shellcheck` + a shell test runner in CI.

**Shizu store**
- `check-shizu-manifest.sh` / `sync-shizu-changelog.sh` now derive the version from
  `origin/production`, not the working tree — under the ladder the working tree is ahead of what is
  live. The resolver block is byte-identical in both scripts *on purpose*; that duplication is the
  lockstep contract, and a test asserts it.
- Warns on PRs, fails where the drift is actionable.

**Docs**
- `release-notes/README.md` rewritten — it is the file a human opens before cutting a release, so it
  ships in the same phase that changes the routing. Now also documents the **back-merge**, which no
  doc has ever covered.
- Retires `two-branches-one-play-version-code.md` (cannot occur with one uploader) and
  `telegram-caption-length-guard.md` (now implemented); inbound links checked first.
- `docs/fdroid-submission.md` corrected against a build with no `-foss` suffix — it contained a
  paste-ready recipe that would have produced a versionName mismatch.

**Build**
- Drops `versionNameSuffix = "-foss"`. Obtainium cannot reconcile `1.94.0-foss` against tag
  `v1.94.0` — `foss` is not in its recognised suffix list — so it sets `versionDetection = false`
  and marks outdated users as up to date. F-Droid's `Binaries:` substitutes only `%v`, which would
  resolve to a tag that never exists.
- Fixes `hi-IN` getting a blank what's-new every release (it had no changelogs directory).

## Testing

- [x] Compiles locally on JDK 21 — `./gradlew testFossDebugUnitTest --rerun-tasks` → **BUILD
  SUCCESSFUL**, 47 tasks executed. (Ran the unit-test task rather than `assembleFossDebug` because
  it exercises the same configuration and compilation; the only build-script change here is the
  `-foss` suffix removal. Only warning is the pre-existing Koin-plugin-vs-Kotlin-2.4.10 notice.)
- [ ] Tested under privilege mode(s): **N/A** — no runtime code changed. The sole `app/` change is
  one line of build config.

Suites, all green on this tree:

| suite | result |
|---|---|
| `.github/scripts/test/run-tests.sh` | 10 files, 0 failed |
| `fastlane/test/test_thor_release.rb` | 26 runs, 32 assertions, 0 failures |
| `vitest` (web repo-facts) | 16 files, 291 passed |
| `shellcheck -x -S style` / `actionlint` 1.7.12 / `ruby -c Fastfile` | exit 0 / exit 0 / Syntax OK |

Budget check at head: `check-notes-budget.sh 1.94.0` → telegram **1007/1024** (17 spare), playstore
483/500.

Both Shizu gates were proven to bite by mutation rather than by inspection: `/1000`→`/100` and
`origin/production`→`origin/master` each exit 1 with a real diff; deleting both blocks reports "a
LOCKSTEP block is empty"; an identical baseline passes.

## Checklist

- [x] Changes are scoped to the stated purpose
- [ ] Bumped `versionCode` — **not release-bound**; `versionCode` stays at `1940` and changes only in
  a `chore(release)` commit
- [x] Updated docs / release notes

## For the reviewer

Three things worth your call rather than mine:

1. **`workflow_dispatch` has no ref guard** on any of the three rungs. A manual dispatch from a
   feature branch would run the real `distribute_dev` lane and consume version code 1940 on Play.
   **This is pre-existing, not introduced here** — old `dev-check.yml:41` and
   `production-deploy.yml:37` carried the same bare trigger. Recorded as a follow-up candidate. If a
   rehearsal path is wanted, a `dry_run` input is the right shape, not a weaker notes gate.
2. **`CLAUDE.md` is gitignored** (`.gitignore:35`), so the plan's instruction to commit a "Release
   routing" section into it could not be honoured. That guidance lives in `release-notes/README.md`
   instead. Track `CLAUDE.md`, or leave it as is?
3. **No CI gate is a required ruleset status context** — the only required context is
   `build-and-test`, so every gate added here is advisory at the ruleset level. Pre-existing repo
   policy.

## Summary by CodeRabbit

* **New Features**
  * Introduced a three-stage release flow: development builds publish to Alpha, master builds promote to Beta, and production builds promote to Production.
  * Added automated version detection, release-note validation, Telegram caption budgeting, pinned release tags, and curated Play Store changelogs.
* **Updates**
  * Removed the `-foss` version-name suffix and added Hindi release notes.
  * Improved manual Telegram release publishing and APK distribution.
* **Bug Fixes**
  * Strengthened validation for versions, changelogs, assets, release configuration, and workflow behavior.


trinadhthatakula:
## Code review pass — 15 findings, 14 fixed, 1 partially declined

Reviewed the full diff plus the surrounding release machinery. Fixes are in `9597733b`, `b761e047`, `112a215e`, `0345c3b5`. Grouped by what each defect actually let through.

### Blocking — would have broken the next release

**1. The next release PR would have gone red.** `release-notes/README.md` step 5 told you to write `en-US/changelogs/<code>.txt` and stop there. But `supply` enumerates locales from the metadata directory and emits an **empty** `LocalizedText` for any locale missing a file for that version code — so a partial copy doesn't leave the previous notes in place, it *blanks* them. `test-changelog-locale-parity.sh` runs on every PR and would have failed with `MISSING: hi-IN/changelogs/<code>.txt`. Step 5 now carries the loop and the reason.

**2. `workflow_dispatch` with a blank `version_code` built v0.0.0.** `""` is truthy in Ruby, so `if version_code` took the manual-override branch, where `"".to_i` is `0`. The input's own description says *"leave empty to use the latest version"*. The lane invocation is now conditional, and a non-positive or non-numeric override is a hard error — `0` is exactly what a typo produces.

**3. `--fail-with-body` had been silently reverted to `--fail`.** It differs from `--fail` in one respect only: it writes the error body to stdout. Both Telegram calls redirected stdout to `/dev/null`. Replaced with a `send()` function that captures and echoes the body, plus `--connect-timeout 15 --max-time 300`. Left a ⚠️ in `check-notes-budget.sh`'s header — re-adding `> /dev/null` is all it takes to recreate this.

**4. An unset secret passed the emptiness check.** `echo "$UNSET" > file` writes a newline, so `[ -s file ]` is true. The keystore and the Play service account are now validated by content — `jq -e 'has("client_email")'` for the JSON — not by size.

### Tests that passed while proving nothing

**5. `test-runner-selfcheck.sh`** — nuance worth recording: CodeRabbit's stated failure mode (*"if discovery regresses to zero matches, the runner exits successfully"*) is **false** — `run-tests.sh:27-30` already exits 1. But the finding was right in substance: the test checked only the executable bit while its header claimed to prove discovery. Rather than refute it, I made the test earn the header. It now copies the real runner into `mktemp -d` fixtures and asserts 7 properties, including the two that matter — discovery is **not** recursive, and discovering zero tests fails the run.

**6. `test-release-tag-pinning.sh` accepted any non-empty `target_commitish`.** `target_commitish: master` is present, non-empty, and *exactly* the bug this test exists to catch — on the master rung it would even look right. Now requires `${{ github.sha }}` verbatim.

**7. `test-release-assets.sh` was a one-sided gate.** "No `.aab`" is satisfied perfectly by a step that publishes nothing, or only a checksum file. Added an explicit zero-APK check. I did **not** adopt the suggested `if line and not line.endswith(".apk")` — that flags legitimate `mapping.txt` and checksum assets.

**8. `test-caption-wrapper.sh`** now parses `release-rung.yml`'s own comment prose for the wrapper size it cites, re-measures with both actors, and asserts the figure is producible *and* exceeds the per-rung table's maximum. A comment drifting from the code it explains is how the 160-vs-176 confusion started.

### Supply chain

**9. Removed the release job's dependency caches.** GitHub Actions cache restoration is a documented poisoning vector, and this is the only job holding the keystore, the Play service account and `contents: write`. Every other input it takes is pinned by sha or checksum; the cache was the one that was not. A cold build is also closer to what IzzyOnDroid does. **Cost: ~2 min per rung, ×3 rungs — this is a one-line revert if you'd rather have the time back.**

**10. `persist-credentials: false`** on all three `pr-ci.yml` checkouts and on `shizu-store-audit.yml`. The Android build runs thousands of lines of third-party Gradle/R8 plugin code with the token sitting in `.git/config`; the audit job holds `issues: write`. Both authenticate through `GH_TOKEN` where they need to, and the shizu `git fetch --depth=1 origin production` needs no credential on a public repo.

**11. Dead input.** `telegram-release.yml`'s `increment` is read by no lane — grep-proven. Removed, and the header reframed: this is not a rung, it uploads nothing to Play, and it exists because `!cancelled()` lets a rejected broadcast leave a version live-but-unannounced.

### Partially declined

**12. Telegram `curl` retries.** Added the timeouts, deliberately **not** `--retry`. `sendDocument` is non-idempotent: a retry after a lost response double-posts an APK to the channel. A failed broadcast is recoverable by hand (that's what `telegram-release.yml` is for); a duplicate is not.

### Docs corrected where prose contradicted working code

**13.** Play's limit is **500 inclusive**. `check-notes-budget.sh` has always used `-gt 500`; two documents said "strictly under 500". The code was right.

**14.** The 160 wrapper default is a hand-run convenience, not a ceiling — ladder branches measure 145–152, this branch measures **176**. `release-rung.yml` measures per run and takes `max(measured, floor)`, so it was never wrong; the docs were.

**15.** `fdroid-submission.md`: what excludes the suffixed tags isn't the prefix — it's that the **whole** tag must be the version, and since fdroidserver applies the pattern with `re.match` (which anchors only the start), the trailing `$` is what makes the match whole.

Also: the consumer table's line-number references became step names. The ones that used to sit there were stale within two commits.

---

### New: `docs/branching-and-releases.md`

The contributor guide. Three permanent branches and why `production` must never be deleted, the ladder diagram, the contributor path, the three rungs and their routing, the one-uploader rule with its three independent enforcement points, why there is **no** hotfix bypass, the back-merge with its mid-cycle exception, where each channel's bytes come from, and why the three GitHub releases for one version carry different APK bytes. Linked from `CONTRIBUTING.md`.

### Verification

| Suite | Result |
|---|---|
| `.github/scripts/test/run-tests.sh` | 10 test files, **0 failed** |
| `ruby fastlane/test/test_thor_release.rb` | 26 runs, 32 assertions, **0 failures** |
| `actionlint` 1.7.7 (the CI pin, fetched fresh) | clean |
| `shellcheck` 0.11.0 (the new CI pin) | clean |

Both tools are now pinned with checksums in `pr-ci.yml` — 1.7.7 and 1.7.12 disagree about SC2153, and shellcheck renamed SC2317 to SC2329 in 0.10, so a newer local build is a green run that proves nothing about the job.

### For the owner, not decided here

- **The cache removal** (#9) — ~2 min/rung. Security-conservative call, trivially reverted.
- **Whether `telegram-release.yml` should exist at all.** I kept and fixed it, on the argument that it's the recovery path for the exact failure the docs warn about. Deleting it is defensible.
- **Sign-off on the four Hindi changelogs** (`112a215e`). The locale-parity gate checks that a file exists, not that it's translated — all four were English placeholders.

trinadhthatakula:
### ⚠️ The red checks are a GitHub outage, not this branch

Every failing job died in **`Set up job`** with:

```
Failed to resolve action download info. Error: Service Unavailable
```

That is before checkout — no repository code ran at all. Three unrelated workflows (PR CI, CodeQL, PR Labeler) failed in the same 15:44–15:49 UTC window, and `label` went green on a bare re-run with no code change.

Confirmed against GitHub's own status API: incident [`qcvjkzcs7j74`](https://stspg.io/rcz3fcm83sff) — *"Incident with Actions"*, **major** impact, opened 15:22 UTC, still `investigating`. Their words: *"Some workflow runs are failing to start or failing partway through."*

The last full green run on this branch was `341e318e`. I'll keep re-running until the incident clears and merge once the checks pass on their own. No changes needed.

Issue #178, opened 2026-06-01
App tagging / list organization feature


xvrpop:
First off, amazing app. Love the features and UI. I wanted to suggest the addition of app tagging feature, so users can browse and apps by tags. When you have many app its hard to remember the names.
Thank you!

trinadhthatakula:
Thanks for this, and sorry for the long silence.

App tagging hasn't been built. It's on the follow-up list and will get picked up alongside the next round of app-list UX work rather than on its own, since it needs the list, the filter sheet and the app info sheet to move together.

Two things landed in v1.93.1 that overlap slightly, in case either is closer to what you need:

- **Permission filter chips** on the Apps tab — narrow the list to everything that can reach Camera, Microphone, Location and so on. A new filter axis, but a fixed one, not your own labels.
- **Freeze Profiles** on the Freezer tab — named groups of apps you can freeze or unfreeze in one tap.

I'd stop short of calling profiles a tagging workaround, though. They only exist on the Freezer tab, their only actions are freeze/unfreeze, the Apps list and app info sheet know nothing about them, and the editor pops a freeze-risk warning when you tick a system app — which is pure noise if you're only trying to label something. There's also an open bug where the editor closes before the save is confirmed, so you'd lose a large selection if the name collided with an existing profile.

One question that would help me build the right thing: is what you want **tags for browsing** (attach labels to apps, then filter the list by label), or something narrower like **custom grouping or sorting** of the app list? Those are quite different builds and I'd rather do the one you'd actually use.

If you want to try the two above: **[Dev Build v1.93.1 (103)](https://github.com/trinadhthatakula/Thor/releases/tag/v1.93.1-dev-103)**.

trinadhthatakula:
Update after v1.96.0: app tagging is still open and tracked in the [follow-up backlog, item 24](https://github.com/trinadhthatakula/Thor/blob/v1.96.0/docs/follow-ups/README.md#band-c--real-and-expensive) and the [current community assessment](https://github.com/trinadhthatakula/Thor/blob/v1.96.0/docs/analysis/community-feature-request-assessment-2026-09.md#live-github-tracker-snapshot).

The intended direction is user-defined tags for browsing/filtering, together with per-app notes. This remains unimplemented; it has no scheduled release. The earlier “zero demand” rationale is outdated, because additional users have requested both tags and notes.

[...]

PR #364, opened 2026-08-07
docs: rank the whole backlog against what users actually asked for

trinadhthatakula:
## What this does

Twenty-three feature requests came in from r/howtomen. This merges them with the existing backlog
and re-sorts **everything open** by impact over effort, in one place.

The ranking lives in `docs/follow-ups/README.md` under **Do next — every open item, ranked**:
four bands, 32 numbered rows plus an unrankable Band D. Read the bands, not the exact numbers —
the gap between 3 and 5 is noise, the gap between band A and band C is not.

The evidence sits in a new `docs/follow-ups/reddit-howtomen-feedback.md`. Every one of the 23 asks
was checked against `origin/dev` before it was sized, and the file records what the code actually
does next to what each person reported.

## What the verification changed

**Three findings**, none of which were visible from the requests alone:

1. **A freeze can *remove* a system app for the user.** Where the platform refuses
   `pm disable-user`, the escalation runs `pm uninstall -k --user N`, clearing `FLAG_INSTALLED`.
   That explains *both* halves of one user's report — the lost Google accounts and the absence of
   the notification spam Hail's disable produces, since dependents cannot resolve GMS at all.
   ⚠️ **Inferred from the code, not measured.** The row says reproduce first, and reproduce it on a
   device where disable is actually refused — on stock Android the escalation never fires and the
   bug is invisible. The gate itself is correct; the fix is disclosure plus an opt-out, not removing
   the fallback.

2. **The debloat descriptions already ship.** `uad_lists.json` carries a per-package `description`,
   `UadHelper` parses it into `UadEntry`, and `UadSnapshot` exposes only `removal` — so it is read
   off disk on every load and thrown away. Both label asks are one accessor plus rendering.

3. **`PACKAGE_USAGE_STATS` is already declared and nothing reads through it**
   (`AndroidManifest.xml:42`). Found while pricing the abandoned-app request as the *most* expensive
   item in the thread. It is not. Either the line goes — IzzyOnDroid screens sensitive permissions
   and Play surfaces usage access on the listing — or the feature is much closer than it looked.
   This is a finding independent of the request either way.

**Two existing verdicts moved:**

- **#178 (app tagging)** was deferred on *"zero demand"*. Two users asked for it, one for tags and
  one for notes, which is the same Room migration. The cost grounds are untouched and may still
  justify deferring — but it now has to be deferred on cost.
- **#55b (process manager)** splits. *"Options for know app running in background"* is answered by a
  running/not-running flag; the fragile `dumpsys`/`top` parsing belongs to the live RAM/CPU half
  nobody asked for. The niche verdict was earned by the expensive half and then applied to the whole
  feature.

**One item is a docs problem wearing a feature request** — Freeze Profiles shipped in v1.93.1 and
does most of what "custom groups" asks for. It was requested anyway, by someone who had read the
Freezer screen carefully.

## No new defect

The one report that read as a bug — adding an app to the Freezer freezes it immediately, costing
unsaved work — is **intended behaviour**, confirmed by the maintainer. It stays on the backlog as an
expectation gap, which `freezer-membership-toggle-semantics.md` had already filed from the opposite
direction.

## Not displaced

The unguarded DataStore writes, the biometric capability check and cross-privilege suspend ownership
are all still ahead of every item in this thread on **risk**, whatever their demand.

## Scope

Documentation only — three files, no code touched.

## Summary by CodeRabbit

* **Documentation**
  * Updated the feature roadmap with clearer prioritization, sequencing, review status, and scope for upcoming improvements.
  * Added a ranked “Do next” section and incorporated recent feedback into follow-up tracking.
  * Documented verified community feedback, including package removal risks, visibility gaps, usage-statistics support, and process-management expectations.
  * Clarified which requests are already supported, partially supported, or require future work.


trinadhthatakula:
## Review findings addressed — one of them was a real error on my side

Thanks @coderabbitai; the `UsageAccessManager` catch was the important one.

### 1. `PACKAGE_USAGE_STATS` — my finding 3 was wrong, and dangerously so

I claimed the permission was vestigial and should probably be deleted, on the strength of `UsageStatsManager` matching nothing under `app/src/main/java`.

It matches nothing because **it is the wrong symbol.** The consumer is `StorageStatsManager`: the manifest line is what makes the `GET_USAGE_STATS` app-op grantable, and that op is what lets `queryStatsForPackage` answer for *other* packages — i.e. it is why Thor can show an app's size at all (`StorageStatsHelper.kt:47`). Acting on my finding would have broken app-size reporting.

And more is built than the deep link you flagged. `UsageAccessManager` is a complete gate:

| Piece | Where |
|---|---|
| Op read for Thor's own uid, per SDK level | `isGranted()`, `:37` |
| Silent grant via the privilege gateway, **then re-verify** | `tryGrantViaPrivilege()`, `:48` |
| Once-per-process auto-grant, latches only on *success* | `maybeAutoGrant()`, `:69` → `AppListViewModel.kt:375` |
| `ACTION_USAGE_ACCESS_SETTINGS` fallback | `usageAccessIntent()`, `:75` → `AppListScreen.kt:372`, `SettingsScreen.kt:424` |

The conclusion it fed **survives and gets stronger** — abandoned-app detection needs only `UsageStatsManager.queryUsageStats` (a different class, and the one thing here that genuinely is absent), a schedule and a notification. So finding 3 is **rewritten with the error stated in place** rather than quietly dropped, and band A's "settle the unused declaration" row is **deleted** — there is nothing to settle.

### 2. #55b — correct, the flag was ranked without an API behind it

I called the running/not-running flag affordable and said Dhizuku's missing shell only degrades it. Neither is supported. `getRunningAppProcesses` has returned only the caller's own process since Android 8, so even the coarse answer goes through a privileged `ps`/`dumpsys` — **Dhizuku is in the same dead-end it is in for the stats.**

Moved from **band B to band D** (unrankable) in both files, with what it needs before it can be ranked: a named API, a stated privilege path, a decided fallback. "Smaller than #55b as filed" is not "costed".

### 3. #55a status — reconciled, and the README was the correct one

PR #295 **merged 2026-07-30**, shipped in v1.93.1. The roadmap said "built, in review" in **four** places; all four corrected.

### 4. Roadmap Purpose conflict — fixed

Purpose no longer claims to drive the build order; it now describes itself as the per-issue analysis and hands ordering to `follow-ups/README.md`. Same for the "greens first" build-strategy line.

### 5. "zero demand" wording — marked historical

Both quotes now say explicitly that this sweep replaced them, so they describe the state *before* it.

### 6. "two findings" → "three"

Fixed in the heading and the intro paragraph.

### 7. "re-verdicted" → "reassessed"

Taken, both occurrences.

---

Also renumbered the bands to 1–30 after the deletions and re-checked **every** `band X #n` cross-reference against the row it actually points at — six were stale from an earlier renumber. Relative links re-verified: all resolve.

trinadhthatakula:
### Round 2 — `getRunningAppProcesses` attribution

Correct, and I had the wrong API. "Caller-only since Android 8" describes **`getRunningServices`** (deprecated API 26, own services only). `getRunningAppProcesses` has been progressively restricted since **Android 5.1** and is privilege-filtered for an ordinary caller.

Fixed at all three sites you named, plus a fourth: the roadmap's own `#55b` risk bullet carried the same "post-Android-8 returns only self" claim and **predates this branch**. Leaving it would have left the one remaining copy of a fact the other three no longer assert, so it went in the same pass.

The conclusion is untouched — coarse flag still needs a privileged `ps`/`dumpsys`, Dhizuku is still a dead-end, band D, unranked, name the API and the fallback before it moves.

Two more from the same pass:

- The roadmap's `**Purpose:**` line was a **lazy continuation** of the blockquote above it — no blank line — so it rendered *inside* the quote and restated it. Separated and deduplicated.
- MD018 at `reddit-howtomen-feedback.md:180`: a paragraph had wrapped so `#55b` began a line. Reflowed.

PR #368, opened 2026-08-08
docs(web): stop telling users Dhizuku destroys system-app data

trinadhthatakula:
## What

Six places on the public site still describe the **pre-#332** Dhizuku freeze path. Three of them tell the reader, in as many words, not to use Dhizuku for preinstalled apps because it will destroy their data.

That has not been true since PR #332 merged.

## Why it's wrong

`DhizukuSystemGateway.freezeSystemApp` on `master` today:

1. short-circuits if the package is already frozen,
2. tries `reflector.setAppEnabledDetailed(pkg, false)` — the **disable** rung,
3. reaches the uninstall rung only if `uninstallFreezeFallbackAllowed(isSystem = true, privilegeMode = DHIZUKU, disableRefusedByPolicy = ...)` returns true — i.e. only where the platform *actually refused*, not merely where the first command failed,
4. and when it does, calls `reflector.freezeSystemAppForUser(pkg)`, which is the `pm uninstall -k --user N` path — `-k` is `DELETE_KEEP_DATA`.

Shizuku and Dhizuku have had **identical** data semantics since that merge. The one asymmetry the gate still encodes is against `ROOT`, which returns `false` on every release and so never reaches the rung at all — the site already says that, correctly.

## Changed

| File | What it said |
|---|---|
| `features.mdx` | prose: fallback was "under Shizuku" only; a `<Callout>` said Dhizuku "is the exception on both counts"; recovery table said "Dhizuku has not been converted yet" |
| `faq.mdx` | a whole `<Callout tone="warning" title="A note on Dhizuku">` telling people to avoid it |
| `privacy.mdx` | "On the Shizuku path only"; a `<Callout tone="warning" title="Dhizuku is the exception">` |
| `download.mdx` | "Dhizuku is the exception: … does not keep its data" |
| `index.astro` | "Shizuku mode removes it"; "Dhizuku is the exception, and the one to avoid for preinstalled apps" |
| `claims.mjs` | the **C1 rule's own rationale** asserted the same falsehood — the gate meant to catch this class of error had it written into its justification. Its `source` list also named `DhizukuSystemGateway.setAppDisabled`, the wrong symbol; the gate is consulted in `freezeSystemApp`. |

## Deliberately not changed

Copy describing the uninstall fallback **itself** — that it exists, that it clears `FLAG_INSTALLED`, the API 37 shell refusal. All of that is true on `master` today and is what v1.94.0 ships. Band A removes the fallback for non-root modes on `dev`; the copy for *that* belongs in the `dev` PR, where it publishes at the release merge — exactly when it becomes true. Rewriting it here would make the site false about the app people are currently running.

## Verification

All gates green locally against the built `dist/`:

```
check-links:       OK — 8 pages, 183 internal links, 63 fragments
check-claims:      OK — 16 rules evaluated, 8 pages scanned
check-markup:      OK — 8 pages, 78 custom properties resolved
check-sitemap:     OK — 2 sitemap files, 7 URLs cross-checked
check-screenshots: OK — 8 pages, 0 placeholders
check-a11y:        OK — 8 pages, 38 axe rules, 0 failing nodes
vitest:            16 files, 291 tests passed
```

Targets `master` rather than `dev` because it is a correction to copy that is **live now** and web-only — no app code, no `versionCode`.

## Summary by CodeRabbit

* **Documentation**
  * Updated freezing guidance to clarify that Shizuku and Dhizuku preserve app data when Android requires removing a preinstalled app for the current user.
  * Removed outdated warnings claiming Dhizuku always deletes system apps and their data.
  * Clarified that Dhizuku follows the same conditional fallback behavior as Root.

### DIFF
--- a/web/src/content/claims.mjs
+++ b/web/src/content/claims.mjs
@@ -100,7 +100,8 @@ export const claimRules = [
       'page, the correction has to be too.',
     source:
       'RootSystemGateway.freezeSystemApp (rung 1 `pm disable --user`, "app data preserved"); ' +
-      'ShizukuSystemGateway.freezeSystemApp (rung 3 `pm uninstall -k`)',
+      'ShizukuSystemGateway.freezeSystemApp (rung 3 `pm uninstall -k`); ' +
+      'DhizukuSystemGateway.freezeSystemApp (same chain since PR #332, same gate, same -k)',
     allow: [],
   },
 
--- a/web/src/content/claims.mjs
+++ b/web/src/content/claims.mjs
@@ -69,11 +69,12 @@ export const claimRules = [
       'does not do is preserve the whole PackageUserState. The fallback still clears FLAG_INSTALLED, so the app ' +
       'reads as uninstalled-for-this-user to anything that does not pass ' +
       'MATCH_UNINSTALLED_PACKAGES — that is worth saying, and it is not the same as losing data. ' +
-      'Dhizuku is the exception: DhizukuSystemGateway still removes the package for the user ' +
-      'unconditionally and without -k.',
+      'Dhizuku was the exception until PR #332; it now routes through the same disable chain and ' +
+      'consults the same gate as Shizuku, so it too reaches the rung only where the platform ' +
+      'actually refused, and carries -k when it does.',
     source:
       'RootSystemGateway.freezeSystemApp; ShizukuSystemGateway.freezeSystemApp; ' +
-      'FreezePolicy.uninstallFreezeFallbackAllowed; DhizukuSystemGateway.setAppDisabled',
+      'FreezePolicy.uninstallFreezeFallbackAllowed; DhizukuSystemGateway.freezeSystemApp',
     allow: [],
   },
 
--- a/web/src/pages/download.mdx
+++ b/web/src/pages/download.mdx
@@ -92,10 +92,9 @@ This is not a switching-channels problem. It applies to **any** uninstall — a 
 reset, or "I'll put it back later."
 
 **Frozen apps stay frozen.** Removing Thor undoes neither freeze mechanic. Thor switches the package
-off in place; on the devices that refuse to allow that, the Shizuku path removes it for your Android
-user while keeping its data. Dhizuku is the exception: it removes a preinstalled package for your
-user every time, refusal or not, and does not keep its data. Android has no reason to undo any of
-that just because Thor is gone.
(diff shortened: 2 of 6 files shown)


trinadhthatakula:
## Code review — both findings applied (`5e4618f4`)

### 1. The API 37 callouts named the wrong scope — **confirmed, fixed**

The review cited `DhizukuHelper.kt:458-478`, which exists on no ref. The content is real, at
`data/source/local/dhizuku/Dhizuku.kt:457-478` — the file is `Dhizuku.kt`, the object inside it
is `DhizukuHelper`. Verified against `origin/master`, this PR's base, not a working tree.

The substance holds, and the copy was wrong in two directions at once:

- **Too narrow.** The guard is `Binder.getCallingUid() == Process.ROOT_UID`
  (`PackageManagerShellCommand.java:2281-2293`, android17-release, per
  `Shizuku.kt:157-178`). It refuses *every* caller that is not uid 0. "The shell user is refused"
  describes a shell-specific rule that does not exist.
- **Wrong audience.** `DhizukuAPI.newProcess` spawns `pm` inside the device-owner app — not shell,
  not root — so Dhizuku meets the same guard. That is measured, not inferred: Android 17,
  Dhizuku as device owner, `com.android.egg`, `Failure [only root can delete system app for a
  particular user]`, package left at `installed=true enabled=0` with `ceDataInode` unchanged.

Naming only Shizuku told a Dhizuku user the case could not reach them. Both callouts now say
"not uid 0" and name both modes as measured.

One nuance the fix deliberately does **not** claim: on that device rung 1 (`pm disable-user
--user 0`) exited 0 at shell uid while being refused for the device-owner uid, so Dhizuku reaches
the fallback more readily than Shizuku does. That is one package on one build, and it would need
its own measurement before it belongs on a user-facing page.

### 2. C1R `source` omitted Dhizuku — **confirmed, fixed**

C1's rationale already turns on Dhizuku joining the shared chain in #332, while the `source` line
beside it cited only `RootSystemGateway` and `ShizukuSystemGateway`. Added
`DhizukuSystemGateway.freezeSystemApp`.

### Gates

All eight (`check:types`, `astro build`, `check:links`, `check:claims`, `check:markup`,
`check:sitemap`, `check:screenshots`, `check:a11y`) plus 291 vitest tests, green locally. C10
(no marketing name beside "API 37") was the rule most at risk from this wording and passes — the
new copy states the API level and no name.

Issue #161, opened 2026-04-18
.apks native installation


nightznero:
Helloz,

I have updated Thor to newest version - 1.80.6 and after changelogs i thought it gonna install apks directly but i guess misunderstood that... so here is my question/request;
It is possible for Thor to install .apks directly from clicking them? Without opening Thor and looking for specific file?
Here could be option to install with Thor, InstallerX Revived
![Screenshot_20260418_083441_Settings.png](https://github.com/user-attachments/assets/0aa0ade3-ab21-46a2-b37d-2458e43c3332)

![Screenshot_20260418_083300_Android System(1).png](https://github.com/user-attachments/assets/b0356e2a-a757-4155-93e1-4220f8dcfc4b)

trinadhthatakula:
got it, you want thor to be the default app which can open apks files directly right? wait a day and i will add it to the app

trinadhthatakula:
Triage (current v1.90.0): this is already implemented. `PortableInstallerActivity` now registers ACTION_VIEW intent-filters plus MIME-type and pathPattern matching for `.apk`/`.apks`/`.xapk`/`.apkm`, so tapping such a file surfaces Thor in the system "open with" chooser and installs it (added after the 1.80.6 you were on). Closing as implemented. Note: robust parsing of every multi-APK bundle is a separate track (the `.xapk` parse fix #159 landed via #211).

nightznero:
> Triage (current v1.90.0): this is already implemented. `PortableInstallerActivity` now registers ACTION_VIEW intent-filters plus MIME-type and pathPattern matching for `.apk`/`.apks`/`.xapk`/`.apkm`, so tapping such a file surfaces Thor in the system "open with" chooser and installs it (added after the 1.80.6 you were on). Closing as implemented. Note: robust parsing of every multi-APK bundle is a separate track (the `.xapk` parse fix #159 landed via #211).

![Screenshot_20260717_231346_Android System(1).png](https://github.com/user-attachments/assets/b810e589-6a5d-4c1e-9d3a-04c497b4a989)
i see You have closed my issue but as You can see, this feature is not implemented yet. v1.92

trinadhthatakula:
hi @nightznero can u share some details with me, like the apks file u used, your device version and the file explorer u used

nightznero:
> hi [@nightznero](https://github.com/nightznero) can u share some details with me, like the apks file u used, your device version and the file explorer u used

[...]

PR #369, opened 2026-08-08
fix(prefs): guard the DataStore write path, and retire the freeze-becomes-a-removal copy

trinadhthatakula:
Band B row 13, plus the docs row that ships with it. Four things, in the order they matter.

## 1. A settings toggle could take the process down

`PreferenceRepositoryImpl` had **29 `edit { }` blocks** across two DataStores — 28 write functions, 33 call sites in `presentation/`, and **zero** `try` blocks among them. `DataStore.edit` writes a temp file and renames it, so a full disk, a read-only filesystem, a storage migration or an SELinux denial throws `IOException`. Every caller is `viewModelScope.launch { … }`, and `viewModelScope` is a `SupervisorJob` with no `CoroutineExceptionHandler`, so that throw is not a dropped setting — it is a crash from tapping a switch.

PR #339 made an unreadable settings file survivable and deliberately stopped at the read. This is the other half.

One `guardedWrite` helper wraps all 29. It catches `IOException` and returns whether the write landed.

**Not a `CoroutineExceptionHandler`** — one line of diff instead of 29, but it silences every unrelated exception in `viewModelScope` for the life of the app and cannot tell an individual caller whether *its* write landed, which is the whole requirement below.

**`IOException` alone, not `IOException` + `CorruptionException`.** `androidx.datastore.core.CorruptionException extends IOException` (`javap`-verified), so naming both is redundant rather than safer. The narrowness matters in the other direction too: `CancellationException` is *not* an `IOException`, so structured concurrency survives without the explicit rethrow a `catch (Throwable)` would have needed — and a rethrow is a line someone later deletes.

**No retry.** The follow-up doc left this open as a product call. `DataStore.edit` already serialises writes and re-reads the current value inside the transform, so a second attempt re-runs the same write against the same full disk. The read path retries because a transient read can genuinely succeed next time; this one has nothing to wait for.

## 2. Two failures the user has to be told about specifically

26 setters swallow and latch a generic notice. Two return `Boolean`:

**`setBiometricLock`.** `SecurityViewModel` writes `false` from inside a `collect { }` when the device can no longer do biometrics. Unguarded, the throw kills that collector for the process and strands the user outside their own app lock. Guarded but silent is worse in a different way: the flow does not flip, and Thor says "App lock turned off" while the lock is still armed on disk. It now says the write failed instead. The `when` still falls open to `NotRequired`, so a failed disarm never locks anyone out.

**`setLanguage`.** `SettingsViewModel` writes the preference then calls `localeManager.applyLocale`. Today a failed write throws and `applyLocale` never runs — the two stay consistent *by accident*. Adding a guard that merely swallows would break that: the platform persists the locale independently of DataStore, so the app would be visibly in the new language while the store held the old one, and the next cold start would silently revert it. It now skips `applyLocale` when the write did not land.

That is what forced the `announce` flag. Without it a reporting setter both latches the generic notice and returns `false` for the caller's specific one, and the user gets two messages about one failure.

Three new strings × 5 locales.

## 3. The website promised behaviour band A removed

`FreezePolicy.uninstallFreezeFallbackAllowed` answers `false` on every branch, so a freeze is `pm disable --user` and nothing else and a refusal is reported as a failure. Six passages across five pages still described the old escalation as current.

They are rewritten, and they keep the history rather than deleting it — that sentence was true for two years and is still true in the release notes for the version most people are running.

New claims rule **C16** stops it coming back, since rewriting six passages does nothing about the seventh. Its `unless` window is the match plus the 60 characters before it, sized so *"earlier builds substituted `pm uninstall -k`"* passes and *"Thor runs `pm uninstall -k`"* fails. C1's rationale and C15's `source` moved with it; C15's own **fixtures** asserted the withdrawn mechanic in prose, and a `pass` fixture must produce no violation from *any* rule, so C16 turned C15 red the moment it landed.

One C16 pattern was dropped before it shipped: `… for your user instead of disabling …` ends in "instead of", which is itself in the `unless` list, so the exemption window always contained it and it could never have fired. The fixture meta-test catches a rule with no fixtures, not a pattern that cannot match.

## 4. #161's FAQ answer, and a share target Thor does not have

Band A shipped the opt-in *"Show Thor when opening any file"* switch and the site never mentioned it. It is now documented as a **last resort**, after "use a different file manager", with both caveats stated: the filter cannot be narrowed (Android consults path patterns only for URIs that name a host, and these have none), and **the diagnosis has not been confirmed on a Samsung device**. The `pm query-activities` diagnostic is still what decides this.

The features page opened with "you can share or open a package file", which the FAQ contradicts outright — Thor declares no `ACTION_SEND` target at all. Fixed.

## Also in here

- `b3175d21` cherry-picks PR #368 (already merged to `master`) so both branches carry byte-identical text and the release merge has nothing to resolve.
- The design spec, committed after the fact with its four as-built deltas recorded next to the reasoning they revise.
- `docs/follow-ups/README.md`: row 13 and both website rows marked done, the release-notes line called out as the piece genuinely still owed on the `master` merge, and the citations to four `docs/site-content/*.md` paths corrected — those files were never committed, so the instruction to "fix the spec and the page together" was unfollowable.

## Verification

- `testFossDebugUnitTest` — **747 passing**, counted from `build/test-results/**/*.xml`. Four new tests in `SecurityViewModelTest` cover: a failed disarm does not claim the lock came off, a failed disarm still opens the app, a failed write is reported once, and the two reporting setters do not *also* raise the generic notice.
- All eight web gates green (`check:types`, `astro build`, `check:links`, `check:claims`, `check:markup`, `check:sitemap`, `check:screenshots`, `check:a11y`) and **296** vitest tests.
- `versionCode` unchanged at 1940.

## Known gap, stated rather than solved

`SettingsViewModel` has no JVM test and does not get one here. `LocaleManager` is a concrete class taking a `Context` and `applyLocale` calls `getSystemService(Context.LOCALE_SERVICE)`, so the ViewModel cannot be constructed under `testFossDebugUnitTest` without extracting an interface — a refactor across DI and the Settings screen that would make a crash fix hard to review and hard to revert. No existing test constructs it either; this is a pre-existing seam gap, now filed.

## Summary by CodeRabbit

* **New Features**
  * Added notifications when settings, language, or biometric-lock changes cannot be saved.
  * Failed saves preserve previous values and may be retried on the next launch.
* **Bug Fixes**
  * Language changes apply only after successful saving.
  * Biometric-lock recovery confirms whether changes were saved.
  * Freezing no longer removes apps as a fallback; refused freezes leave apps unchanged.
* **Localization**
  * Added failure messages in Arabic, Chinese, English, French, and Spanish.
* **Documentation**
  * Updated freezing, installation, uninstall, and recovery guidance.
  * Added Samsung file-manager guidance and an optional file-opening setting.


trinadhthatakula:
## Code review — 12 findings raised, 11 confirmed, 1 refuted

Adversarial review of the whole PR head (`59874a5d`), each finding verified against the source before acting on it. Everything confirmed is fixed in `feaa6bd0`, `1d7b5a24` and `b8f75046`.

### The one that matters: the PR made a failure *quieter* than the crash it replaced

`SettingsViewModel.setBiometricLock` called the setter for effect and discarded its `Boolean`. The impl passes `announce = false`, so with the return value dropped a failed app-lock write from the Settings screen reached the user through **no channel at all** — no crash, no toast, and no store-wide notice either, because this setter deliberately suppresses it.

On `dev` that same tap was an unguarded `edit {}` that crashed the process, which is at least information. And it landed on the one preference whose interface KDoc says its failure *"must be heard in its own words"*, because "some settings could not be saved" does not tell anyone which way their front door is facing.

Fixed with two strings rather than one, since the direction is the point: a dropped `true` leaves the lock **off** after Thor appeared to arm it, a dropped `false` leaves it **on**. `biometric_lock_not_saved_still_off` / `..._still_on`, across all five locales. `biometric_lock_disable_not_saved` stays what it was — the auto-disarm's own message, which has something different to say.

### The latch replays a notice it already delivered

`settingsWriteFailed` is one-way and lives on the repository singleton; its only collector lives on a ViewModel. Exit finishes the activity without ending the process (`MainScreen` → `finish()`), so the next launch builds a fresh `SecurityViewModel`, reads a `true` it has already reported, and opens on a notice about nothing having failed. Config-change recreation was never affected — the ViewModel survives that.

`acknowledgeSettingsWriteFailure()` lowers it right after the send. That does not claim the disk recovered: the next dropped write raises it again, which is the distinction that makes clearing safe. The latch tracks an **unreported failure**, not a **broken store**. The `// one notice per process` comment was wrong and is corrected.

### The guard had no test of its own

The existing coverage drives `FakePreferenceRepository.write()` — a hand re-implementation of `guardedWrite` that can drift from it silently. So the tests proved the fake behaves like the fake.

`PreferenceWriteGuardTest` drives the real helper against a `DataStore<Preferences>` whose `updateData` throws: failure returns `false` instead of throwing, the latch is raised, `CorruptionException` is caught by the same clause, `announce = false` stays quiet, a non-IO failure is rethrown untouched, `CancellationException` propagates, and — the one pinning the design decision rather than a comment — **a failed write is attempted exactly once**.

`setLanguage` was also never exercised by any test, so half of "the two reporting setters do not also raise the generic notice" was unverified. It is now, at the repository level; `SettingsViewModel` itself still cannot be constructed on the JVM (`LocaleManager` is a concrete class taking a `Context`) and that seam gap stays filed.

### C16 forbade a claim that is true

Pattern 2 anchored on the subject (`Thor|Shizuku|Dhizuku|modes?|paths?`) rather than on freezing, so "Thor removes it for your Android user" tripped it — which is exactly what the Uninstall and Debloat paths do, truthfully, through `pm uninstall --user N`. Same over-reach on the switch-off-or-remove pattern. A gate that forbids a true sentence gets suppressed rather than obeyed, so both now require a freezing word: what band A closed is the *freeze* rung, not removal as such.

### Three copy corrections, all from reading the manifest instead of the previous copy

- **`features.mdx`: "the match is on the filename extension".** Only one of the three VIEW filters on `PortableInstallerActivity` is extension-gated. The other two have no host and no path and match on MIME alone, so a file manager declaring a type Thor claims matches with no filename anywhere in the URI.
- **`faq.mdx` #161: the wrong reason the filter cannot be narrowed.** It claimed Android consults path patterns only for URIs naming a host and that these have none. Content URIs *do* name an authority; the constraint is on the **filter** — `IntentFilter.matchData` reads its path list only inside the `mDataAuthorities != null` branch. The honest reason needs no platform detail at all: what the switch exists to catch is a handover carrying no name and no type, so there is nothing left in it to match on, and putting an extension pattern back would rebuild the filter that already misses these files.
- **`features.mdx` status chips still said "a system app that a freeze removed for your Android user"** — 180 lines after the callout saying that substitution was withdrawn. The chip behaviour is real and stays (`AppInfo` folds `FLAG_INSTALLED` into `enabled`), because a device an older build ran on can still be sitting in that state; the sentence now names the cause rather than implying Thor still does it.

### Housekeeping

`guardedWrite` had been inserted between `guardedRead`'s KDoc and its declaration, orphaning 24 lines of documentation onto the wrong function. Moved below it.

### Refuted

`settings_not_saved` saying the setting *"will go back to its old value next time you open the app"* was flagged as wrong on the grounds that nothing was written, so there is nothing to go back from. It reads correctly from where the user is standing: the switch on screen shows what they chose, disk holds the old value, and the next launch shows the old one. No change.

### Verification

- `:app:testFossDebugUnitTest --rerun-tasks` — **757 tests, 0 failures, 0 errors, 0 skipped** (65 result files).
- `npm run build` — types, astro build, 185 links + 65 fragments, **17 claim rules over 8 pages**, markup (78 custom properties), sitemap, screenshots: all OK.
- `npm run test` — **296 tests, 16 files**, all passing.
- `npm run check:a11y` — 8 pages, 38 axe rules, 0 failing nodes.

trinadhthatakula:
## CodeRabbit findings — all six triaged, five fixed in `c30e9ff0`

Each verified against source before acting on it. One was already fixed, four are now, and the last one turned out to be true of my own fix as well.

### `claims.mjs` C1 / C1R — "a freeze is `pm disable --user` and nothing else" ✅ fixed

Confirmed, and it was mine. Grepping the three gateways:

| gateway | first rung | second rung |
|---|---|---|
| `ShizukuSystemGateway.kt:96-97` | `IPackageManager.setApplicationEnabledSetting` (reflection) | `pm disable-user --user N` |
| `RootSystemGateway.kt:459,571` | `pm disable` | `setApplicationEnabledSetting` |
| `DhizukuSystemGateway.kt:93` | `pm disable-user --user N` | `IPackageManager` reflection |

Two of the three try reflection *first*, and none of them runs the command the rationale names. Band A row 1 did not make the mechanism uniform — it made the **outcome** uniform. Both rules now assert that: a freeze disables in place and never removes. A rule whose stated justification is falsifiable by `rg` is a rule the next person edits around.

### `claims.mjs` C16 — no passive pattern, and a bare `not` in `unless` ✅ both fixed

The passive gap is not hypothetical: *"a system app that a freeze removed for your Android user"* is the exact sentence this PR had to fix on the features page, and C16 — the rule written to catch that claim — read straight past it. Past tense, no subject, no object pronoun. Fifth pattern added.

Its first draft used pattern 2's 120-character window and immediately failed the build on `download/index.html`:

> If you have already uninstalled with apps still frozen: reinstall Thor, accept the Import Disabled Apps prompt, and use the app list, which shows packages uninstalled for your user too.

Two true clauses in one sentence with no claim between them. It is `[^.;:]{0,40}` now — what makes this *the claim* is the removal being predicated of the freeze, and that puts the two words adjacent. The narrower window is the more accurate rule, not just the quieter one.

`unless` is split in two rather than one flat list. A retraction marker stands alone (nothing writes "earlier builds" or "withdrawn" next to a removal by accident); a negation has to attach to the removal itself, within three words and without crossing punctuation. Probed:

```
VIOLATION p2   | A system app a freeze removed for your Android user is gone from the launcher…
VIOLATION p1   | Where the disable is not exposed, a Shizuku freeze removes it for your user…
EXEMPT p1      | A freeze does not remove it for your user in any privilege mode, and no…
no match       | Uninstalling a system app removes it for your Android user.
no match       | …already uninstalled with apps still frozen: … packages uninstalled for your user too.
```

Both requested fixtures are in `C16/fail` — the passive claim and the unrelated-negation claim. The pass fixture gains the correctly-negated sentence, which has to keep passing or the rule is unusable.

### Spec `:263` — `guardedWrite` called "private" ✅ fixed

`internal suspend fun DataStore<Preferences>.guardedWrite` (`PreferenceRepositoryImpl.kt:514`), matching `internal fun Flow<Preferences>.guardedRead` at `:473`. The visibility is load-bearing — `PreferenceWriteGuardTest` drives the real helper, not a copy of it.

### `docs/follow-ups/README.md:137` — two due dates for one retraction ✅ fixed

Real contradiction. The spec's version wins: the v1.94.0 line was **true when it shipped**, so it is superseded rather than erroneous, and `gh release view` reads the published body, which is a snapshot taken at run time — editing the repo file corrects nothing a reader sees while creating the appearance that it did. The index now says next release, in that release's own `github.md`.

### `SettingsViewModel.setBiometricLock` — already fixed

Duplicate of findings 1 and 4 in my own review above; landed in `feaa6bd0` with two strings rather than one, since which way the lock is facing is the whole point of the message.

### And one CodeRabbit did not raise, found while reading the one above

My own `features.mdx` fix from `1d7b5a24` said *"Nothing Thor does now puts an app in that state"*. That is false in the other direction: **Uninstall and Debloat still run `pm uninstall --user N`** on a system app, which is what they are for. Only the *freeze* rung stopped reaching that state. Corrected in both the page and the spec, which had the same sentence.

### Verification on `c30e9ff0`

- `npm run build` — types, astro, 185 links + 65 fragments, **17 claim rules over 8 pages**, markup, sitemap, screenshots: OK
- `npm run test` — 296 tests, 16 files
- `npm run check:a11y` — 8 pages, 38 axe rules, 0 failing nodes
- No Kotlin changed in this commit; the 757-test run on `b8f75046` stands.

PR #370, opened 2026-08-08
feat(freezer): Band B Freezer cluster — watchlist prune, save ordering, per-profile verbs, optional confirmation, result expiry

trinadhthatakula:
Band B's Freezer half — backlog rows **15, 16, 17, 18 and 22**, one commit each. Part C of
`docs/superpowers/specs/2026-08-08-band-b-design.md`.

**792 unit tests green** (762 → 792 across the branch), `lintFossDebug` clean.

Four review findings were fixed in `48b2e601` — see the [review
comment](https://github.com/trinadhthatakula/Thor/pull/370#issuecomment-5225716851) for the full
reasoning. In short: `BulkOutcome.NothingToDo` now carries a `NoOpReason`, because folding *no
privilege* and *no targets* into one value made Settings' Unfreeze-all tell a user with a full
watchlist that their Freezer was empty; the watchlist snapshot is taken **before** the package scan
it is judged against, closing a window where a row added mid-scan could be pruned; the profile
editor's save-success event names the editor that issued the write; and `refreshPinnedShortcutIcons`
is deleted with its last caller.

---

## Row 15 — watchlist rows whose package is gone (`80a0dc4f`)

A watchlist row names a package by string, so uninstalling the package leaves the row behind. It
renders nothing, cannot be restored, and cannot be removed by the one gesture that removes rows —
that gesture unfreezes first, and there is nothing left to unfreeze.

**This is option 3 of four in the linked doc; the "Remove anyway" options are still open.**

The prune keys off the app cache's **`ScanVerdict`**, not off the app list. That is the one thing the
doc's sketch did not say and the one thing that matters: downstream, a `GET_INSTALLED_APPS` collapse
on a Chinese ROM and a mass uninstall are indistinguishable, so pruning against the list would delete
the entire watchlist on the devices least able to recover it. `prunableWatchlistRows` carries **its
own emptiness guard on top of `Accept`**, because `scanVerdict`'s first rule accepts unconditionally
when nothing is cached — an empty scan against an empty cache arrives carrying `Accept`, which is a
verdict about the cache, taken on a scan that saw nothing.

`FreezerRepository.removeAll` deliberately has **no** restore step, unlike `remove`: its only caller
has already proved the package absent, so an unfreeze could only fail. Both KDocs say so, since the
asymmetry reads like an oversight.

## Row 16 — the profile editor closed before its save landed (`800b41e3`)

The editor was dismissed in the same statement that issued the write, so the two writes Room can
legitimately refuse — a name the unique index already holds, and a members-table foreign key —
reported themselves as a toast floating over a sheet that had already thrown the draft away. Both
refusals were already *named*; there was just nothing left on screen to act on the answer.

`runProfileWrite` now returns whether it landed, `saveProfile` emits `ProfileSaveSucceeded` only
then, and `FreezerScreen` closes on that event **and on nothing else**. A refused save leaves the
sheet up with the draft intact and Save as the retry — which is also why the in-flight flag clears in
`finally` rather than on success.

The write stays on `viewModelScope` rather than being awaited in the sheet: the editor's
`rememberCoroutineScope` dies with the composition, so a rotation mid-save would cancel
`updateProfile`'s `@Transaction` part-way.

## Row 17 — per-profile suspend and force stop (`b648e405`)

The Reddit ask was custom groups with per-group freeze/suspend/kill. Freeze Profiles shipped the
groups and two of the four verbs; this is the remainder. Both go in the row's overflow menu — a fifth
and sixth icon button would push the profile name into an ellipsis.

**Nothing is persisted on the profile.** The entity stays verb-agnostic because nothing else wants to
know a profile's verb: the launcher restore gate, `FreezerBridgeProvider` and `targetsFor` all key on
membership alone, and persisting one buys a single saved tap for a Room 6→7 migration the debug build
cannot exercise.

Suspend needs **no new `BulkOp`** — it is already `bulkActionFor`'s second branch. `BulkRequest`
gains an optional `FreezerMode` the runner consults only when present, so "Freeze" still means
whatever the user has said freezing means. That widens the coalescing key, which is the point: a
suspend-run and a disable-run of the same profile are different operations over the same packages.

**Kill stays out of `BulkOp`.** It does not change an app's frozen state, so putting it in the runner
forces an answer to *"does a kill cancel an in-flight freeze of the same profile?"* where the honest
answer is that they do not interact. It routes through `MultiAppAction.Kill`, which already has a
confirm dialog and a progress log. Targets are the installed, non-frozen members — a disabled package
has no components to start and a suspended one is already stopped — and the menu item is disabled
when that set is empty.

## Row 18 — optional per-freeze confirmation (`6e8fd63f`, `8c7a3cc7`)

Debloating a fresh device means answering the same system-freeze dialog dozens of times, and the last
answer carries no more information than the first. A Settings → Freezer switch skips it — for
**`FreezeTier.NORMAL` system apps only**.

The gate is one pure function, `freezeNeedsConfirmation`, in `FreezePolicy.kt` beside `freezeTierOf`.
Both freeze surfaces call it instead of retyping `shouldFreeze && appInfo.isSystem`, which is the
drift `AppRiskDialog` was created to end.

**EXPERT is deliberately outside the setting's reach.** Its warning is a verdict about that specific
package, and there is no backstop below the dialog — `FreezeAppUseCase` refuses `BLOCKED` and nothing
else — so a preference about tedium must not quietly become one about risk. **BLOCKED still raises
the dialog**, because rendering no confirm button is how it is refused. Band A #4's watchlist-add
dialog is untouched: `ManageFreezerSheet` raises its own, and this gate does not reach it.

The flag rides on `AppInfoDetailsUiState`, which both surfaces already reach, so no parameter threads
through `AppListScreen`, `FreezerScreen` or `MainScreen`. Collected in `init` rather than in
`loadAppDetails`, because the freeze action is live at the sheet's *partial* detent, before any
detail load has run.

`8c7a3cc7` translates the two new strings into all four locales — `MissingTranslation` is an error in
`:app`, so the switch shipped a red lint on its own.

## Row 22 — a parked bulk result had no expiry (`6a9e7e80`)

`BulkFreezeRunner` lives for the whole process, so a result parked for the QS tile subtitle waited
until the shade next opened — possibly the following morning. "Froze 12 apps" then reads as *just
now*, because that subtitle is the tile's live status line everywhere else.

`ParkedBulkResult` stamps the result with `SystemClock.elapsedRealtime()`, and `freshParkedResult`
decides whether it is still worth showing. Monotonic, not wall clock: wall clock can be set backwards
by the user or by NTP, which would make an hours-old report look fresh again. A negative age reads as
**stale**, so the fail-closed direction holds for any later caller that passes a clock which can go
backwards.

**The linked doc's two objections to this fix both dissolved, and its prerequisite did not apply.**
It expected an injected clock ("to stay testable") and the runner's concurrency tests first. The rule
is a pure function, so it is asserted without constructing the runner at all — a `Clock` port would
have been a fifth collaborator on a class that already has four untestable ones — and neither half of
this change widens what mutates shared state concurrently. `RESULT_TTL_MS` is five minutes and its
comment says the number is arbitrary.

**The check is read-side and there is no timer.** `paint()` already runs synchronously at the top of
every `onStartListening`, so every moment the staleness can be *observed* is a moment the check
already executes. `consumeResult` now takes the parked value rather than the bare `BulkResult`,
because two runs over the same watchlist routinely produce an equal result and the stamp is what
tells them apart.

`SettingsViewModel.unfreezeAll` was the last watchlist-wide path outside the runner, which is the
same finding's other half. Routing it there is close to a deletion — `bulkResultMessage` already
produced the two strings it hand-built — and it picks up the `Semaphore(5)` bound in place of an
unbounded `awaitAll`, the 30 s deadline, the completions-driven pinned-icon rebuild it had been
requesting by hand, and the `_lastResult` clear. It also stops **over-reporting**: it unfroze and
counted every watchlist row, including apps that were already running, where `targetsFor` counts only
the frozen ones.

Single-app and small-selection unfreezes are left alone — the runner's API is list-shaped and
watchlist-or-profile scoped, and they are neither. **No `origin` field on `BulkRequest`**: its
equality is the coalescing key, so widening it changes concurrency behaviour.

---

## Tests

30 new, all pure-JVM. `:app` has no mocking library by policy, so these are hand-written fakes.

- `FreezePolicyTest` — the five `freezeNeedsConfirmation` cases, including that BLOCKED still
  confirms and that the setting cannot reach EXPERT.
- `BulkFreezeTest` — the TTL boundary (half-open), the overnight case, the backwards-clock case, and
  that two identical results parked at different times are different values.
- `AppInfoDetailsViewModelTest` — the flag is known *before* any detail load, a flip reaches a sheet
  that is already open, and a later load does not undo it.
- `FreezerViewModelTest` — a refused profile save reports itself, never emits the success event, and
  clears the in-flight flag anyway; the success event names the editor that issued the write; and a
  profile run blocked by privilege names the privilege rather than the profile.

`SettingsViewModel` still has no unit test: `FreezerShortcutManager` is a concrete class over
`Context`, so it cannot be built on a JVM. Pre-existing and out of scope here.

**Two of the review fixes are only half-covered, deliberately.** The editor-session comparison
(`event.editorSession == editorSession`) is Compose state in `FreezerScreen`, so the test pins the
half a JVM can see — that the token survives the write untouched and comes back on the event. The
prune's read ordering needs a real `PackageManager` to exercise; the obligation is written into
`prunableWatchlistRows`' KDoc instead, where the next caller will look.

## Not device-tested

None of this has run on hardware. The highest-value manual checks are the QS tile subtitle after a
freeze (fresh, then again after five minutes), Unfreeze-all from Settings against a watchlist with a
mix of frozen and running apps, and a profile Suspend followed by a profile Freeze to confirm they do
not coalesce. Worth adding after the review fixes: revoke Shizuku and run Unfreeze-all — it should
now say **"Grant Root / Shizuku / Dhizuku first"** rather than "No apps in Freezer".

## Docs

`docs/follow-ups/README.md` rows 15/16/17/18/22 and the lower tier table, plus the five follow-up
docs. Three of them now record a call the implementation overturned — see `e13ab17b`. The band B
pointer at the head of the ranking now says to start at **#14**; rows 14, 19, 20 and 21 are the
Apps-tab remainder and are PR3.


trinadhthatakula:
## Review

Four findings, all accepted and fixed in `48b2e601`. Two are CodeRabbit's, one is from an
adversarial pass over the `_lastResult` lifecycle, and one is mine, from re-reading the `unfreezeAll`
reroute against what `NothingToDo` actually means.

### 1. `NothingToDo` cannot say *why*, and the reroute made that visible — `SettingsViewModel.kt:308`

`BulkFreezeRunner.run()` returns `null` in **two** unrelated situations:

```kotlin
if (!privilegeManager.state.first { it.isReady }.hasAnyPrivilege) return null
val targets = targetsFor(request)
if (targets.isEmpty()) return null
```

Both fold into `BulkOutcome.NothingToDo`, and my new `unfreezeAll` renders that as
`tile_no_apps_toast` — **"No apps in Freezer"**. So a user with twelve rows on the watchlist and a
revoked Shizuku binder is told their Freezer is empty, while looking at a screen listing twelve
apps. The old hand-rolled path reported `tile_unfreeze_partial_failure` there, which was at least
true. That is a regression I introduced in `6a9e7e80`.

The type being lossy is not new, and every consumer has been compensating for it in a comment:

- `FreezerLaunchActivity.kt:135` — *"Privilege was checked above, so this is the empty target list"*,
  which is only true because that file runs its own redundant privilege probe first.
- `FreezerViewModel.kt:459` — *"A no-op: no privilege, or nothing left to act on"*, and it picks a
  string vague enough to cover both.

Three call sites, three workarounds for one missing field. Fixed by giving the outcome a reason:

```kotlin
enum class NoOpReason { NO_PRIVILEGE, NO_TARGETS }
data class NothingToDo(val reason: NoOpReason) : BulkOutcome
```

The privilege await moves from `run()` up into the `async` body so the two causes are distinguished
at the point they are decided — same position in the sequence, still inside the runner, so the
"the runner must not depend on its caller having awaited anything" rule its KDoc states is intact.
No new strings: `tile_grant_privilege_toast` already exists in all four locales, and it is the one
message here that is *actionable*.

### 2. The watchlist is read after the package scan, so a row added in between can be pruned — `AppRepositoryImpl.kt:310` (CodeRabbit, major)

Agreed, and the consequence is exactly the failure mode row 15 exists to prevent. `currentPackageNames`
is fixed at the `getInstalledPackages` call; `pruneWatchlist` then reads `getAllPackageNames()`
afterwards. A row added between those two points names a package this scan never saw, so it lands in
`stale` and `removeAll` deletes it — and since adding to the Freezer freezes immediately, that leaves
a **frozen app with no Freezer row**, which is the state with no escape hatch.

It needs a newer scan to have surfaced the package (a fresh install) while this older scan is still
between its two reads, so it is narrow. It is also four lines to close: snapshot the watchlist
*before* the scan and prune only rows that predate it. Rows added later simply wait for the next
scan, which is the safe direction.

The read has to stay failure-isolated, though — the scan loop's own `catch` swallows and continues,
so a Room throw hoisted into it would skip `producer.send` and drop that scan's emission entirely.
It goes through a `watchlistSnapshot()` helper that returns `null` on failure, and the prune skips.

### 3. Save success closes whichever editor is open, not the one that saved — `FreezerScreen.kt:191` (CodeRabbit, minor)

Agreed. `if (editorProfileId != null) closeProfileEditor()` — the guard my own comment describes as
covering hand-dismissal is the thing that makes this reachable: dismiss a saving editor, open
another, and the first write's `ProfileSaveSucceeded` closes the second and takes its draft.

The profile id cannot stand in for the identity, which is worth stating because it is the obvious
fix and it does not work: two "new profile" editors both carry `NEW_PROFILE_ID`. So the event
carries an editor-session counter the screen mints on open and compares on close.

Note the blast radius is bounded by something already shipped — `profileSaveInFlight` is global, so
the second editor's Save is disabled for the whole window. Losing the draft is the only outcome, not
a double write.

### 4. The reroute orphaned a method whose KDoc still advertises the bypass — `FreezerShortcutManager.kt:164`

`refreshPinnedShortcutIcons()` has no callers left. Its KDoc says it exists "for a caller that runs
its own batch instead of going through `BulkFreezeRunner` — currently only Settings' Unfreeze-all",
and this PR is what rerouted that caller.

Dead code is the small half. The KDoc is the problem: it names the bypass as the supported way to
build a new bulk surface, at the exact moment there is no longer any such surface. `SettingsViewModel`
still holds the concrete `FreezerShortcutManager` for `syncDynamicShortcuts`, so it stays reachable
from one import-free edit away. Deleted rather than re-documented — the rebuild hangs off the
runner's `completions`, which is the entire reason that subscription exists.

### What was checked and found clean

Recorded so it does not get re-derived: the `_lastResult` clear on the UNFREEZE path has no
suspension point between `run()` returning and the assignment, so cancellation cannot preempt it;
`consumeResult`'s CAS is correct across a result replaced between the tile's read and the shade
closing, and correct when `shownResult` holds an expired value; `freezeNeedsConfirmation` has no
path that lets BLOCKED or EXPERT through, on either freeze surface, with `FreezeAppUseCase` as the
backstop below both; the prune cannot delete a row for an `UNINSTALL`-frozen system app, because the
scan carries `MATCH_UNINSTALLED_PACKAGES`; and no test in the branch passes on a reverted fix.

---

**792 tests** (789 → 792), `lintFossDebug` clean. What is still untested and why — the screen-side
session comparison is Compose state, the prune ordering needs a real `PackageManager` — is in the
commit body and in the PR description.

Issue #130, opened 2026-02-23
InstallWithOptions


nightznero:
Helloz,

2 things that should be added;
1. Thor doesn't recognize [InstallWithOptions](https://github.com/zacharee/InstallWithOptions) from @Zacharee
![Screenshot_20260223_090953_Thor(1).png](https://github.com/user-attachments/assets/6da5673e-c2f9-407f-92b5-a59847e64a82)
2. After clicking the name of installer it would be nice if it could shows what exactly apps was installed by. E.x - when click "RevivedX" it  could load list with installed app by RevivedX

Best Regards, NIGHTZNERO

trinadhthatakula:
Hi @nightznero 
I will go through install with options once and check if they are installing under different package name, and the second suggestion you gave is actually very good i will implement it soon, along with clickable piechart.

trinadhthatakula:
Update on both parts of this — neither has shipped.

The home screen is untouched in v1.93.1, so apps installed via InstallWithOptions still show as a raw, uppercased package id in the App Distribution legend, and the chart still isn't tappable. My "I will implement it soon" from February is still outstanding; sorry for the wait.

It is now properly tracked and sized in the follow-up list — the friendly-label part is small, the tap-through-to-a-filtered-list navigation is the bulk of the work — so it's queued rather than floating.

Two things worth knowing right now:

**There's already a workaround for the second half.** In the Apps tab, the filter sheet's **Installation Source** chips list every installer present on the device, and picking one shows only that installer's apps. The chip label falls back to the installer app's own name, so if InstallWithOptions is installed and was recorded as the installer, that chip may already read properly. What's missing is *reaching* that view by tapping the chart — the function exists, the shortcut doesn't.

**Reliable attribution is partly not fixable, and I'd rather flag it than overpromise.** Shell-based installs — which is how many of these go through — record `com.android.shell` or nothing at all as the installer. When that happens there's no package id left to make friendly; Android simply didn't record one. So what I can commit to is a proper label wherever an id *was* recorded, plus tap-through navigation from the chart. I can't commit to every InstallWithOptions install being recognised, because some of them are indistinguishable from any other shell install.

I'll comment here when it lands.

trinadhthatakula:
[...]

PR #371, opened 2026-08-08
feat: band B's Apps-tab cluster — Fix Store picker, scroll indicator, list export, installer names

trinadhthatakula:
Band B rows **14, 19, 20 and 21** — the Apps-tab half. Follows #369 (row 13) and #370 (rows 15–18, 22); with this the band is built. Spec: `docs/superpowers/specs/2026-08-08-band-b-design.md`, Part D.

`versionCode` stays **1940**. None of this has run on a device.

## What ships

**Row 14 — Fix Store: selection, cancel, plainer copy.** The scan hands over to a picker: every candidate listed with the installer Android currently records for it, everything pre-ticked, the count in the confirm button. The accident being fixed is *"I did not know what it would touch"*, not *"I did not mean to tap Confirm"*, so forcing forty taps would punish the case the feature exists for.

The target set was wrong in two ways and wrong in two places. It excluded Google's package installer but **not AOSP's** `com.android.packageinstaller`, so on a de-Googled ROM every normally-sideloaded app was a candidate — and the Home card's badge computed the same rule from its own, further-drifted copy. Both now call one pure `fixStoreCandidates`, which also drops **Thor itself**: all three gateways refuse to reinstall it by name, so it could only ever be a guaranteed failure occupying a slot. Thor's id is *passed in*, not read from a constant, because the debug build's `applicationIdSuffix` means the installed package is not the one the source spells.

`performLoggedMultiAction` gained a stop — checked **between** apps and never during one, since killing a `pm install` halfway is how a package ends up half-written. Every logged batch gets it; a single-app batch does not offer it, because by the time the button could be pressed the work is done.

**Row 19 — scroll position indicator.** `material3`'s `Modifier.nonInteractiveScrollbar` on the Apps tab and the Freezer's main list: position without a drag target, so it costs nothing in touch surface. Both screens hoist their lazy state *only* so the scrollbar can read it, and each branch still mints its own, so grid ↔ list still resets the scroll position. Deliberately **not** in `ManageFreezerSheet` or `FreezeProfileEditorSheet` — inside a `ModalBottomSheet` a scroll affordance competes with drag-to-dismiss.

**Row 20 — export the app list.** CSV of `displayedApps` **verbatim**: tab, search, filter and sort are already applied to it, so the file is exactly the list on screen. Save and share are separate paths — a share is not a save, so it stages a copy and hands over a `content://` URI without writing anything to Downloads; save goes through `ExportAppUseCase.openSession`, stale-SAF-tree clear included. Fields are neutered against **formula injection** (a leading `'` when a value starts with `=`, `+`, `-`, `@`, TAB or CR, on top of RFC 4180 quoting), because app labels are third-party-declared and this file's whole purpose is to be opened in a spreadsheet. No install-size column: a size is a `StorageStatsManager` round trip per package, which is the difference between a write that finishes before the sheet closes and one needing a progress bar.

**Row 21 — #130, installer names and the chart drill-down.** The chart keyed its buckets on `substringAfterLast(".").uppercase()`, so `com.aurora.store` drew as "STORE" and **any two installers sharing a last segment were silently summed into one bar**. Buckets now key on the package id; a bar names the installer, resolving curated → `PackageManager` label → raw id. Tapping a bar opens the Apps tab filtered to that installer, carrying the chart's User/System selection across. The new `InstallerLabelResolver` port also closed a second gap: the Apps tab resolved an installer's name by looking it up in the list it happened to be showing, so Aurora's apps read "Aurora Store" on the User tab and the raw id on the System tab.

## Notes for review

- **`stageText` wipes its directory on entry, and that is deliberate.** `ExportAppUseCase.exportInto` deletes its staged copy in a `finally`, which is illegal for a shared file: the receiving app opens the URI long after the call returns. Wiping on entry is the one moment nobody can still be reading the last one.
- **`FakeAppBundleFileStore` became functional** rather than throwing. Its old comment blamed `Uri.parse`; that constraint belongs to `ShareAppUseCase`, not to this port, whose `shareUri` returns a plain `String`. Bundle *building* still fails loudly — that half genuinely needs a device.
- **Strings land in one blocking first commit**, all five locales together, because `MissingTranslation` is a fatal lint error here.
- **MD was dropped** from row 20's "CSV/MD" title: a Markdown table of three hundred apps is neither readable by a person nor parseable by a tool.

## Still owed

- The **release-notes line** retracting `release-notes/v1.94.0/github.md:84` — band A #1 is a capability removal and the notes currently say the opposite. Due on the `master` merge.
- Two device probes (spec §9): `pm set-installer` as a replacement for the reinstall, and whether Dhizuku's shell can reach `PackageManagerService` at all. If it cannot, Reinstall should be hidden in that mode and `web/src/pages/features.mdx:111` corrected.

## Tests

**842 unit tests, 0 failures** (was 821 on `dev`), counts read from `app/build/test-results/**/*.xml`. `./gradlew lint` is green.

> Updated after self-review: the 842nd test pins that Clear All Cache excludes the *running* application id rather than the literal `"com.valhalla.thor"`. See the review comments below.

## Summary by CodeRabbit

* **New Features**
  * Export visible app lists as CSV or share them directly.
  * View installer distribution charts with localized labels and filter apps by installer.
  * Use the Fix Store picker to select apps before repairing installation records.
  * Stop eligible batch operations while they are running.
  * See scroll-position indicators in app grids and lists.
* **Improvements**
  * Improved installer grouping and handling of unknown installation sources.
  * Added localized text for export, installer, repair, and operation controls.
  * Improved export feedback for empty lists and failed operations.


trinadhthatakula:
## Self code-review

Read the whole diff (33 code files, 2 096 insertions) end to end. Four findings, none of which
fail today; two of them are the same class of defect this PR set out to end, which is why they
are worth fixing here rather than filing.

---

### 1. `LIST_STAGING_DIR` and `TEXT_STAGING_DIR` are the same string — and one of them is a `deleteRecursively()`

`ExportAppListUseCase.kt:88` — `const val LIST_STAGING_DIR = "list_export"`
`AppBundleFileStoreImpl.kt:184` — `const val TEXT_STAGING_DIR = "list_export"`

`stageText` wipes its directory on entry (`AppBundleFileStoreImpl.kt:94-96`):

```kotlin
val dir = File(context.cacheDir, TEXT_STAGING_DIR)
if (dir.exists()) dir.deleteRecursively()
```

`LIST_STAGING_DIR`'s own KDoc says it exists so that *"a future writer"* cannot *"assume a list
export shares the single-export scope and wip[e] a bundle mid-stream"* — and then gives it the
exact value of the directory that gets wiped. A bundle session opened with
`stagingSubDir = "list_export"` stages at `cacheDir/list_export/<pkg>` (`AppBundleBuilderImpl`),
which any concurrent list share deletes out from under it.

**No live failure.** `writeStaged` never reads `session.stagingSubDir` — only `exportSingle`'s
`build()` does, and this path never calls it — so the constant is inert today. But the collision
is exactly what the comment claims the name prevents. Fix: give it a value that cannot alias.

### 2. `clearAllCache` still hardcodes the two package ids this PR just centralised

`MainViewModel.kt:405-409`:

```kotlin
// 2. Filter out self and Play Store to be safe
val safeList = targetList.filter {
    it.packageName != "com.valhalla.thor" &&
            it.packageName != "com.android.vending"
}
```

This is the *same drift class* the PR's own `Installers.kt` KDoc describes — a rule written out
as literals in more than one place, where one copy learns something the others don't. Fix Store's
copy got `fixStoreCandidates` and `BuildConfig.APPLICATION_ID`; this one was left behind three
methods away.

Worse than cosmetic: `"com.valhalla.thor"` is **wrong on debug builds**. `applicationIdSuffix`
makes the running package `com.valhalla.thor.debug`, so the self-exclusion silently stops
excluding self and Clear All Cache clears Thor's own cache. That is precisely why
`fixStoreCandidates` takes `thorPackageName` as a parameter instead of reading a constant.

### 3. `FixStoreRow` has two interactive nodes for one toggle

`FixStoreSheet.kt:174-181` — the `Row` is `clickable(onClick = onToggle)` **and** the `Checkbox`
keeps a live `onCheckedChange`. Pointer input is fine (the inner node consumes the tap, so no
double-fire), but semantics are not: TalkBack gets an unlabelled clickable container *and* a
separate checkbox, and neither announces the app name together with its checked state.

Both existing checkbox rows in the codebase (`AppList.kt:530`, `FreezerScreen.kt:242`) leave the
row inert and let only the checkbox act, so this row is the first to take a third shape. The
idiomatic single-node form is `Modifier.toggleable(role = Role.Checkbox)` with
`Checkbox(onCheckedChange = null)` — one merged node, announced as *"<app>, installed by <x>,
checked"*.

### 4. `LegendItem`'s new `clickable` carries no role

`AppDistributionChart.kt:188-191` — the drill-down made the legend entry clickable but left it
role-less, so TalkBack reads it as text that happens to say *"double tap to activate"*. It is a
button; `role = Role.Button` says so at no cost. (No new string: an `onClickLabel` would need a
five-locale addition for a marginal gain, and `MissingTranslation` is fatal here.)

---

### Considered and not changed

- **The redundant `installer == null ||` in `fixStoreCandidates`.** `null !in setOf<String>(…)`
  is already true, so the disjunct is logically dead. Leaving it: it is the only place in the
  source where the case the whole feature exists for — *Android recorded no installer at all* —
  is visible as a branch rather than as a property of `Set.contains`. Removing it would also drop
  the smart-cast that makes the `in` check operate on a non-null `String`.
- **`stopRequested` is a single flag for a single logger.** Two overlapping batches would share
  it, but the logger is a modal dialog and `startLogger` resets the flag, so there is no path
  that starts a second batch while the first is running.
- **`InstallerLabelResolverImpl` never invalidates.** A store renamed while Thor is alive keeps
  its old label until the process restarts. Deliberate — the alternative is a `PackageManager`
  hit per row per recomposition.
- **`EXTRA_STREAM` through `createChooser`.** Correct as written: `createChooser` runs
  `migrateExtraStreamToClipData`, which is what propagates `FLAG_GRANT_READ_URI_PERMISSION` to
  the chosen target. Matches `MainScreen.kt:274-281`.

trinadhthatakula:
## Review findings addressed — `68d9815b`

| # | Finding | Outcome |
|---|---------|---------|
| 1 | `LIST_STAGING_DIR` aliased the directory `stageText` wipes | **Fixed** — `"list_export_session"`; the KDoc now names the value it must never take instead of gesturing at the hazard |
| 2 | `clearAllCache` hardcoded `"com.valhalla.thor"` / `"com.android.vending"` | **Fixed** — `BuildConfig.APPLICATION_ID` + `Installers.PLAY_STORE` |
| 3 | `FixStoreRow` was two semantics nodes for one toggle | **Fixed** — one `toggleable(role = Role.Checkbox)` node, `Checkbox(onCheckedChange = null)` |
| 4 | `LegendItem`'s `clickable` had no role | **Fixed** — `role = Role.Button` |

**Finding 2 was pinned by its own tests.** `MainViewModelTest` spelled Thor's package as
`"com.valhalla.thor"` in the fixture, so the two clear-all-cache tests asserted the literal rather
than the behaviour and stayed green while the debug build cleared its own cache. Both now build
the fixture from `BuildConfig.APPLICATION_ID`, and a third test puts **both** ids in the list:

```kotlin
appRepository.apps.value = listOf(
    userApp("com.valhalla.thor"),
    userApp("com.valhalla.thor.debug")
)
…
val expected = listOf("com.valhalla.thor", "com.valhalla.thor.debug")
    .filter { it != BuildConfig.APPLICATION_ID }
    .map { "clearCache:$it" }
assertEquals(expected, system.calls)
```

Exactly the id Thor is running as is spared and the other is treated as an ordinary third-party
app, so the literal cannot drift back in without going red.

**842 unit tests, 0 failures** (was 841 — the new one is the suffix test). `./gradlew lint` BUILD
SUCCESSFUL. Counts taken from `build/test-results/**/*.xml`, not the log line.

PR #373, opened 2026-08-09
fix(cache): per-app clear is root-only; the tile trims the whole device, and says so

trinadhthatakula:
## What this fixes

Clearing **one app's** cache calls `IPackageManager.deleteApplicationCacheFiles`, which
PackageManagerService gates on `android.permission.INTERNAL_DELETE_CACHE_FILES`. That permission is
`protectionLevel="signature"` — platform-signed callers only. `com.android.shell` does not request
it, and `pm grant` refuses it outright (`… is not a changeable permission type`), so a
Shizuku-delegated call reaches PMS as uid 2000 and is answered with:

```
Calling uid 2000 does not have android.permission.INTERNAL_DELETE_CACHE_FILES, silently ignoring
```

*Silently.* No exception, no failed `ShellResult` — Thor showed a success message and nothing had
happened. Dhizuku has no Device Owner cache API at all and was in the same position. Root works
because `ActivityManager.checkComponentPermission` short-circuits on `ROOT_UID` before the
permission is consulted.

So the per-app clear is now **root-only** — Shizuku and Dhizuku lose a button that never worked
rather than a capability.

The **Home tile is the opposite case and keeps Shizuku**: it runs `pm trim-caches`, which reaches
PMS via `freeStorage` and is gated on `CLEAR_APP_CACHE`, a permission shell *does* hold. But
`freeStorage` takes no package list — PMS evicts by LRU across the whole volume. "User apps" and
"system apps" were never choices Thor could honour, so the two dialogs offering them are gone.

## What's new

- **One confirmation sheet** before the tile clears anything, saying plainly that every app is
  included, **system apps as well**.
- **Freed bytes are measured and shown.** Per-app: `queryStatsForPackage` either side, reported as a
  toast. Whole-device: summed `queryStatsForUser`, shown in the result sheet.
- **The result sheet auto-dismisses after 3 s** unless dismissed first.
- **`getFreeBytes` as a target, not a measurement.** It returns usable space *plus* clearable cache,
  so a before/after delta cancels to noise — but it names exactly the point where all clearable
  cache has been reclaimed, which is what `pm trim-caches` wants as its argument. Never a round
  number: `freeStorage` is an escalating ladder and an unsatisfiable target walks past cache into
  `pruneUnusedStaticSharedLibraries` and instant-app eviction.
- **Null ≠ zero.** Both paths report `null` when usage access is missing, and the UI distinguishes
  "freed nothing" from "could not tell".

## Notes for review

- `ClearAllCacheSheet` is hosted by `MainScreen`, not `HomeScreen`: the clear outlives the screen
  that started it, so switching tabs mid-operation must not cancel it or lose the count.
- While running, the sheet vetoes its own settle to `Hidden` **and** swallows the dismiss request.
  The callback guard alone leaves an invisible sheet holding the result — `ModalBottomSheet`'s
  scrim/back path calls `onDismissRequest()` once `hide()` returns whether it animated or not.
- The double-confirm guard is deliberately **synchronous and outside** `viewModelScope.launch`. A
  guard inside the coroutine can observe `Done` (not `Running`) and start a second trim.
- `HomeActionsTest`'s five-reachable-state table is unchanged: `canClearCache` ⇒ `hasPrivilege`
  still holds, Dhizuku is now the one privileged mode that cannot clear.
- `ObserverCallSitesTest` lost the `Shizuku.kt#clearCache` site because the call site itself is gone.

## Verification

- `:app:compileFossDebugKotlin` — BUILD SUCCESSFUL
- `:app:testFossDebugUnitTest --rerun-tasks` — **847 tests, 0 failures, 0 errors, 0 skipped**
  (counted from `build/test-results/**/*.xml`)
- `:app:compileFossDebugAndroidTestKotlin` — compiles
- `:app:lintFossDebug` — BUILD SUCCESSFUL (six orphaned strings removed, seven added and translated
  across `fr`, `es`, `ar`, `zh-rCN`)

Not device-tested.

## Summary by CodeRabbit

- **New Features**
  - Added whole-device cache clearing with confirmation, progress, and completion states.
  - Reports freed storage when measurements are available, including zero-byte results.
  - Added cache clearing for Root and Shizuku access.
  - Per-app cache clearing remains available with Root access.
- **Bug Fixes**
  - Removed unsupported cache-clearing actions from Dhizuku and unavailable access modes.
  - Expanded cleanup to include device-encrypted cache locations.
- **Localization**
  - Added translated messages for the updated cache-clearing flow.

### DIFF
--- a/app/src/androidTest/java/com/valhalla/thor/presentation/home/HomeActionsBentoTest.kt
+++ b/app/src/androidTest/java/com/valhalla/thor/presentation/home/HomeActionsBentoTest.kt
@@ -42,7 +42,7 @@ class HomeActionsBentoTest {
     private fun setFullGrid() = rule.setContent {
         HomeActionsBento(
             reinstallVisible = true,
-            isRoot = true,
+            canClearCache = true,
             hasPrivilege = true,
             unknownInstallerCount = 7,
             selectedTypeName = "user",
@@ -108,7 +108,7 @@ class HomeActionsBentoTest {
         rule.setContent {
             HomeActionsBento(
                 reinstallVisible = true,
-                isRoot = false,
+                canClearCache = false,
                 hasPrivilege = true,
                 unknownInstallerCount = 7,
                 selectedTypeName = "user",
--- a/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
@@ -60,9 +60,24 @@ class DhizukuSystemGateway(
         else Result.failure(Exception("Dhizuku: Force stop failed. Shell command and reflection both denied."))
     }
 
-    override suspend fun clearCache(packageName: String): Result<Unit> {
-        return if (reflector.clearCache(packageName)) Result.success(Unit)
-        else Result.failure(Exception("Dhizuku: Clear cache failed. System reflection and shell rm -rf both failed."))
+    /**
+     * Always a failure, and that is the correction rather than a gap.
+     *
+     * `pm trim-caches` is a shell command, and Dhizuku has no shell: [executeShellCommand] runs
+     * through the Dhizuku app's own uid, which `PackageManagerShellCommand` refuses. The device
+     * owner API has no cache-clearing member at all — `DevicePolicyManager` can wipe a profile, not
+     * a cache — and the reflective `deleteApplicationCacheFiles*` rung this gateway used to carry
+     * died in a double-wrapped binder belonging to a privilege mode the user had not set up. Three
+     * doors, all shut, so this says so in a sentence the user can act on instead of failing with a
+     * shell error that reads like a bug.
+     */
+    override suspend fun clearAllCaches(targetFreeBytes: Long?): Result<Unit> {
+        return Result.failure(
+            Exception(
+                "Dhizuku cannot clear caches: it has no shell to run `pm trim-caches` in, and the " +
+                    "device owner API has no equivalent. Switch to Root or Shizuku."
+            )
+        )
     }
 
     override suspend fun clearAppData(packageName: String): Result<Unit> {
@@ -380,10 +395,6 @@ class DhizukuSystemGateway(
         } else {
             Result.failure(Exception("Dhizuku: Install failed: ${result.second}"))
         }
-    }
-
-    override suspend fun getAppCacheSize(packageName: String): Long {
-        return 0L
     }
 
     override suspend fun reinstallAppWithGoogle(packageName: String): Result<Unit> {
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -386,6 +386,14 @@ class RootSystemGateway(
      * same three directories per package that [clearCache] does, with the package component globbed.
      * It is also the reason the trim's exit code is checked rather than assumed — a `pm` that fails
      * for any reason still leaves a mode that works.
+     *
+     * **The two rungs do not have the same reach, and the narrower one is the deliberate part.**
+     * `pm trim-caches` is volume-wide, so it takes every Android user with it; the sweep is scoped to
+     * [thorUserId] and leaves a work profile's or a Second Space's caches alone. That is not an
+     * oversight to be widened later — `PerUserCommands` exists because deleting another user's data
+     * from a command that names no user is the defect this codebase kept finding, and an `rm -rf`
+     * across every user's tree would be the largest instance of it. A user's other profiles are not
+     * Thor's to empty on a tile tap taken in this one.
      */
     override suspend fun clearAllCaches(targetFreeBytes: Long?): Result<Unit> {
         if (targetFreeBytes != null) {
--- a/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt
@@ -349,18 +349,58 @@ class RootSystemGateway(
         )
     }
 
-    override suspend fun clearCache(packageName: String): Result<Unit> {
+    /**
+     * Deletes [packageName]'s cache directories for [thorUserId].
+     *
+     * **Not an override.** This is the only privilege mode that can clear one app's cache, so it
+     * sits off [SystemGateway] rather than on it — the same shape `copyFile` and `getAppPaths`
+     * already have, and `SystemRepositoryImpl` reaches it the same way, behind `isRootAvailable()`.
+     * The Shizuku and Dhizuku implementations that used to satisfy an interface method here are
+     * gone; [SystemGateway.clearAllCaches] records why they could not have worked.
+     *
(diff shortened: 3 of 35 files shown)


trinadhthatakula:
## Code review (self-review of this branch)

Read the whole diff back against the platform behaviour it claims. One finding, fixed in
`780a1a52`; everything else below is a note recording what was checked and why it stands.

### 🐞 A measured zero was reported as a missing permission — fixed in `780a1a52`

`SystemRepositoryImpl.measuringCacheFreed` deliberately keeps `null` and `0` apart, and its KDoc
spells out why. `MainScreen` then collapsed them again on the way to the sheet:

```kotlin
formattedFreedBytes = (cacheClear as? CacheClearState.Done)
    ?.freedBytes
    ?.takeIf { it > 0L }          // <-- a measured zero becomes null here
    ?.let { Formatter.formatShortFileSize(context, it) },
```

The sheet renders a null as *"Thor could not measure how much this freed — grant it usage access in
Settings to see the figure next time."* So clearing twice in a row — the second clear genuinely
freeing nothing — told a user who **demonstrably has** usage access to go and grant usage access.
The one piece of advice on screen was both wrong and unactionable.

Zero is now formatted like any other number and the `Done` branch picks between three sentences:

| `freedBytes` | Message |
|---|---|
| `null` | "Thor could not measure…" (usage access missing) |
| `0` | "There was no cache left to clear." (**new** — `clear_all_cache_done_nothing`, translated ×4) |
| `> 0` | "Freed 12 MB." |

Negative deltas still cannot reach the UI: the repository clamps them to `null`, because cache an
app rebuilt between the two readings is an unmeasurable, not a zero.

The per-app toast path was checked for the same conflation and does not have it — it degrades to
the neutral "Cleared cache for X" for both `null` and `0`, which claims nothing either way.

### ✅ Checked and standing

- **Sheet cannot be swiped away mid-clear.** `onDismissRequest = { }` while `Running` only swallows
  the *callback*; the sheet would still settle to `Hidden` and sit there invisible holding the
  result. `confirmValueChange` now vetoes the settle itself. Both guards are needed —
  `ModalBottomSheet`'s scrim/back path (`settleToDismiss`) calls `onDismissRequest()` once `hide()`
  returns, veto or no veto — and `hide()` consults `confirmValueChange` too
  (`SheetDefaults.kt:303`), so the two cover different halves.
- **`confirmValueChange` is a single stable instance.** `rememberBottomSheetState` passes it
  straight into `rememberSaveable` as a **key** (`SheetDefaults.kt:717-722`), so a lambda whose
  identity changed per recomposition would discard the live `SheetState` mid-operation and snap the
  sheet shut. It is `remember`ed once and reads the running flag through `rememberUpdatedState`.
- **Double-confirm guard is synchronous and outside `viewModelScope.launch`.** Inside the coroutine
  it can observe `Done` (not `Running`) and start a second trim; there is a test for exactly that.
- **`pm trim-caches` target is never a round number.** `freeStorage` is an escalating ladder on
  which app cache is only rungs 4 and 8 — an unsatisfiable target walks on into
  `pruneUnusedStaticSharedLibraries` and instant-app eviction. Shizuku **fails** rather than
  guessing when `getFreeBytes` is unreadable; root survives a null target because it still has the
  direct sweep.
- **Root's sweep names three directories, matching `InstalldNativeService::clearAppData` under
  `FLAG_CLEAR_CACHE_ONLY`** — CE, DE and external. `code_cache` is deliberately excluded: different
  installd flag, Settings' Clear cache does not touch it, and it holds compiled artifacts apps
  expect to keep. The glob is expanded by the uid-0 shell and `*` never crosses a path separator, so
  it cannot reach outside the three parents.
- **Dhizuku fails with a sentence, not a shell error.** No shell for `pm`, no device-owner cache
  API. The Home tile is gated to Root ∪ Shizuku so it should never be reached, but it says why if it
  is.
- **`ObserverCallSitesTest` lost the `Shizuku.kt#clearCache` site because the call site is gone**,
  not because the guard was relaxed — the remaining count assertions were tightened to match.

### Verification

- `:app:testFossDebugUnitTest --rerun-tasks` — **847 tests, 0 failures, 0 errors, 0 skipped**
  (counted from `build/test-results/**/*.xml`, directory deleted first)
- `:app:lintFossDebug` — BUILD SUCCESSFUL
- `:app:compileFossDebugAndroidTestKotlin` — compiles

Not device-tested.

trinadhthatakula:
## Review round: what I took and what I didn't

Pushed as `c8b3d54a`. 849 unit tests green, `lintFossDebug` clean, `compileFossDebugAndroidTestKotlin` clean.

### Taken

**The unmeasured result over-attributed itself to usage access.** This was the best finding in the round — it's a real user-facing bug, not a style note. One sentence covered two outcomes: the op genuinely missing (advice is actionable) and the op held but the reading empty (advice is unfollowable, and it's the only thing on screen). `CacheClearState.Done` now carries `hasUsageAccess` next to the byte count, sampled in the ViewModel when the result lands — not at draw time, so the flag describes *that* measurement and the branch stays unit-testable. The gate is only consulted when the number is missing; a byte count arriving is already proof the op is held. Injected as `UsageAccessGate`, not `UsageAccessManager`, because the concrete class needs a `Context` for its Settings deep-link and that would take `MainViewModel` off the JVM test classpath.

That split needed a fourth branch too, which I found while wiring it: a *measured zero* — a second clear with nothing left — was falling into the unmeasured branch. Separate string (`clear_all_cache_done_nothing`), fixed in `780a1a52`.

**fr:** `l'accès aux données d'utilisation`, matching the existing `usage_access` string at `values-fr:530`. Two terms for one permission is worse than a slightly long one.

**zh-rCN:** the confirm line read "system apps and your apps are the same" rather than "both are included". Reworded.

**The failure test asserted `assertEquals(1, effects.size)`.** Correct — a success message is also one effect. It now asserts the whole `MainSideEffect.Message`.

### Not taken

**Widening the `rm` fallback across all Android users.** `pm trim-caches` is volume-wide so rung one already takes every user with it, but the fallback sweep is deliberately scoped to `thorUserId`. Deleting data belonging to a user you weren't asked about, from a command that names no user, is precisely the defect `PerUserCommands` was introduced to stop — and an `rm -rf` over every user's tree would be the largest instance of it in the codebase. A work profile's or a Second Space's caches are not Thor's to empty on a tile tap taken in this profile. I've written the reasoning into the `clearAllCaches` KDoc so the next reader doesn't file it as an oversight.

**Localizing the three gateway failure messages via `UiTextException`.** `UiTextException` is a domain/ViewModel-layer pattern: **0** of the sites under `data/gateway/` use it, and **38** raise plain-English `Result.failure(Exception(...))`. Converting 3 of 38 doesn't localize the layer, it creates a fourth copy of a rule that already has three — which is the exact failure mode this repo spent Band B cleaning up. If the gateway layer should speak `UiText`, that's one sweep of 38 sites, not a drive-by on the ones this PR happened to touch.

### Found in self-review, not flagged

The sheet could settle to `Hidden` mid-clear. `onDismissRequest = { }` swallows the *callback*; it does nothing about the settle, so a swipe or scrim tap left the sheet animated away and invisible while the trim was still running, taking the byte count with it. Fixed with `confirmValueChange` — and the lambda is `remember`ed rather than capturing `state` inline, because `rememberBottomSheetState` passes `confirmValueChange` into `rememberSaveable` as a **key**: an identity-unstable lambda would discard the live `SheetState` on the next recomposition and snap the sheet shut. Both guards are needed; M3's back/scrim path calls `onDismissRequest()` once `hide()` returns whether the veto fired or not.

PR #376, opened 2026-08-10
feat: OBB support in .xapk export and install (#164)

### LINKED ISSUE #164: .APK, .XAPK, .APKS and Split bundle export?

trinadhthatakula:
Closes #164.

`.xapk` export dropped every app's OBB files, and `.xapk` install ignored any that were bundled.
Both halves are fixed.

#164 was closed on 2026-08-03 with the export gap **stated** as a limitation. It reopened three
days later when @playagain96 posted a recording of an export whose `.xapk` carried no OBB *"even
tho app has it visible and accessible"*. A limitation stated in an explain string is still a
missing feature.

## Export

The builder probes `Android/obb/<pkg>` through the privileged shell, has that shell stage each
expansion into `externalCacheDir` — the only location Thor's uid and the shell's uid can both
reach — streams them into the archive at `Android/obb/<pkg>/<leaf>.obb`, and declares them in the
manifest's `expansions` block. `file` and `install_path` are written equal, which is both what
reference installers assume and what makes a manifest-blind installer land the files correctly.

## Install

A bundled `expansions` block is validated against the package actually being installed and placed
into `Android/obb/<pkg>/`. An archive whose game data cannot be placed is refused **before**
anything is installed, rather than leaving a game that installs cleanly and fails at play time —
the same broken outcome #164 reports, arrived at from the other direction.

## Not root-only

`executeShellCommand` is routed through `runGatewayAction`, so root and Shizuku both work. Only
Dhizuku's device-owner process cannot see another package's external directories: there the
`.xapk` chip is **disabled with the reason shown** rather than hidden, per the chosen failure
policy — a chip that silently vanishes leaves the user with no way to find out why the format they
came for is missing.

⚠️ **Behaviour change for unprivileged users.** With no privilege mode active, installing an
OBB-carrying `.xapk` is now refused up front instead of installing a game that won't run. The
copy says "needs root or Shizuku".

## Two corrections this also ships

**A factually wrong justification.** Thor told users that *"Android 11 and later stops any app from
reading another app's OBB folder"*. True of an ordinary app, false of Thor, which reads protected
paths through a privileged shell on every split export. Retracted in all five locales (en, ar, es,
fr, zh-rCN).

**A dead card.** The app-info OBB row was computed with `File(...).exists()`, which returns false
for another package's OBB directory whether or not one exists — so it was absent for every game on
a modern device. It now reads the probe, and renders one of three states: the directory with a
size, "No game data", or "Can't check without root or Shizuku". The three-state model is the point:
folding *"I cannot see it"* into *"there is none"* is the bug this issue is about.

## Verification

- `:app:assembleFossDebug` and `:app:lintFossDebug` — **BUILD SUCCESSFUL**, lint clean (it is fatal
  here: `warningsAsErrors` plus `checkTestSources`).
- **927 unit tests / 0 failures / 0 errors across 74 classes**, counted from
  `app/build/test-results/**/*.xml` rather than from Gradle's log line. Baseline before this branch
  was 842.
- ⚠️ **Nothing here has run on a device.** The unit tests cover every pure decision — the probe
  parser, the wire format both directions, the path convention, entry-name mapping, the extraction
  budgets, the placement decision — and **none** of the privileged paths, which are not
  JVM-testable. The nine device checks are listed in the plan's Task 12 Step 6 and are unrun.

## Eleven defects found and fixed in review

Each was found by reviewing a task's output against its brief, and each is recorded in the plan so
the plan stops disagreeing with the code:

| Defect | Why it mattered |
|---|---|
| Root-shell injection in `isSafeObbLeafName` | An `.obb` filename is attacker-controlled and reached a shell |
| `THOR_NODIR` checked *after* the end sentinel | "No OBB directory" was reported as "probe failed" |
| Builder didn't refuse an `Undetermined` probe | A comment claimed a re-check that did not exist; the UI chip was the only defence |
| `Present` was not a promise | A legitimate `.obb` containing `'` passed the probe and failed at staging |
| No entry-count cap on extraction | A manifest-free `.xapk` with a million 1-byte `.obb` entries stays inside the 16 GiB budget |
| `extractExpansions` didn't dedup leaves | Two entries would resolve to one file and charge the budget twice |
| `isInstalled()` right after `session.commit()` | Not a completed-install check — a **fresh** install via a session silently placed no OBB, reproducing #164 on a new path |
| Staging dir outside every `catch`'s reach | A failed export leaked multi-GB into `externalCacheDir` |
| `return@withContext` inside a `try` | Skips the `catch`, so failures returned success |
| `AppInfoSheet` never got `obbProbe` | A **defaulted** parameter hid the omission; the new card would have shipped invisible on the surface most users reach |
| Arabic transliterated the app name | One locale file disagreeing with itself 48 times over |

Design: `docs/superpowers/specs/2026-08-10-xapk-obb-support-design.md`
Plan: `docs/superpowers/plans/2026-08-10-xapk-obb-support-implementation.md`

App **data** backup (#51 phase 2 / band C row 23) is deliberately **not** in this branch.

## Summary by CodeRabbit

* **New Features**
  * `.xapk` exports now include supported OBB game data and installation metadata.
  * XAPK installation places expansion files in the correct location after APK installation.
  * App details show whether OBB data is present, absent, or inaccessible.
  * Export screens display included OBB size and partial-inclusion warnings.
* **Bug Fixes**
  * Added validation for unsafe paths, missing files, storage limits, and incomplete transfers.
  * XAPK export is disabled when required storage access cannot be verified.
* **Localization**
  * Updated Arabic, Chinese, English, French, and Spanish messaging for OBB status and XAPK exports.


trinadhthatakula:
verified on physical device and emulators.

PR #379, opened 2026-08-12
feat(backup): app data backup and restore (#51 phase 2)

### LINKED ISSUE #51: New feature requested

trinadhthatakula:
#51 phase 2 — app **data** backup and restore. The `README.md` promise "BackUp App Data" has been
half-kept since phase 1 shipped the APK half; this is the other half.

Built across 18 tasks to the spec and plan committed alongside it
(`docs/superpowers/specs/2026-08-10-app-backup-and-restore-design.md`), then reviewed by six
reviewers over disjoint slices, fixed by six sequential fixers, and re-reviewed across five lenses
with an adversarial verifier on every finding. **1362 → 1467 tests**, lint held at 0 Error /
0 Warning / 6 Hint the whole way.

## The format

`.thorbak` is a plain zip named `<pkg>-<versionCode>.thorbak`, holding `thorbak.json`, `app.xapk`,
and one AES-256-GCM member per selected storage class. Nothing about it is bespoke: any zip tool
opens the container, the header is readable without the passphrase, and only the data members are
encrypted. `schemaVersion` is written even when it equals its default, and a test pins that a v1
reader survives a v2 document carrying an extra field.

## The privilege gate is a probe, not a root check

It runs a real command through `executeShellCommand`, which `runGatewayAction` routes to whichever
gateway is active — so a root-started Shizuku shell passes it. Writing "requires Root" would have
been a lie on that device.

## Passphrase handling

The passphrase is never the source of truth for whether one is stored: `PassphraseVault` is a cache
over `AndroidKeyStore` + DataStore, and `isRemembered` reads through to the store. The derived key is
**process-scoped by design** — a killed job is reported through a breadcrumb, not resumed, because a
resume would have to survive the same kill that stopped it.

**The passphrase is not recoverable.** That is a property of the format, and the release notes have to
say so; a user who learns it after losing a passphrase has already lost the archive.

## The job seam

A reusable WorkManager foreground-job seam carries both workers. It was built to carry the existing
APK/XAPK export and the bulk actions too, but moving shipped export paths inside a branch that cannot
be device-tested would put a regression on working code — so those migrations are deferred, each with
its reason, in the linked follow-up.

## What the review found, and what it means

The two Criticals are worth stating because they had one cause. Both were independent ways to produce
an **unrestorable backup**, and they were found by two reviewers who could not see each other's
slices — because *nothing on the branch wrote an archive and then restored it*. Every test on either
side used a fixture the other side never produced. `ArchiveRoundTripTest` now closes that seam as far
as the JVM allows, and one of the Criticals was a missing `zip.finish()` that had left two green test
halves and one archive `ZipFile` could not open.

The re-review returned **zero blockers**. Its refutations are the more useful half: of eleven "the fix
wave broke this" attributions, **six did not survive verification** — including one alleging that a
failed publish now destroys a finished multi-gigabyte archive, refuted on platform
`DocumentsProvider` sources. Six findings are recorded in the follow-up doc as *decisions, not
defects*, precisely so the next reviewer doesn't re-raise them and "fix" a deliberate choice.

## ⚠️ Release obligations — not advice

- **File the Play Console Foreground Service declaration before the first `store` upload** carrying
  `FOREGROUND_SERVICE_DATA_SYNC`. This is not a post-release fix; the upload is rejected.
- Declare the type as **`dataSync`**, matching the workers' `ForegroundInfo`. A manifest that
  disagrees with the `ForegroundInfo` is a **crash** on Android 14, not a warning — fix whichever is
  wrong, never suppress the lint check.
- Record the demonstration video on **bulk APK export on an unrooted device**, not on a backup. A data
  backup needs root, Play will not review on a rooted device, so the backup path is not one a reviewer
  can run. Bulk export uses the same foreground service for the same reason.
- `strings_backup.xml` is **English-only**, behind a file-level `tools:ignore="MissingTranslation"`.
  ~50 strings. That suppression is the debt, and it must not be copied into any other file.

## ⚠️ Desk-verified only — 21 device checks, every one unrun

Every task's gate was a build plus JVM unit tests. **No part of this has run on a device.** The full
list is in `docs/follow-ups/app-data-backup-and-xapk-export.md`; the checks are grouped by what they
cover (cryptography and storage, the privileged shells, zip and the archive, the job seam, the sheet,
restore) because a pass in one group says nothing about another.

The honest caveat is not any single check, it is what no check in this repository can reach:

> **`ThorJobLauncher` → WorkManager → worker → privileged shell — the region that actually replaces a
> user's app data — is exercised by no test here, and cannot be.** It needs `Context`, and this module
> has no Robolectric, no `koin-test` and no `work-testing`. There are also zero Compose tests over
> `ArchiveRestoreScreen`, `AppBackupSheet` or the passphrase sheet. Compounding it, a shipped Thor
> emits nothing to logcat under any tag it owns, so a failure there is both untested and
> undiagnosable from a user bug report.

**Merge on the review; do not ship to users until someone has taken a backup and restored it on
hardware.** That single check is worth more than everything above it.

## Do not close #51 on this merge

The code exists, but *"has run on a device"* is the bar for closing it, and 21 checks say it has not.

## Relationship to #378

#378 (the `.xapk` OBB probe fix) is a separate PR into `dev` and does **not** overlap in effect. Both
merge cleanly into `dev` and cleanly against each other, verified with `git merge-tree`, despite three
shared files (`AppBundleBuilderImpl.kt`, `InstallerRepositoryImpl.kt`, `SystemRepositoryImpl.kt`).
Merge order does not matter.

## Gate

**1467 tests / 0 failures / 0 errors / 0 skipped** across 107 XML files, counted from
`app/build/test-results/**/*.xml`, on a full `--rerun-tasks` run of the exact tree in this PR. Lint
**0 Error / 0 Warning / 6 Hint**. `:app:assembleFossDebug` green. Every behavioural fix in the review
wave was proven by mutation — apply, run, confirm killed, revert, confirm the revert.

## Summary by CodeRabbit

* **New Features**
  * Added encrypted `.thorbak` app backups with selectable data and optional app bundles.
  * Added archive restore with passphrase protection, app compatibility checks, progress tracking, and expansion-file restoration.
  * Added backup and restore screens, app actions, settings controls, and support for opening `.thorbak` files directly.
  * Added remembered passphrase management with secure storage.
  * Added foreground notifications, cancellation, and recovery notices for long-running jobs.

* **Bug Fixes**
  * Improved cleanup of interrupted or incomplete archives and staging files.
  * Added validation for unsafe archive paths and oversized expansion-file names.


trinadhthatakula:
## Review adjudication — 13 findings, 4 fixed, 9 refuted

Every finding was verified against the code, and each survivor was then handed to an adversarial
check whose job was to refute it. Nine did not survive. Fixed in 02e89df1.

### Refuted

| Finding | Why it does not hold |
|---|---|
| 🔴 **`LaunchGuarded.kt` — both progress collectors can terminate the process** | The mechanism is real Kotlin semantics; the *defect* is unreachable. Both inner `launch`es are `registry.progressOf(id).collect { _uiState.update { it.copy(progress = …) } }`. Upstream is a bare `MutableStateFlow` whose only throw is `ensureActive()`'s cancellation; the body is a `compareAndSet` loop over a data-class copy, and neither `AppBackupUiState` nor `ArchiveRestoreUiState` has an `init`, `require` or `check`. No input reaches the crash. See the residue below — the comments were wrong, the code was not. |
| **`ArchiveJobLauncher.kt:67` — secrets out of `Data`** | No secret is written. The passphrase never reaches the launcher, and the key travels through `ArchiveKeyHolder` in memory precisely so that it does not go to WorkManager. |
| **`AppArchiveStoreImpl.kt:167` — rethrow `CancellationException`** | Already correct on the path as written. |
| **`ArchiveRestoreViewModel.kt:437` — a vault failure reported as unreadable** | The two are distinct arms already. |
| **`MeasureAppDataUseCaseTest.kt:114` — distinct validation result** | The assertion does discriminate. |
| **`design.md:6`** | A historical snapshot, correct as of its own date. |
| **`AppArchiveStoreImpl.kt:471` — guard `output.close`** | Deliberate and recorded: `docs/follow-ups/app-data-backup-and-xapk-export.md`, "Do NOT file these as defects". |
| **`AppDataArchiveGateway.kt:99` / `ArchiveJobLauncher.kt:77`** | Style. `runningJobFor` was never an enqueue guard and does not claim to be. |

### Fixed — three of the four differ from the suggested diff

**`AppBackupSheet.kt:493` — accessible name.** Real, and it covers seven rows including the
confirm-replace acknowledgement. Fixed with `toggleable(role = Role.Checkbox)` + `onCheckedChange =
null`, matching `FixStoreSheet`, **plus `heightIn(min = 48.dp)` which the suggestion omits**:
`Checkbox` applies `minimumInteractiveComponentSize()` *only while it owns the click*, so handing the
click to the row gives up the 48dp target and `toggleable` enforces no minimum. The three rows with a
null `detail` — including `restore_confirm_replace` — would have shrunk to ~28dp. The
"the touch target is the checkbox alone" clause is false today, and false in the direction that
damages the remedy.

**`ArchiveRestoreViewModel.kt:331` — a failed open blocks retrying the same file.** Real. Released via
`forgetUriIfStill`, on the two `FILE_UNREADABLE` paths only. **Not unconditionally**, as suggested:
`open(B)` does not cancel `open(A)`, so a late A-failure would clear a URI `open(B)` had already
stored, and `beginRestore` would then return silently with a valid header on screen. The two
`NOT_AN_ARCHIVE` arms deliberately keep the guard — there the advice is to pick a *different* file.
Both arms are now pinned.

**`ArchiveRestoreScreen.kt:134` — stale passphrase.** Real, but not a Security & Privacy issue, and the
screen's own comment says why: a Kotlin `String` cannot be zeroed, so `passphrase = ""` drops a
reference and nothing more, and `open()` already wipes the view model's `CharArray`. What it actually
fixes is a field pre-filled with text the user never typed for this archive, with Unlock enabled,
spending 210,000 PBKDF2 iterations to answer "wrong passphrase".

**`ThorJobWorker.kt:115` — unbounded values in `Data`.** Real, and the most useful of the thirteen —
but it named two write sites and missed both the busiest one and the root cause. `fail()` is where
every failure in both workers routes through, and `isSafeObbLeafName` had **no length bound** on a
leaf that `ObbExpansions.kt` itself documents as attacker-controlled and that gets quoted into a
restore warning. The site that matters is the *success* result: `Data.Builder.build()` throws rather
than truncating, the throw leaves `runJob`, and **a restore that had already completed is reported as
failed** — which sends the user to run it again over data that is already correct. Fixed at the source
(`NAME_MAX`, 255) and at the boundary (`boundedForJobData`, top-level so a JVM test can reach it).

### The Critical's residue

Two comments promised `onFailure` covered "either collector" while the guard reached only the status
one. Closed by running `block` through `coroutineScope`, which makes the comments true rather than
narrowing them — no observable behaviour change today, and the next `launch` added inside a guarded
block does not have to know this was ever load-bearing.

### Filed rather than fixed

Five checkbox/switch rows **outside this branch** carry the same accessible-name shape
(`SettingsScreen.kt:1197`, `FreezerSettingsSheet.kt:306`, `PermissionManagerScreen.kt:508`,
`FreezerScreen.kt:242`, `AppList.kt:530`). That is why the severity came down: this branch was
matching the app-wide status quo, not regressing against a convention. Recorded in
`docs/follow-ups/app-data-backup-and-xapk-export.md`.

1473 tests, lint clean.

PR #383, opened 2026-08-12
feat(settings): eight doors, and a second pane on wide windows

trinadhthatakula:
Settings was one `Column(verticalScroll)` holding 29 rows under 10 hand-written section headers — 1275 lines, and the last two rows had been added with a `koinInject` in the middle of the layout. This is options **C** and **D** from the previews, plus the two-line subtitles.

## What changed

**C — eight doors.** `SettingsCatalog.kt` declares eight categories and, per row, which category it belongs to. `SettingsCategoryScreen`'s single `when (row)` is exhaustive over `SettingsRowId`, so **a setting added to the catalogue and never drawn is a compile error**. That is the gate the old screen never had. The index is fixed at eight entries; the number of settings is not.

Depth stops at two. The index has a search field matching titles and static descriptions, deep-linking to the row and briefly highlighting it on arrival — so categories are a way to browse, not a maze to memorise. Each index row carries a summary line derived from live state.

**D — two panes.** On a wide window the category opens *beside* the index via `ListDetailSceneStrategy`, and the index shows which one is open. Opening a second category **replaces** rather than stacks, so Back means "leave Settings" on a tablet and "go up" on a phone without either needing to know which it is.

**Two-line subtitles.** Asgard's row already supported it; the descriptions were clipped at one line, which is why several were written as fragments that stop before the useful half.

## Three things that were wrong, not merely long

- **The route carries a string id**, not the enum and not its ordinal. `rememberNavBackStack` is persisted for task restoration and survives an app update, so a restored stack can name a category this build no longer has. Unknown id → null → the entry pops itself. An ordinal would have silently opened a *different* category the day the enum was reordered.
- **`ThemeMode.label()` returned `"Light"`/`"Dark"`/`"System"` as Kotlin literals** — a domain enum answering a question only the UI asks, in English, on all five locales, for the setting users change most. Deleted; every closed-set label now has a string resource. The privilege picker was worse: it rendered `mode.name`, so the segmented control read `ROOT / SHIZUKU / DHIZUKU` while `install_mode_root` and its two siblings had existed for the installer sheet all along.
- **`ANIMATION_INTENSITY`** had a section header duplicating its own row title; it is now an ordinary Appearance row. **`UNFREEZE_ALL`** is marked destructive and drops its chevron — it is an action, not a door.

## Review fixes (`3bf14f45`)

CodeRabbit found three, all real:

- **Back stack leak.** `ExtensionManager` is *itself* a `detailPane`, so on a wide window the index stays visible beside it — the user can pick a second category while the top of the stack is not a `SettingsCategory`, and the replace-guard fell through to `add`. Now truncates to the index.
- **Two dead switch rows.** Usage Access and Notification Access both guarded on the permission being *absent*, so once granted the switch bounced back and nothing opened. Both deep-link in either direction now.
- **The row was unreadable to a screen reader.** `clickable` + a `Switch` with cleared semantics meant nothing carried the on/off state. Now `toggleable(role = Role.Switch)` with a null handler on the Switch.

## Verification

| Gate | Result |
|---|---|
| `:app:compileFossDebugKotlin` + `:app:compileStoreDebugKotlin` | ✅ both flavours |
| `:app:lintFossDebug` | ✅ **0 errors, 0 warnings** (from 33 errors) |
| `:app:testFossDebugUnitTest --rerun-tasks` | ✅ **1523 tests, 0 failures** — counts parsed from `test-results/**/*.xml` |
| `:app:connectedFossDebugAndroidTest` (`SettingsSwitchRowTest`) | ✅ **7 tests, 0 failures** — Pixel 10 Pro Fold AVD, API 37 |
| Hand-driven on the same fold, 851 dp and 411 dp | ✅ see below |

`SettingsCatalogTest` (9 unit tests) pins what the exhaustive `when` cannot: no empty categories, no duplicate route ids, declaration order preserved, every row round-tripping through the route as a focus target. `SettingsSwitchRowTest` (7 instrumented) pins the role, the state, the single-control invariant and the report-the-requested-value contract that the two dead rows depended on.

**Device-checked by hand** (details in the thread): two panes populate correctly at 851 dp with the index row highlighted; the placeholder renders before a selection and is correctly absent at 411 dp; the search deep-link scrolls to the row and lights it; and both review fixes were reproduced broken and then confirmed fixed on hardware — including toggling a *granted* Usage Access, which now opens the system revoke screen.

## Please read before merging

- **The 27 new strings' fr / es / ar / zh-rCN translations are machine-quality and want a native reader.** They are here so `MissingTranslation` stays a real gate rather than a `lint.xml` override — not because I can vouch for the register. Arabic especially, and the two-pane layout is new so an RTL pass is worth doing.
- **One emulator, one API level (37), `foss` only.** No physical-device pass; `store` is compile-verified only.
- **Six strings were deleted** from five files: `general`, `appearance`, `security`, `work_mode`, `about`, `permissions` — orphaned when the section headers went, and flagged `UnusedResources`. If anything outside `:app` reads them, that is a lint blind spot I did not check.
- The bottom nav bar **stays visible** on a category screen, unlike every other non-root route. Deliberate: a category is one tap deep in a two-deep hierarchy, and it keeps the bottom inset identical between index and category. Say the word if it should hide.

### DIFF
--- a/app/src/androidTest/java/com/valhalla/thor/presentation/settings/SettingsSwitchRowTest.kt
+++ b/app/src/androidTest/java/com/valhalla/thor/presentation/settings/SettingsSwitchRowTest.kt
@@ -1,0 +1,157 @@
+// SPDX-FileCopyrightText: 2025-2026 Trinadh Thatakula <github.com/trinadhthatakula/Thor>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package com.valhalla.thor.presentation.settings
+
+import androidx.activity.ComponentActivity
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.SemanticsProperties
+import androidx.compose.ui.test.SemanticsMatcher
+import androidx.compose.ui.test.assert
+import androidx.compose.ui.test.assertIsOff
+import androidx.compose.ui.test.assertIsOn
+import androidx.compose.ui.test.assertIsNotEnabled
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.isToggleable
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTouchInput
+import androidx.compose.ui.test.percentOffset
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.valhalla.thor.R
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * What [SettingsSwitchRow] promises the twelve settings that use it.
+ *
+ * Two of those promises were being broken at once, in opposite directions, and neither was visible
+ * from reading a call site. The row announced itself as a *button* — `clickable`, no state — while
+ * the `Switch` beside it had its semantics cleared so the setting would not be offered twice, so
+ * between them nothing carried on or off and a screen reader user could change any toggle in
+ * Settings without ever hearing what it was set to. Meanwhile the Usage Access and Notification
+ * Access handlers both read the row as reporting "the state I am now in" and guarded on the
+ * permission still being missing, which made both rows dead once granted.
+ *
+ * So: the row is one control, it says which control it is, it says what it is set to, and it
+ * reports the value the user asked for rather than the one already in force.
+ */
+@RunWith(AndroidJUnit4::class)
+class SettingsSwitchRowTest {
+
+    @get:Rule val rule = createAndroidComposeRule<ComponentActivity>()
+
+    private val title = "Auto freeze"
+    private val subtitle = "Freeze apps on a schedule"
+
+    private var reported = mutableListOf<Boolean>()
+
+    private fun setRow(checked: Boolean, enabled: Boolean = true) = rule.setContent {
+        SettingsSwitchRow(
+            icon = R.drawable.frozen,
+            title = title,
+            subtitle = subtitle,
+            checked = checked,
+            enabled = enabled,
+            onCheckedChange = { reported += it }
+        )
+    }
+
+    private val hasSwitchRole = SemanticsMatcher.expectValue(
+        SemanticsProperties.Role,
+        Role.Switch
+    )
+
+    /**
+     * The row is a switch and it is on — the two facts the old markup lost between the `clickable`
+     * that had no state and the `clearAndSetSemantics` that threw the state away.
+     */
+    @Test fun theRowAnnouncesItselfAsASwitchAndSaysWhatItIsSetTo() {
+        setRow(checked = true)
+
+        rule.onNodeWithText(title).assert(hasSwitchRole).assertIsOn()
+        rule.onNodeWithText(subtitle).assertIsOn()
+    }
+
+    /** And off, so that "on" above is a reading rather than a constant. */
+    @Test fun anUncheckedRowSaysOff() {
+        setRow(checked = false)
+
+        rule.onNodeWithText(title).assert(hasSwitchRole).assertIsOff()
+    }
+
+    /**
+     * The setting is offered once.
+     *
+     * The `Switch` used to carry its own toggle semantics, which is why it was silenced by hand; a
+     * null handler is what silences it now, and this is the assertion that the substitution held.
+     * Two toggleable nodes means a screen reader walks past the same setting twice and reads the
+     * title on only one of them.
+     */
+    @Test fun thereIsExactlyOneControlToFind() {
+        setRow(checked = true)
+
+        assertEquals(
+            1,
+            rule.onAllNodes(isToggleable()).fetchSemanticsNodes().size
+        )
+    }
+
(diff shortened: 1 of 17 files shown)


trinadhthatakula:
All three review findings are fixed in `3bf14f45`. Each was real; verifying them turned up a bit more than the comments said.

**1. Back stack (Major) — confirmed, and the premise was the part I had wrong.** I had assumed a category could only be replaced by another category. It cannot: `ExtensionManager` is itself a `detailPane` (`MainScreen.kt:628`), so on a wide window the index stays visible *beside* it and the user can pick a second category while the top of the stack is `ExtensionManager`. The guard fell through to `add`, and Back then walked out through the extension manager and the category visited before it — the exact property the comment above it claims to guarantee. Truncating everything above index 0 is the only spelling of "replace" that holds for every route a category can open. Index 0 is `ThorRoute.Settings` by construction (`rememberNavBackStack`, `MainScreen.kt:179`).

**2. The two access rows (Minor — undersold, they were dead controls).** Correct: both handlers asked whether the permission was currently *missing* and returned if it was not, so once granted the switch bounced back and nothing happened, with no in-app route to the toggle the row is a picture of. Usage access now deep-links in both directions — the screen that grants it is the screen that revokes it. Notifications keeps the system dialog for the one case it can serve (not granted, 33+) and deep-links otherwise, since re-requesting a permission you already hold returns granted without showing anything.

**3. Switch semantics (Major) — confirmed, and worse than "expose the role".** The `Switch` had `clearAndSetSemantics {}` on it to stop the setting being offered twice, and the row underneath was only `clickable`. Between them *nothing* carried the checked state: a screen reader user could change any of Thor's toggles without ever hearing what it was set to. The row is `toggleable(role = Role.Switch)` now.

One deviation from the suggested fix: rather than `onCheckedChange = null` *plus* the existing `clearAndSetSemantics`, the null handler is doing the whole job — Material3 only attaches `toggleable` semantics to `Switch` when a handler is present, so there is no second announcement left to suppress. It also stops the Switch consuming the pointer, which is what lets a tap landing on the thumb reach the row's toggleable.

I did not add the suggested semantics test. There is no Compose UI test harness on this module — `app/src/androidTest` has no Compose rule and unit tests run on the JVM without Robolectric — so it would mean standing one up, which is worth doing but not inside this PR.

Re-verified after the fixes: `compileFossDebugKotlin` + `compileStoreDebugKotlin` clean, `lintFossDebug` 0 errors / 0 warnings, `testFossDebugUnitTest --rerun-tasks` 1523 tests / 0 failures.

trinadhthatakula:
## Correction, and the desk-only disclosure is now retired

Two things in my last comment were wrong, and fixing the first one made the second one fixable.

**I said this module has no Compose UI test harness. It does** — `ui-test-junit4` is wired up and three tests already use it (`BentoTileTest`, `HomeActionsBentoTest`, `AppActionRowTest`). I checked `app/build.gradle.kts` for a `createComposeRule` string, found nothing, and stopped. So the semantics test the review asked for is written and passing: **`SettingsSwitchRowTest`, 7 tests, green on a Pixel 10 Pro Fold AVD (API 37)**.

**And the PR body says the feature is desk-verified only. That is no longer true.** There was a fold emulator attached the whole time. Driven by hand at 851 dp unfolded and 411 dp folded:

| Claim in the PR body | Result |
|---|---|
| Two-pane path is untested and a wrong scene key gives a silent blank pane | **Refuted.** Index left, category right, both populated. `SettingsDetailPlaceholder` renders before a selection and the index row highlights green once one is made. |
| — | Search deep-link works end to end: "amoled" → result → Appearance opens in the detail pane **scrolled to AMOLED Mode with the row lit**. |
| — | Folded to 411 dp: single pane, placeholder correctly **absent**, category takes the full window, Back returns. |

### The back-stack finding, reproduced and then fixed on hardware

Worth recording because the premise was the part I got wrong. Tapping **Extensions → Manage Extensions** does leave the index visible on the left with Extension Manager in the right pane — so the user really can pick a second category while the top of the stack is not a `SettingsCategory`, and the old guard appended.

With the fix, from that exact state: tap **Freezer** → Freezer fills the detail pane → **one Back lands on the bare index** (`Pick a category to see its settings`), not on Extension Manager → a second Back leaves Settings for Home. Stack is `[Settings]`, which is the property the comment claims.

### The dead-switch finding, likewise

The emulator happened to have usage access **granted**, which is precisely the state that was broken. Toggling it off now opens `com.android.settings.Settings$AppUsageAccessSettingsActivity` showing *Thor / 1.94.1 / Permit access to app usage data* — the screen where it is revoked. Before the fix that tap did nothing at all.

### What is still not verified

- **The fr / es / ar / zh-rCN translations are still machine-quality and still want a native reader.** Nothing above touches that.
- Only the `foss` flavour was run on a device; `store` is compile-verified only.
- One emulator, one API level (37). No physical-device pass, no RTL pass — worth a look at Arabic given the two-pane layout is new.
- Screenshots of both panes are in the run if you want them; say the word and I will attach.

Full gate state after `9892cfae`: both flavours compile, `lintFossDebug` 0 errors / 0 warnings, `testFossDebugUnitTest` **1523 / 0 failures**, `connectedFossDebugAndroidTest` for the new class **7 / 0 failures**.

PR #384, opened 2026-08-12
chore(release): 1942 — backup and restore, in its first test phase

trinadhthatakula:
Bumps `versionCode` to **1942** (`versionName` 1.94.2, derived) and writes the notes for the **ten** merges since v1.94.1. Bump and notes in one commit, as `release-notes/README.md` requires.

> ⚠️ **Merge this one last, and only on the owner's say-so.** It is the only push in the release that publishes anything. The dev rung passes `on_unchanged_version: skip`, so every *other* merge into `dev` builds for verification and uploads nothing; this is the push that carries the version-code bump and therefore uploads 1942 to Play alpha. Play allows one upload per version code per app, so it is not repeatable. #385, #386 and #387 are already merged and cost nothing to have merged early.

## What is in the release

| PR | |
|---|---|
| #379, #381, #380 | app backup and restore — part of #51 phase 2 |
| #376, #378 | game data inside a `.xapk` — part of #164, plus the hardware fix |
| #383 | Settings: eight categories, search, and a second pane |
| #385 | the batch actions stop claiming work they did not do; the job seam generalises |
| #386 | the subscription sweep runs on the paths where it was being skipped |
| #387 | self-granted permissions, a job's notification ask, and an export that stops deleting first |
| #377 | maven group ×4, AGP alpha08, Gradle 9.7.0 |

`origin/dev` is merged into this branch (`548f1308`) so the PR's own CI builds what the release will actually contain, rather than a tree three merges behind it.

## Folded in after the notes were first written

#385, #386 and #387 all landed after the original notes were drafted, and all three are user-visible, so they are written up under *What's Changed* rather than tucked into Internal. `github.md` records what is **not** a win, as well:

- the background-jobs notification-channel rename is **English-only**;
- the one new string (`error_self_skipped`) was **translated into four locales without a native reader**, and the notes say so;
- the second job queue and the after-the-fact completion notification **still have no producer** — verified, not assumed: `THOR_SWEEP_CHAIN` has no enqueue site anywhere in `app/src/main/java`;
- **none of the three has been on a device**, and #386 has no JVM seam available to it *at all* — the billing library is a `storeImplementation` dependency and the only unit-test task is `testFossDebugUnitTest`, so that file is on no test classpath by construction. It is verified against billing 9.1.0 bytecode instead, and the notes say that rather than implying test coverage.

Every string and behaviour claim quoted in the notes was checked against the source (`tile_unfreeze_partial_failure`, `log_stopped`, `error_self_skipped`, the two `processed < …` gates, both `forceUnfreeze` bulk call sites), not carried over from a summary.

## The test-phase caveat, in all three channels

Per the instruction, **backup & restore is named as being in its first test phase in every channel** — `playstore.txt`, `telegram.md` and `github.md`. `github.md` gives the reason rather than just the label:

- Nothing that replaces an app's data has run inside a test. That region is a WorkManager job driving privileged shell against another app's data directory; there is no JVM seam in it. The 1,548 unit tests cover the container format, the KDF, the gates, the sizing and the view models — none of them writes to a real app's data directory.
- Both Criticals the review found were *"the backup cannot be restored"*, and both surfaced only once `0e9fef25` finally wrote an archive with the real backup path and read it back with the real restore path.
- Backup only reads, so a failed backup cannot damage its source. **Restore is the destructive half** and that is what the notes tell people to be careful with.

The test count is corrected **1,523 → 1,548**, parsed from `app/build/test-results/**/*.xml` with ElementTree on this exact tree (112 classes, 0 failures, 0 errors, 0 skipped) rather than read off a Gradle log line.

## Budget arithmetic

`telegram.md` and `playstore.txt` both had less headroom than the new copy needed. Existing bullets were **tightened rather than the new ones dropped** — the alternative was a release that fixes a refund bug and a batch that lied, and mentions neither where users read.

| file | before | after | cap |
|---|---|---|---|
| `telegram.md` | 774 | **811** in-file UTF-16 units | 989/1024 against the worst-case 178-unit caption wrapper (35 spare); 963 on the dev rung |
| `playstore.txt` | 437 | **449** characters | 500 hard, inside the recommended 440–490 band |

Sized with the same UTF-16 arithmetic `check-notes-budget.sh` uses, at both the dev-rung wrapper (152) and the long-branch worst case (178) — because the wrapper is not a constant and a file that fits on `dev` can be refused from a `workflow_dispatch`.

## Step 6 of the guide, all green

```
check-notes-budget.sh 1.94.2 152   exit 0   telegram 963/1024 units (61 spare)
                                            playstore 449/500 characters
check-notes-budget.sh 1.94.2 178   exit 0   telegram 989/1024 units (35 spare)
closing-keyword grep               clean    (nothing would auto-close on the master PR)
diff vs both fastlane locales      identical
check-shizu-manifest.sh            exit 0   all checks passed
run-tests.sh                       exit 0   10 files, 0 failed (incl. locale parity)
:app:compileFossDebugKotlin        BUILD SUCCESSFUL
:app:compileStoreDebugKotlin       BUILD SUCCESSFUL
:app:testFossDebugUnitTest         1548 tests, 0 failures, 0 errors, 0 skipped
```

Both fastlane changelogs were **re-copied** from `playstore.txt`. `copy_playstore_notes` overwrites `en-US` only and skips a non-empty `hi-IN` as *"kept (translated)"* — and `hi-IN/1942.txt` is still the English placeholder, so leaving it alone would have shipped the pre-edit text to Play and F-Droid with every check green.

`shizu_store.json` is deliberately **untouched**. It derives its version from `origin/production` (still 1940), and the guide moves that sync to after the production promotion — running it here would silently re-sync the *last production* changelog and exit 0. The checker asserting against v1.94.0 above is that expected state, not drift this PR caused.

## Staging

Explicit paths only. Left out on purpose: the uncommitted `gradle/wrapper/gradle-wrapper.properties` edit in the working tree (a local `networkTimeout`/`retries` change, not mine) and the untracked `docs/discussions/`.

## Housekeeping in the same session

Stale branches cleaned: `docs/app-backup-spec`, `docs/retract-workmanager-throttle-in-plan`, `feat/backup-sheet-progress-states`, `feat/settings-categories` — each verified as an ancestor of `dev` and unreferenced in `.github/` first. `master`, `dev` and `production` untouched.

## Summary by CodeRabbit

* **New Features**
  * Added encrypted app and data backup and restore support with safeguards.
  * Added `.xapk` support for games with additional data.
  * Redesigned Settings with search and two-pane layouts.
  * Added controls for remembering and managing Thor backup passphrases.
* **Bug Fixes**
  * Improved unfreeze and batch-operation reporting.
  * Improved subscription confirmation reliability.
* **Documentation**
  * Added v1.94.2 release notes and localized changelogs.
* **Release**
  * Updated the application version to 1.94.2.

### DIFF
--- a/fastlane/metadata/android/en-US/changelogs/1942.txt
+++ b/fastlane/metadata/android/en-US/changelogs/1942.txt
@@ -1,7 +1,9 @@
 • 💾 Back up an app and its data to one encrypted file, and restore it later. Backup & restore is in its first test phase — try it on an app you can afford to lose.
 
-• 🎮 Games with extra data now export and install as a single .xapk, game data included.
+• 🎮 Games with extra data export and install as one .xapk.
 
-• ⚙️ Settings now has eight sections and a search box, and opens two panes on a tablet or an unfolded phone.
+• ⚙️ Settings now has eight sections, a search box, and two panes on a tablet.
 
 • 🔑 Thor can remember your backup passphrase, or forget it when you ask.
+
+• 🧊 Unfreeze now unpauses too, and batches report what really worked.
--- a/fastlane/metadata/android/en-US/changelogs/1942.txt
+++ b/fastlane/metadata/android/en-US/changelogs/1942.txt
@@ -1,0 +1,7 @@
+• 💾 Back up an app and its data to one encrypted file, and restore it later. Backup & restore is in its first test phase — try it on an app you can afford to lose.
+
+• 🎮 Games with extra data now export and install as a single .xapk, game data included.
+
+• ⚙️ Settings is eight sections with a search box, and opens two panes on a tablet or an unfolded phone.
+
+• 🔑 Thor can remember your backup passphrase, or forget it when you ask.
--- a/fastlane/metadata/android/en-US/changelogs/1942.txt
+++ b/fastlane/metadata/android/en-US/changelogs/1942.txt
@@ -2,6 +2,6 @@
 
 • 🎮 Games with extra data now export and install as a single .xapk, game data included.
 
-• ⚙️ Settings is eight sections with a search box, and opens two panes on a tablet or an unfolded phone.
+• ⚙️ Settings now has eight sections and a search box, and opens two panes on a tablet or an unfolded phone.
 
 • 🔑 Thor can remember your backup passphrase, or forget it when you ask.
--- a/fastlane/metadata/android/hi-IN/changelogs/1942.txt
+++ b/fastlane/metadata/android/hi-IN/changelogs/1942.txt
@@ -1,7 +1,9 @@
 • 💾 Back up an app and its data to one encrypted file, and restore it later. Backup & restore is in its first test phase — try it on an app you can afford to lose.
 
-• 🎮 Games with extra data now export and install as a single .xapk, game data included.
+• 🎮 Games with extra data export and install as one .xapk.
 
-• ⚙️ Settings now has eight sections and a search box, and opens two panes on a tablet or an unfolded phone.
+• ⚙️ Settings now has eight sections, a search box, and two panes on a tablet.
 
 • 🔑 Thor can remember your backup passphrase, or forget it when you ask.
+
+• 🧊 Unfreeze now unpauses too, and batches report what really worked.
--- a/fastlane/metadata/android/hi-IN/changelogs/1942.txt
+++ b/fastlane/metadata/android/hi-IN/changelogs/1942.txt
@@ -1,0 +1,7 @@
+• 💾 Back up an app and its data to one encrypted file, and restore it later. Backup & restore is in its first test phase — try it on an app you can afford to lose.
+
+• 🎮 Games with extra data now export and install as a single .xapk, game data included.
+
+• ⚙️ Settings is eight sections with a search box, and opens two panes on a tablet or an unfolded phone.
+
+• 🔑 Thor can remember your backup passphrase, or forget it when you ask.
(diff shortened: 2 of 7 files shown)


trinadhthatakula:
Two commits since the description was written; the numbers in it are one edit stale.

**`aa552d5a` — a break this PR surfaced rather than caused.** `web` went red on open: `site.test.ts` asserts the site's Telegram handle matches what the app links to, and reads `SettingsScreen.kt` to find it. #383 split that screen into eight categories and the link went with About & support, so the assertion had been reading a file with no `t.me` in it since that merge.

It stayed green because **`web-ci` is path-filtered to `web/**`, `gradle.properties` and `libs.versions.toml`** — no Kotlin change can trigger it. A release PR bumping `versionCode` is the only thing that reliably schedules the job, so this class of break surfaces one release late, on someone else's PR. The path is fixed and the comment now says why a stale one is invisible.

**`5aae5701` — the review's one finding.** "Settings is eight sections" swept to "Settings now has eight sections" across all three files and both fastlane copies, not just the quoted line — the same sentence was in the `github.md` highlights.

Gates re-run after both:

```
check-notes-budget.sh 1.94.2   exit 0   telegram 934/1024 units (90 spare)
                                        playstore 437/500 characters   (was 433)
run-tests.sh                   exit 0   10 files, 0 failed
closing-keyword grep           clean
fastlane copies                byte-identical to playstore.txt
web unit tests (local)         296/296
```

All required checks green, `build-and-test` included.

PR #389, opened 2026-08-13
feat(export): run a single-app export on WorkManager, and release 1943

trinadhthatakula:
## What this is

Two things in one branch, because the second exists to ship the first:

1. **Single-app export moves onto the job seam** — a WorkManager job behind the same `dataSync` foreground service the archive jobs use, so you can start an export and leave.
2. **Release 1943**, carrying v1.94.2's notes forward. v1.94.2 was built and **never published**: the dev rung failed at `client.commit_current_edit!` with *"You must let us know whether your app uses any Foreground Service permissions"*, and that rejection is atomic — no `v1.94.2` tag, no GitHub release, no broadcast on any channel. Code 1942 was then spent on a hand-uploaded internal build, so it cannot be reused.

Per the owner's instruction this ships from the feature branch rather than a separate `chore/release-1943` branch.

## Why export, specifically

Play requires a video declaration for `FOREGROUND_SERVICE_DATA_SYNC`, and a reviewer can only verify what they can reproduce on a stock device. Backup and restore need root or Shizuku on **every** path, so a video of either is unverifiable by the person who has to approve it. Export needs neither privilege. It is the surface that exists to be filmable — that reasoning is now written into `docs/workers/README.md` rather than living in a chat log.

## What changes for a user

* **You can leave.** The running frame has a **Background** button; dismissing the sheet drops the screen's watcher and the export carries on. Reopening finds the same job running rather than a fresh form.
* **It reports either way.** Progress in the notification, outcome in a second notification — *"Saved <app> to Downloads"*, or the real reason it failed, with the builder's own wording preserved so *"about 1.4 GB more is needed"* survives to the shade instead of flattening to "Export failed". Previously the result was a toast, which only arrived if Thor was still foregrounded when a 4 GB game finished packaging.
* **A second tap cannot eat the first export.** Both runs stage through one folder the builder wipes on entry, so a duplicate would have deleted the other's half-written copy and finished a truncated zip *successfully*. The sheet reattaches instead of re-offering the form.
* **The form disappears while a job runs** rather than sitting greyed out above the only thing that is moving.

## Three things deliberately **not** done

* **Multi-app export stays a plain coroutine.** "Backup all" is unchanged and its non-adoption is argued in `BackupRunner`'s own KDoc. This branch moves the single-app path only.
* **A WorkManager-stopped export reports nothing from the worker.** WorkManager stops workers for its own reasons and then re-runs them; a shade row saying "nothing was saved" would be a lie told at the moment the export was about to succeed. A genuine cancel is worded by the screen.
* **`sheetTarget` is null for `APP_EXPORT`.** Tapping the progress notification opens Thor, not a sheet — the export sheet is per-app-info and there is no route that reconstructs it from a package name alone.

## The race this exposed, and `LaunchSweepBarrier`

`APP_EXPORT` is the first job on this seam that genuinely **completes** on a re-run. Both archive workers refuse one, because their encryption key is process memory; an export's whole input is four strings that are still true tomorrow.

That falsifies an assumption the launch-time cleanup was built on. `ArchiveOrphanSweeper.sweep()` `deleteRecursively()`s `externalCacheDir/obb_out` wholesale, reasoning that anything using it "died with the process" — which stops being true the instant WorkManager re-runs a worker **in** that process. A one-shot `CompletableDeferred`, completed from a `finally` in `ThorApplication`, holds an export back until the sweep is over; on timeout (120 s) it **refuses to stage** rather than proceeding.

Stated as a gap rather than glossed: **the barrier is opt-in.** A future job that stages into the swept tree and forgets to take it loses the same race, silently. That is written into the doc's known-gaps section and into the "adding a new job kind" checklist.

## Docs

`docs/workers/README.md` opened by saying *two* operations use WorkManager. There are three. Beyond the arithmetic, two of its claims were true only of a seam carrying archive work:

* *"Being on WorkManager here does not mean durable"* now holds for **two of the three**, and the doc names which side each kind sits on.
* *"Export" means two different things in this repo* — a warning up top says to read the kind, not the verb.

Also folded in, because a doc commit is the only honest place for it: `JobPhase.kt` and `JobPhaseTest.kt` carried the **retracted** *"the launcher does not await `enqueue()`"* comment, which I copied out of the backup side four commits earlier. `enqueueUniqueJob` has awaited its `Operation` since it grew `awaitSuccess()`; the doc has a callout counting **four** stale copies and I had quietly made it six. Both now give the reason that is actually load-bearing — a null `WorkInfo` is "no row for this id" as much as "the row was pruned", and only the order tells them apart. The four remaining sites are now listed by file and line.

⚠️ `CLAUDE.md` says "exactly two operations" and is **not** in this branch: it is gitignored (`.gitignore:35`) and has never been tracked, so its copy of the claim cannot be corrected from the repo. `docs/workers/README.md` is the copy to trust.

## Release 1943

`github.md` keeps every v1.94.2 section intact — they are still news to everyone, since nobody received them — and opens by saying what happened to the release they were written for. Added: the export section, a Highlights bullet, #388 in Internal (the dev rung now assigns its one upload to `internal` as well as `alpha`), and two corrections to Internal claims that went stale inside the same release.

`sync-shizu-changelog.sh` is deliberately **not** run: it resolves `versionCode` from `origin/production`, so during prep it syncs the last *production* release, prints "already current" and exits 0 while the new version never lands. `shizu_store.json` is untouched and still pins 1940, which `check-shizu-manifest.sh` confirms is correct.

## Verification

| Check | Result |
|---|---|
| `:app:testFossDebugUnitTest --rerun-tasks` | **1579 tests, 0 failures** across 116 suites (counts parsed from `test-results/**/*.xml`) |
| `check-notes-budget.sh 1.94.3` | playstore **484/500** chars; telegram **976/1024** caption units |
| `.github/scripts/test/run-tests.sh` | **10 files, 0 failed** — locale parity among them |
| `check-shizu-manifest.sh` | all checks passed |
| `1943.txt` in both locales | byte-identical to `playstore.txt` (a locale missing its file does not fall back — it blanks that locale's what's-new) |
| Closing keywords in notes | none (`master` is the default branch, so a `dev` → `master` body would act on them) |

New JVM coverage on this branch: `AppExportRequestTest` (10), `JobPhaseTest` (10), `LaunchSweepBarrierTest` (6), `AppExportWorkerReasonTest` (4) — the `Data` round trip, the status reduction the screen renders, the startup barrier, and the failure wording.

**Not on a device.** The parts that most want one are exactly the parts no JVM test can stand in for: the foreground service, the platform file APIs, and the notification behaviour across a real process death. Filming the declaration video is the next step and will exercise all three.

## Summary by CodeRabbit

* **New Features**
  * Added background single-app exports with APK and XAPK support.
  * Export to Downloads or custom folders continues while the app is backgrounded.
  * Added progress tracking, completion notifications, cancellation, retry, and actionable failure messages.
  * Export screens reconnect to active jobs after reopening.
  * Added detailed export status messages in supported languages.

* **Bug Fixes**
  * Improved startup cleanup coordination to prevent exports from beginning prematurely.

* **Documentation**
  * Updated worker documentation and release notes for background exports.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/ThorApplication.kt
+++ b/app/src/main/java/com/valhalla/thor/ThorApplication.kt
@@ -15,6 +15,7 @@ import com.rosan.dhizuku.api.Dhizuku
 import com.valhalla.bypass.Bypass
 import com.valhalla.thor.core.ThorShellConfig
 import com.valhalla.thor.data.backup.ArchiveOrphanSweeper
+import com.valhalla.thor.data.backup.job.LaunchSweepBarrier
 import com.valhalla.thor.data.permission.SelfPermissionGranter
 import com.valhalla.thor.data.service.AutoFreezeManager
 import com.valhalla.thor.data.source.local.dhizuku.DhizukuHelper
@@ -190,6 +191,14 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
     private val archiveOrphanSweeper: ArchiveOrphanSweeper by inject()
 
     /**
+     * Released once the sweep below is over, and the only thing standing between a re-run export and
+     * the sweep deleting its staging tree. Resolved here rather than injected into the worker alone,
+     * because a barrier only works if both halves hold the *same* instance — which is what `@Single`
+     * guarantees and what a worker-side `LaunchSweepBarrier()` would silently not.
+     */
+    private val launchSweepBarrier: LaunchSweepBarrier by inject()
+
+    /**
      * Resolved here because a Koin `@Single` nobody injects is never constructed, so its observer
      * would never start — the same reason [autoFreezeManager] is resolved eagerly.
      *
@@ -308,13 +317,24 @@ class ThorApplication : Application(), SingletonImageLoader.Factory {
             // launch that cannot delete a stale temp file is still a launch. The interruption warning
             // the report may carry is the restore screen's to show; Application has no UI, so the
             // report here is a log line and the breadcrumb is left standing for Task 17 to clear.
-            runCatching { withContext(ioDispatcher) { archiveOrphanSweeper.sweep() } }
-                .onFailure { throwable ->
-                    // Same rethrow as above: runCatching swallows CancellationException, and
-                    // appScope.cancel() in onTerminate must not be logged as a sweep failure.
-                    if (throwable is CancellationException) throw throwable
-                    Logger.e("ThorApp", "archive orphan sweep failed", throwable)
-                }
+            try {
+                runCatching { withContext(ioDispatcher) { archiveOrphanSweeper.sweep() } }
+                    .onFailure { throwable ->
+                        // Same rethrow as above: runCatching swallows CancellationException, and
+                        // appScope.cancel() in onTerminate must not be logged as a sweep failure.
+                        if (throwable is CancellationException) throw throwable
+                        Logger.e("ThorApp", "archive orphan sweep failed", throwable)
+                    }
+            } finally {
+                // AppExportWorker blocks on this before it stages anything, because the sweep above
+                // deletes `externalCacheDir/obb_out` wholesale and a worker WorkManager re-ran at
+                // process start is writing into that very tree. See LaunchSweepBarrier.
+                //
+                // In a `finally`, so the release survives a sweep that threw and a scope that was
+                // cancelled: neither is a reason to leave an export pinned on "Preparing" until the
+                // barrier's own timeout, and a sweep that failed has still stopped deleting.
+                launchSweepBarrier.markSwept()
+            }
         }
 
         // Below API 33 nothing else notices a language change: the platform has no per-app locale to
--- a/app/src/main/java/com/valhalla/thor/data/backup/job/AppExportWorker.kt
+++ b/app/src/main/java/com/valhalla/thor/data/backup/job/AppExportWorker.kt
@@ -1,0 +1,179 @@
+// SPDX-FileCopyrightText: 2025-2026 Trinadh Thatakula <github.com/trinadhthatakula/Thor>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package com.valhalla.thor.data.backup.job
+
+import android.content.Context
+import androidx.annotation.StringRes
+import androidx.work.WorkerParameters
+import com.valhalla.thor.R
+import com.valhalla.thor.domain.model.AppExportRequest
+import com.valhalla.thor.domain.model.EXPORT_LABEL_KEY
+import com.valhalla.thor.domain.model.ThorJobKind
+import com.valhalla.thor.domain.model.ThorJobProgress
+import com.valhalla.thor.domain.model.ThorJobStage
+import com.valhalla.thor.domain.repository.AppBundleFileStore
+import com.valhalla.thor.domain.repository.AppRepository
+import com.valhalla.thor.domain.usecase.ExportAppUseCase
+import com.valhalla.thor.domain.usecase.ExportSession
+import com.valhalla.thor.util.Logger
+import org.koin.android.annotation.KoinWorker
+
+private const val TAG = "AppExportWorker"
+
+/**
+ * One app packaged as `.apk`/`.apks`/`.xapk` and written where the user asked, behind a foreground
+ * service.
+ *
+ * The whole job is one call into [ExportAppUseCase.exportInto]; what this class actually owns is
+ * everything around it — the four strings the request arrived as, the `AppInfo` re-resolution, a
+ * destination check that fails before the packaging rather than after it, and **an outcome sentence
+ * on every terminal path**, which is what the notification the user sees at the end is made of.
+ *
+ * ### It does not open its own session
+ *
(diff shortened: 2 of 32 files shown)


trinadhthatakula:
Worked through all seven, each verified against the code on this branch rather than taken on the summary. Three fixed, three declined with reasons, one noted for a device pass. Pushed as `e077f942`, `217ec7ae`, `427333ba`.

## Fixed

**1. `watching` survives a failed collector — `ExportViewModel.kt`** (`e077f942`)

Real, and the mechanism is worse than described. `watch`'s `onFailure` cannot null the field its own `launchGuarded` is still being assigned into: `viewModelScope` is `Dispatchers.Main.immediate`, so a collector that throws runs to completion *inside* the `launch` call — the handler fires, and then the assignment parks a dead `Job` in `watching`. Nulling the field from `onFailure`, as suggested, is therefore the one fix that would not work.

Guarded on liveness instead: `if (id != null && watching?.isActive != true) watch(id)`. A live watcher keeps `isActive` true, so every state that has one is unchanged; only the dead-`Job` state differs. The two archive view models never had this bug — both settle through `finish(result = null)`, which nulls the field. Export was the copy that dropped it.

Added `ExportViewModelTest` — there was no test for this view model at all. On the old guard the reattach test fails with the second job's id missing from the launcher's `status` calls; the other two pass on both guards, which is the point, since they pin the invariant the original `== null` was protecting.

**2. Arabic `export_job_cleanup_busy`** (`217ec7ae`) — fixed on line 548 only. Left the other transliteration in `values-ar`: it arrived in an unrelated commit over copy this branch does not own.

One trap for whoever does sweep it: `rg 'ثور'` over that file reports nine hits and **seven are false**. ث-و-ر is a substring of العثور, the ordinary word for "finding". Anchor the pattern or a sweep corrupts seven unrelated strings.

**5. "unblocks a `store` release"** (`427333ba`) — reworded. Category and severity are both off (this is release-note prose, not Security & Privacy, and the following clause was already future tense), but the sentence did read as if the block had lifted, so it now says what remains: the video still has to be recorded and the declaration submitted.

**Nitpick 1: the `RequestNotificationsWhenJobStarts` comment** (`427333ba`) — correct catch, and I confirmed it independently. `start` sets `running = true` synchronously at `ExportViewModel.kt:92`, before `openSession` and before `startExport` returns an id, so the prompt fires on the tap. Comment corrected; trigger left alone.

## Declined

**3. Durability overclaim (`docs/workers/README.md`, `github.md`).** The document defines the axis three lines above the sentence, and the sentence is its exact antonym: `README:106-107` — "a reboot or an OOM kill does **not** resume either archive job. WorkManager re-runs the worker, and **the worker refuses**", under a heading that reads "Being on WorkManager here does **not** mean durable — for two of the three". The binary is *refuses-at-the-door* vs *runs*, and `:114`'s own next clause fixes the referent: "It is the first job on this seam that **does**" — does *run*, not *succeed*.

Three of the four failure paths cited are not durability facts. App-uninstalled, folder-grant-gone and disk-shortfall all sit in `runJob`, so they apply identically to the **first** run in the original process — enqueue and execution are separated by WorkManager either way. The fourth, `awaitSwept`, is circular as counter-evidence: `LaunchSweepBarrier` exists *because* the re-run reaches the builder.

Both documents also already word the failures. `README:291-294`: "every failing `return` in `runJob` goes through `failNoted` … There is no path out of that worker that ends in silence." And `github.md:99-101`, twenty-four lines *above* the sentence called an overclaim, states the failure sentence reaching the shade verbatim.

**4. Qualify the notification claims (6 files).** The gate is real — `ThorJobNotifications.kt:110` and `:145` both open with `areNotificationsEnabled()`, and it does cover app-wide mute. But the finding cites `github.md:49-50` and `:98-101` while the same file discloses the dependency at `:367-370`: "**A job that starts without notification access now asks for it**, at the moment the job starts, from the sheet that started it… As of this release the export sheet asks the same question, for the same reason." That is a read of part of a document presented as a read of the document.

The app also closes the gap: `ExportBottomSheet.kt:145` asks at enqueue time, and `update`/`postResult` re-check per call, so a mid-job grant restores both surfaces. `v1.94.2` shipped the identical unqualified claim for backup and restore. And the remedy does not fit three of the five files it names — `playstore.txt` is at 484/500 against a hard CI cap, and the shortest usable qualifier costs 26.

**Nitpick 2: TalkBack focus when the form is swapped for `JobRunningFrame`.** Not verifiable without a device, and I am not going to guess at a focus fix I cannot observe. Genuinely worth checking — it goes on the device-test list with the FGS recording, since both need the same stock handset.

## Not raised, worth recording

`ThorJobNotifications` checks only `areNotificationsEnabled()` and never channel importance, unlike `BulkResultNotifier.kt:48`. Silencing the **Background jobs** channel passes the gate and still shows nothing, and the permission ask cannot detect that state either. The KDoc at `ThorJobNotifications.kt:137-139` argues that is deliberate, so this is a follow-up line rather than a fix — but it is the sharp version of finding 4, and the one the finding did not name.

---

Validation: `:app:testFossDebugUnitTest --rerun-tasks`, 117 suites / **1582 tests / 0 failures**. Note-budget check re-run unchanged (playstore 484/500, telegram 976/1024); both fastlane `1943.txt` copies still byte-identical to `playstore.txt`.

PR #417, opened 2026-08-20
fix: package visibility, dynamic privilege acquisition, and sort touch fixes

trinadhthatakula:
## Summary

This PR addresses package visibility collapse on fresh installs, privilege manager detection/acquisition (KernelSU, Shizuku, Dhizuku), and UI touch responsiveness in the App List bottom sheet.

### Key Changes
1. **Package Visibility & Permissions Recovery**:
   - Resolved package visibility collapse on fresh installs and Chinese OEM ROMs (ColorOS, OxygenOS, HyperOS, MIUI) by applying fallback flag queries and automatic AppOps grants.
   - Synchronized `SelfPermissionGranter` with `AppRepository` rescans.

2. **Privilege Acquisition & Dynamic Detection**:
   - Added `PrivilegeManagerApp` registry with package name resolution for known root and privilege managers (KernelSU, KernelSU-Next, APatch, Magisk, Shizuku, Dhizuku, etc.).
   - Added inline in-app permission grant requests for Shizuku and Dhizuku.
   - Removed `FLAG_MOUNT_MASTER` to prevent KernelSU / APatch root failure.
   - Added automatic privilege re-probing in `HomeActivity.onResume()` so switching back from KernelSU/Magisk manager auto-acquires root without app restart.
   - Made `HomeViewModel` reactively reload data on any privilege availability transition.

3. **App List Sort & Filter Touch Responsiveness**:
   - Replaced nested fixed-height `LazyColumn(200.dp)` in `AppFilterSheet` with a unified scrollable column to eliminate touch slop/gesture collisions in `ModalBottomSheet`.
   - Added automatic `scrollToItem(0)` when sort, order, filter, or search query changes so the list always resets to the top.

### Verification
- Ran `./gradlew test lintFossDebug lintStoreRelease`: **PASSED (0 errors, 0 lint warnings)**
- Ran test suite: **ALL PASSED**

## Summary by CodeRabbit

* **New Features**
  * Added detection and management controls for installed privilege-manager apps.
  * Added options to request Shizuku or Dhizuku access and open manager apps.
  * Enhanced app-list filtering, sorting, and scroll-position behavior.
  * Improved privilege-mode selection with tap and long-press actions.

* **Bug Fixes**
  * Improved installed-app permission handling and refreshed app scans after permission changes.
  * Improved app discovery, fallback labels, scan resilience, and compatibility across supported setups.
  * Prevented incomplete visibility scans from triggering unsafe app removal.

* **Localization**
  * Added translated “Open” and Brazilian Portuguese “Close” labels.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -18,10 +18,6 @@
         android:name="android.permission.QUERY_ALL_PACKAGES"
         tools:ignore="PackageVisibilityPolicy,QueryAllPackagesPermission" />
 
-    <permission
-        android:name="android.permission.QUERY_ALL_PACKAGES"
-        tools:ignore="ReservedSystemPermission" />
-
     <!--
       Non-AOSP. Declared by Chinese-market ROMs (MIUI/HyperOS, ColorOS, OriginOS, MagicOS)
       per T/TAF 108-2022, where package visibility is a three-state runtime permission
--- a/app/src/main/java/com/valhalla/thor/HomeActivity.kt
+++ b/app/src/main/java/com/valhalla/thor/HomeActivity.kt
@@ -329,6 +329,9 @@ class HomeActivity : ComponentActivity() {
         // unacknowledged for three days; coming back to the app is the first chance to catch it.
         // The store implementation does its work on a background scope and the foss one is a no-op.
         billingProcessor.refreshPurchases()
+        // Re-probe privileges (e.g. user went to KernelSU/Magisk/APatch manager to grant root,
+        // or authorized Shizuku/Dhizuku externally).
+        privilegeManager.refresh()
         if (hasRequestedShizuku) return
         lifecycleScope.launch {
             val privileges = privilegeManager.state.first { it.isReady }
--- a/app/src/main/java/com/valhalla/thor/HomeActivity.kt
+++ b/app/src/main/java/com/valhalla/thor/HomeActivity.kt
@@ -43,6 +43,7 @@ import com.valhalla.thor.presentation.settings.BillingProcessor
 import com.valhalla.thor.presentation.theme.ThorTheme
 import com.valhalla.thor.presentation.utils.ObserveAsEvents
 import com.valhalla.thor.util.AppLocale
+import com.valhalla.thor.util.AppScanRevision
 import com.valhalla.thor.util.Logger
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.first
@@ -107,6 +108,7 @@ class HomeActivity : ComponentActivity() {
     private val shizukuHandler = ShizukuPermissionHandler(
         onPermissionGranted = {
             Logger.d("HomeActivity", "Shizuku Ready")
+            AppScanRevision.bump()
             homeViewModel.loadDashboardData()
         },
         onPermissionDenied = {
--- a/app/src/main/java/com/valhalla/thor/HomeActivity.kt
+++ b/app/src/main/java/com/valhalla/thor/HomeActivity.kt
@@ -331,7 +331,18 @@ class HomeActivity : ComponentActivity() {
         billingProcessor.refreshPurchases()
         // Re-probe privileges (e.g. user went to KernelSU/Magisk/APatch manager to grant root,
         // or authorized Shizuku/Dhizuku externally).
-        privilegeManager.refresh()
+        //
+        // Gated on isReady, which is the difference between "probe again" and "probe twice".
+        // PrivilegeManager starts its first probe from its own init, i.e. from
+        // ThorApplication.onCreate, so on a cold start that probe is still in flight when the
+        // first onResume lands — refreshing here would bump the trigger and buy a second full
+        // root/Shizuku/Dhizuku sweep for an answer already on its way. That is the duplicate
+        // c86aa565 ("perf(privilege): one root probe per cold start") removed. isReady latches
+        // true on the first emission and never goes back, so every *later* resume — the ones
+        // where the user actually did go and grant something — still re-probes.
+        if (privilegeManager.state.value.isReady) {
+            privilegeManager.refresh()
+        }
         if (hasRequestedShizuku) return
         lifecycleScope.launch {
             val privileges = privilegeManager.state.first { it.isReady }
--- a/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt
+++ b/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt
@@ -17,13 +17,11 @@ object ThorShellConfig {
         // Set logging based on build type
         Shell.enableVerboseLogging = BuildConfig.DEBUG
 
-        // Configure the default builder.
-        // FLAG_MOUNT_MASTER: Essential for global namespace operations (mounting, etc.)
+        // Configure the default builder. Do NOT set FLAG_MOUNT_MASTER as KernelSU / APatch su
+        // does not accept --mount-master and would block root acquisition.
         Shell.setDefaultBuilder(
             Shell.Builder.create()
-                .setFlags(Shell.FLAG_MOUNT_MASTER)
-            // If you have specific initializers, add them here
-            // .setInitializers(MyInitializer::class.java)
+                .setTimeout(10)
         )
     }
 }
--- a/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt
+++ b/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt
@@ -17,11 +17,32 @@ object ThorShellConfig {
         // Set logging based on build type
         Shell.enableVerboseLogging = BuildConfig.DEBUG
 
-        // Configure the default builder. Do NOT set FLAG_MOUNT_MASTER as KernelSU / APatch su
-        // does not accept --mount-master and would block root acquisition.
+        // Do NOT set FLAG_MOUNT_MASTER — but not for the reason it is tempting to write down. The
+        // flag cannot "block root acquisition" on its own: Odin's BuilderImpl.start() wraps the
+        // `su --mount-master` attempt in `catch (_: NoShellException)` and falls through to plain
+        // `su`, then to `sh`, so an su that rejects the argument is recoverable by construction.
+        //
+        // What the fallback is not is free, and that is the part that bit KernelSU/APatch users.
+        // An su that rejects --mount-master by *exiting* is cheap — ShellImpl's shell check calls
+        // process.exitValue() first, sees a dead process and throws immediately. An su that instead
+        // sits there is not: nothing gives up until the shell check times out, and only then does
+        // the plain-`su` attempt start, from zero, with its own superuser prompt. Which of the two
+        // KernelSU and APatch actually do has not been confirmed here — the field symptom (root
+        // never acquired within any reasonable wait) fits the second. Thor never needs the global
+        // mount namespace, so there is nothing on the other side of that bet worth holding.
+        //
+        // setTimeout aligns the builder's shell-check budget with Odin's own root probe, which gives
+        // up at RealShellRepository.SHELL_INIT_TIMEOUT_MS = 10s; the 20s BuilderImpl default would
+        // leave a `su` process and a blocked executor thread alive for ten seconds after the probe
+        // has already reported "no root". The budget covers the time the *user* spends answering the
+        // superuser dialog, so the trade is real: a grant given between 10s and 20s now has its
+        // shell destroyed instead of cached, and costs one more prompt on the next probe.
         Shell.setDefaultBuilder(
             Shell.Builder.create()
-                .setTimeout(10)
+                .setTimeout(SHELL_INIT_TIMEOUT_SECONDS)
         )
     }
+
+    /** Matches Odin's `RealShellRepository.SHELL_INIT_TIMEOUT_MS`; see [init]. */
+    private const val SHELL_INIT_TIMEOUT_SECONDS = 10L
 }
--- a/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
@@ -12,9 +12,11 @@ import com.valhalla.thor.data.source.local.shizuku.SystemAppRemovalOutcome
 import com.valhalla.thor.data.source.local.shizuku.displayLine
 import com.valhalla.thor.data.source.local.shizuku.isRootOnlySystemAppRemoval
 import com.valhalla.thor.data.source.local.installCommand
+import com.valhalla.thor.data.source.local.installedAppsAppOpGrantCommands
 import com.valhalla.thor.data.source.local.pmPathCommand
 import com.valhalla.thor.data.source.local.thorUserId
 import com.valhalla.thor.domain.gateway.SystemGateway
+import com.valhalla.thor.domain.model.GET_INSTALLED_APPS_PERMISSION
 import com.valhalla.thor.domain.model.PrivilegeMode
 import com.valhalla.thor.domain.model.uninstallFreezeFallbackAllowed
 import kotlinx.coroutines.CoroutineDispatcher
@@ -464,6 +466,11 @@ class DhizukuSystemGateway(
         val escapedPermissionName = permissionName.escapeForShell()
         return try {
             val result = DhizukuHelper.execute("pm grant --user $userId $escapedPackageName $escapedPermissionName")
+            if (permissionName == GET_INSTALLED_APPS_PERMISSION) {
+                installedAppsAppOpGrantCommands(escapedPackageName, userId).forEach { cmd ->
+                    DhizukuHelper.execute(cmd)
+                }
+            }
             if (result.first == 0) Result.success(Unit)
             else Result.failure(Exception("Dhizuku: pm grant failed with exit code ${result.first}: ${result.second}"))
         } catch (e: Exception) {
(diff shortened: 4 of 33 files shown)


trinadhthatakula:
### Code review

**Model:** `claude-opus-5` (Claude Opus 5), run as a 64-agent review: 7 independent review lenses (CLAUDE.md compliance, shallow bug scan, git history, prior PR feedback, code-comment contracts, coroutines/lifecycle, Android platform behaviour) producing 40 raw findings → 19 unique after dedup → each adversarially verified by 3 independent refuters (refutation, reproducibility, scope/attribution).

**Verdict:** Approve with changes. The PR does what it says and the shape is right — the `AppScanRevision` seam, the per-user AppOp grants, the `syncCacheSucceeded` gate and the `AppFilterSheet` rewrite are all sound. Two issues in newly added code are worth fixing before merge; three further notes are lower confidence and can be judged on the device.

---

## 1. The scroll-to-top `LaunchedEffect` also fires on first composition, discarding restored scroll position

`rememberLazyListState()` / `rememberLazyGridState()` are `rememberSaveable`-backed, so Compose restores `firstVisibleItemIndex` on activity recreation and on re-entering the destination. The new effect then runs on that first composition with unchanged keys and immediately scrolls back to 0.

https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/widgets/AppList.kt#L617-L623

Repro: scroll 200 apps down the Apps list → rotate the device (`.HomeActivity` declares no `android:configChanges`), or switch to Home and back → the list snaps to the top. Before this PR the position survived.

Four sites are affected: [`AppList.kt#L591-L595`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/widgets/AppList.kt#L591-L595) (grid), [`AppList.kt#L618-L622`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/widgets/AppList.kt#L618-L622) (list), [`FreezerScreen.kt#L371-L375`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/freezer/FreezerScreen.kt#L371-L375) and [`FreezerScreen.kt#L415-L419`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/freezer/FreezerScreen.kt#L415-L419).

Suggested change — skip the initial run with a plain `remember` (deliberately **not** `rememberSaveable`: it must reset to `true` on exactly the recompositions where the saved scroll position is restored):

```diff
 val listState = rememberLazyListState()
 
-LaunchedEffect(sortBy, sortOrder, selectedFilter, filterType, appListType, searchQuery) {
-    listState.scrollToItem(0)
-}
+// `remember`, not `rememberSaveable` — on activity recreation the lazy state restores the
+// user's position, and this flag must be `true` again so the restore is not undone.
+var skipInitialReset by remember { mutableStateOf(true) }
+LaunchedEffect(sortBy, sortOrder, selectedFilter, filterType, appListType, searchQuery) {
+    if (skipInitialReset) skipInitialReset = false
+    else listState.scrollToItem(0)
+}
```

## 2. The `flags = 0` visibility fallback is blind to uninstalled-for-user packages, but still gets full pruning authority

https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt#L257-L270

The fallback result replaces `installedPackages` wholesale and then flows unchanged into `currentPackageNames`, `toDelete`, `scanVerdict(...)` and `pruneWatchlist(...)`. But a `flags = 0` query is by construction blind to packages whose per-user `FLAG_INSTALLED` bit is clear — which is exactly what the codebase says freeze rung 2 produces:

> `-k` is `DELETE_KEEP_DATA` … what it does *not* keep is the per-user installed bit, which is why a package frozen this way reads as gone to anything querying without `MATCH_UNINSTALLED_PACKAGES`
> — [`RootSystemGateway.kt#L534-L539`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt#L534-L539)

Rung 2 is unreachable in new freezes, but the same doc is explicit that this state is still live on real devices — *"Devices in the field carry apps frozen by the uninstall-only builds and must still thaw after this ships"* ([`#L547-L549`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/data/gateway/RootSystemGateway.kt#L547-L549)) — and a system app the user uninstalls for their own user through Thor lands in exactly the same place.

Repro: HyperOS device, 305 cached rows, 5 of them uninstall-frozen. The primary scan collapses to 1, the fallback returns the 300 normally-visible packages, `300 * 2 > 305` so `scanVerdict` returns `Accept` — and those 5 get their Room row deleted, their `app_icons/<pkg>.png` deleted, and their Freezer watchlist row pruned. Per [`InstalledAppsVisibility.kt#L186-L193`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/domain/model/InstalledAppsVisibility.kt#L186-L193) the watchlist row is the record that makes them thawable, so they become unthawable from Thor. This is the same class of damage the guard added in `6c010067` exists to prevent.

Note that a union of the two lists does **not** fix this — the primary scan returned 1 package, so there is nothing to union back in. The scan has to be marked non-authoritative:

```diff
+var usedVisibilityFallback = false
 if (installedPackages.size <= 1) {
     val fallback = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
         pm.getInstalledPackages(PackageManager.PackageInfoFlags.of(0L))
     } else {
         pm.getInstalledPackages(0)
     }
     if (fallback.size > installedPackages.size) {
         installedPackages = fallback
+        usedVisibilityFallback = true
     }
 }
```

```diff
-val verdict = scanVerdict(
-    scannedPackageNames = currentPackageNames,
-    cachedCount = initialCachedCount,
-    consecutiveSuspectScans = consecutiveSuspectScans,
-    permission = installedAppsPermission.state()
-)
+// A flags=0 scan cannot see FLAG_INSTALLED-clear packages, so "absent from this scan"
+// is not evidence of uninstallation. Updates still land; deletions must not.
+val verdict = if (usedVisibilityFallback) {
+    ScanVerdict.Retain(RetainReason.VisibilityFallback)
+} else {
+    scanVerdict(
+        scannedPackageNames = currentPackageNames,
+        cachedCount = initialCachedCount,
+        consecutiveSuspectScans = consecutiveSuspectScans,
+        permission = installedAppsPermission.state()
+    )
+}
```

(`RetainReason` would need a fourth constant at [`InstalledAppsVisibility.kt#L102-L111`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/domain/model/InstalledAppsVisibility.kt#L102-L111); reusing `Collapsed` would work but would misreport the reason in the log line.)

---

## Lower-confidence notes

**3. `onResume`'s unconditional `refresh()` reintroduces the duplicate cold-start probe.**

https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/HomeActivity.kt#L332-L336

`PrivilegeManager` is constructed in `ThorApplication.onCreate`, so probe #1 is already in flight when `onResume` first runs — every cold start now pays two full probes. The new call is also placed *above* the `if (hasRequestedShizuku) return` guard whose KDoc explains why `c86aa565` ("perf(privilege): one root probe per cold start") removed the second probe. Guarding it on `isPrivilegeReady`, or moving it below the early return, keeps the resume-reacquisition behaviour without the cold-start cost.

**4. The `ThorShellConfig` comment does not match Odin's behaviour, and `setTimeout(10)` interacts with the probe.**

https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt#L19-L26

Odin's `BuilderImpl.start()` wraps `exec("su", "--mount-master")` in `catch (_: NoShellException)` and unconditionally falls through to plain `exec("su")`, so `--mount-master` not being accepted does not on its own "block root acquisition" — the plausible real mechanism is that the mount-master attempt *times out* and burns the shell-init budget before the fallback runs. Worth rewording so the next reader does not act on the stated reason.

Separately, `setTimeout(10)` halves Odin's documented 20s default (`BuilderImpl.kt:13`), and that budget gates `check.get(builder.timeout, SECONDS)` in `ShellImpl`'s `init` — i.e. it covers the time the user takes to answer the superuser prompt. `RealShellRepository.isRootGranted()` already gives up at 10s, so the *probe* result is unchanged; what changes is that a grant answered at ~13s used to still complete in the background and cache a root shell for the next probe, and now has its `su` destroyed instead. The new cached-shell invalidation in `isRootAvailable()` should recover this on the following resume, at the cost of a second prompt. Worth a device check on a slow first grant.

**5. `requestDhizuku()` does blocking binder IPC on the main thread and hands Dhizuku an Activity context.**

https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/presentation/home/HomeViewModel.kt#L200-L208

The dialog's `onClick` calls straight through to [`DhizukuHelper.requestPermission`](https://github.com/trinadhthatakula/Thor/blob/b83fb664a555509e3a0979400bcc3dcb67e91459/app/src/main/java/com/valhalla/thor/data/source/local/dhizuku/Dhizuku.kt#L149-L156), whose `DhizukuAPI.init` + `isPermissionGranted()` are synchronous binder round-trips — `DhizukuSystemGateway` confines the same helper to the IO dispatcher for exactly this reason. The bind is normally already latched by `ThorApplication.onCreate`, so this is a stall rather than an ANR, but `viewModelScope.launch(ioDispatcher) { … }` costs nothing. The `context` passed from `LocalContext.current` is the Activity, and `DhizukuAPI.init` retains it in a static field for the process lifetime — `context.applicationContext` avoids the retained Activity.

---

Also: CodeRabbit's fourth comment (add `state.appListType` to the Freezer `LaunchedEffect` keys) is already satisfied at `b83fb66` — both effects key on `state.searchQuery, state.appListType`.

Not flagged, checked and cleared: `Shizuku.requestPermission(1001)` matches `HomeActivity`'s `requestCode = 1001`; `openManagerApp` correctly null-checks `getLaunchIntentForPackage` and sets `FLAG_ACTIVITY_NEW_TASK`; `PrivilegeManagerApp`'s `PackageManager` lookups are covered by the existing `QUERY_ALL_PACKAGES` declaration; the removed `<permission android:name="android.permission.QUERY_ALL_PACKAGES">` block was a reserved-system-permission redeclaration and its removal is correct; the new string is present in all eight in-app locales.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

trinadhthatakula:
### Review fixes applied

**Model:** `claude-opus-5` (Claude Opus 5). All five findings from [the review above](https://github.com/trinadhthatakula/Thor/pull/417#issuecomment-5356767002) are fixed in `8ffb5ee`.

**Verification:** `:app:compileFossDebugKotlin` and `:app:lintStoreRelease` both clean (no new warnings — the two that remain, the Koin-compiler version note and the deprecated `rememberModalBottomSheetState` in `BackupRestoreHubScreen.kt:763`, are pre-existing). `:app:testFossDebugUnitTest --rerun-tasks`: **1617 tests, 0 failures, 0 errors, 0 skipped** (counts read from `build/test-results/**/*.xml`, not the log line).

---

## 1. The `flags = 0` fallback no longer gets pruning authority

New `RetainReason.VisibilityFallback`, short-circuited at the verdict so none of `scanVerdict`'s rules — which cannot see that `MATCH_UNINSTALLED_PACKAGES` was dropped — ever get to Accept the ~300 packages such a scan does return.

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt#L355-L371

The flag is set only where the fallback actually replaces the list:

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt#L265-L280

**The PR's own fix is untouched by this.** `currentList` is still built from the fallback list and still emitted as `currentList + retained`, and `appDao.syncCache(toUpdate, emptyList())` still runs on the `Retain` path — so a fresh install on a Chinese ROM populates the cache and shows every app exactly as this PR intends. Only three things are withheld: the deletions (and the icon PNGs with them), `pruneWatchlist`, and `recordLabelLocale`.

One follow-on that the review did not name: routing through `is ScanVerdict.Retain ->` would have incremented `consecutiveSuspectScans`, spending the `SUSPECT_SCAN_TOLERANCE` that exists so a *genuinely* shrinking device is eventually believed. On a device permanently in fallback that budget is gone by the third scan, and the next truncated scan gets the `Accept` the guard was built to withhold. So the new reason is excluded from the counter — and does not reset it either, since a fallback scan is no evidence the device is healthy:

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/data/repository/AppRepositoryImpl.kt#L402-L414

Tests: `aDisbelievedScanPrunesNothingWhateverItsReason` already iterates `RetainReason.entries`, so it covers the new constant automatically; added `visibilityFallbackIsNeverAVerdictThisFunctionReaches`, which sweeps every scan shape × permission state × tolerance value and asserts `scanVerdict` never mints this reason itself. That is the boundary the KDoc claims, and it is now enforced rather than described.

## 2. Scroll-to-top no longer discards restored scroll position

Extracted rather than copy-pasted four times, because the subtlety is in the `remember`-not-`rememberSaveable` choice and that comment should exist once:

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/widgets/ScrollToTopOnChange.kt#L13-L40

Applied at all four sites — [`AppList.kt#L590-L595`](https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/widgets/AppList.kt#L590-L595) (grid), [`AppList.kt#L617-L622`](https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/widgets/AppList.kt#L617-L622) (list), [`FreezerScreen.kt#L370-L375`](https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/freezer/FreezerScreen.kt#L370-L375) (grid), [`FreezerScreen.kt#L414-L419`](https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/freezer/FreezerScreen.kt#L414-L419) (list). The sort/filter/search reset behaves exactly as this PR added it; only the first run is skipped.

## 3. One privilege probe per cold start, restored

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/HomeActivity.kt#L332-L346

`isReady` latches on `PrivilegeManager`'s first emission and never goes back, so this is precisely "skip the resume that races the cold-start probe, keep every later one". The resume-reacquisition behaviour the PR added is intact.

## 4. `ThorShellConfig` comment corrected, timeout named

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/core/ThorShellConfig.kt#L20-L47

The rewrite separates what Odin's source proves from what the field symptom suggests. Proven: `BuilderImpl.start()` catches `NoShellException` around `su --mount-master` and falls through to `su`, then `sh`, so the flag cannot block acquisition by itself; and an su that *exits* on the bad argument is cheap, because `ShellImpl.shellCheck()` calls `process.exitValue()` first and throws immediately. Not proven, and now labelled as such: that KernelSU/APatch hang instead of exiting, which is the only path by which the flag costs a full `setTimeout` budget plus a second superuser prompt. Removing the flag is right either way — Thor never needs the global mount namespace.

`setTimeout(10)` is **kept**, now as a named constant with the reasoning attached: it aligns the builder's shell-check budget with Odin's own `RealShellRepository.SHELL_INIT_TIMEOUT_MS = 10_000L`, so a hung `su` doesn't outlive the probe that already gave up on it. The trade the review flagged is recorded in the comment rather than silently reversed — a grant answered between 10s and 20s now has its shell destroyed instead of cached, costing one extra prompt on the next probe. Still worth the device check on a slow first grant; it is a decision now, not an unexplained number.

## 5. `requestDhizuku` off the main thread, and off the Activity context

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/presentation/home/HomeViewModel.kt#L200-L220

The grant callback now posts back through `viewModelScope` instead of running inline on Dhizuku's listener thread, and reuses `refreshPrivileges()` — the same three steps it had inlined, so the two grant paths cannot drift.

The `applicationContext` normalisation went into `DhizukuHelper` rather than the call site, since that object is what hands the context to `DhizukuAPI`, which retains it statically for the process lifetime. Both `init` sites are covered, not just the one the review found:

https://github.com/trinadhthatakula/Thor/blob/8ffb5ee4309382f91335176ad8384a43b486d214/app/src/main/java/com/valhalla/thor/data/source/local/dhizuku/Dhizuku.kt#L147-L163

---

Nothing in the PR's intended behaviour is reverted: the fresh-install package-visibility fix, the per-user AppOp grants, the in-app Shizuku request, the `AppScanRevision` seam and the `AppFilterSheet` rewrite are all as submitted.

PR #419, opened 2026-08-20
fix(visibility): stop gating GET_INSTALLED_APPS app-ops on the pm grant exit code

trinadhthatakula:
## The regression

On Chinese ROMs (MIUI / HyperOS and friends) the app list came back showing **Thor as the only installed app** — the exact symptom #329 and #417 were written to fix. It reproduced on a fresh install after those PRs merged.

## Root cause

`b83fb664` (a review-feedback commit on #417) tightened the self-grant path in all three gateways so that `installedAppsAppOpGrantCommands` only ran **after `pm grant` had already succeeded**:

```kotlin
// before this PR
if (res.isSuccess && permissionName == GET_INSTALLED_APPS_PERMISSION) {
    installedAppsAppOpGrantCommands(...).forEach { runCommand(it) }
}
```

That reads as a sensible "don't do follow-up work if the grant failed" tidy-up, and it is backwards for this one permission. `com.android.permission.GET_INSTALLED_APPS` is vendor-defined (T/TAF 108-2022), not AOSP:

- AOSP `pm grant` frequently **exits non-zero** for it on these ROMs
- the **app-op** is what actually opens package visibility, in any of three spellings — `GET_INSTALLED_APPS`, `android:get_installed_apps`, numeric `10022`

So the app-ops are a **parallel route** to visibility, not a follow-up to the grant. With the gate in place they were never issued, visibility stayed closed, and the list stayed at one entry.

## Fix

The app-ops now run whatever `pm grant` returned, and **either** route succeeding is reported as success. Same shape in `RootSystemGateway`, `ShizukuSystemGateway` and `DhizukuSystemGateway`.

## Two further holes on the same path

Each of these could reproduce the symptom on its own, so both are fixed here rather than left as latent duplicates of the same bug:

**1. `SelfPermissionGranter` only bumped the rescan if `checkSelfPermission` came back granted.**
A gateway can open visibility while the re-read still says denied — setting the app-op does not move `checkSelfPermission` at all. The bump now also fires when a grant command reported success, so the DB resync happens on the route that actually works.

**2. `AppScanRevision` was a `MutableSharedFlow(replay = 0)`, which silently discards a `tryEmit` made with no collector.**
The self-grant bumps from the privilege probe started in `ThorApplication.onCreate`, which on a fast device finishes *before* the first ViewModel subscribes — so the one signal that mattered landed in an empty room and was dropped, with nothing scheduled to fix it. It is now a monotonic `StateFlow<Int>` whose value survives having no subscriber. `AppRepositoryImpl` collects it with `.drop(1)`, so the initial read does not buy a second scan on top of the one the collector already performs on subscribe.

## Validation

- `lintStoreRelease` — green
- `testFossDebugUnitTest --rerun-tasks` — **1621 tests, 0 failures**, including 4 new `AppScanRevisionTest` cases that lock in the dropped-bump behaviour
- **device-tested on a Chinese ROM** — tested and verified on a Chinese HyperOS rom by the maintainer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved granting of the installed-apps permission by trying all supported grant methods, even when one method fails.
  * Permission requests now succeed when any supported grant route works.
  * App scanning no longer performs an unnecessary duplicate scan on startup.
  * Scan refresh requests are preserved and reliably delivered, including requests raised before monitoring begins.

* **Tests**
  * Added coverage for scan refresh persistence, delivery, timing, and revision ordering.

### DIFF
--- a/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
+++ b/app/src/main/java/com/valhalla/thor/data/gateway/DhizukuSystemGateway.kt
@@ -466,15 +466,31 @@ class DhizukuSystemGateway(
         val escapedPermissionName = permissionName.escapeForShell()
         return try {
             val result = DhizukuHelper.execute("pm grant --user $userId $escapedPackageName $escapedPermissionName")
-            if (result.first != 0) {
-                return Result.failure(Exception("Dhizuku: pm grant failed with exit code ${result.first}: ${result.second}"))
+            val grantFailure = {
+                Result.failure<Unit>(
+                    Exception("Dhizuku: pm grant failed with exit code ${result.first}: ${result.second}")
+                )
             }
-            if (permissionName == GET_INSTALLED_APPS_PERMISSION) {
-                installedAppsAppOpGrantCommands(escapedPackageName, userId).forEach { cmd ->
-                    DhizukuHelper.execute(cmd)
-                }
+            if (permissionName != GET_INSTALLED_APPS_PERMISSION) {
+                return if (result.first == 0) Result.success(Unit) else grantFailure()
             }
-            Result.success(Unit)
+
+            // The app-ops are a *parallel route* to package visibility, not a follow-up to the
+            // grant, so they run whatever `pm grant` returned. On the ROMs this permission exists
+            // for — MIUI/HyperOS, ColorOS, OriginOS — the AOSP `pm grant` of a vendor-defined
+            // permission frequently exits non-zero while the app-op is the thing that actually
+            // opens the package list, which is why installedAppsAppOpGrantCommands fires three
+            // spellings of it. Gating them on the grant succeeding is what made a Chinese-ROM
+            // install come back with Thor as the only visible app: the grant failed, the app-op
+            // was never set, and nothing else in the app knows how to open that gate.
+            val appOpTook = installedAppsAppOpGrantCommands(escapedPackageName, userId)
+                .map { DhizukuHelper.execute(it) }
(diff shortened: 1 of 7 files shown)


trinadhthatakula:
Addressed in 070fc5c7 — the finding is valid. `worker` is launched onto a multi-threaded dispatcher at the top of `callbackFlow`, so it can be inside `getInstalledPackages()` before `scanWatcher` subscribes ~290 lines later; a bump in that window moved the StateFlow value and the late `drop(1)` discarded it as the replay, leaving only the scan that read the package list before the grant landed.

Fixed with an explicit baseline rather than `CoroutineStart.UNDISPATCHED`. Both close the window, but the baseline doesn't depend on dispatch semantics — nothing later inserting a suspension point ahead of the `collect` can silently reopen it — and, unlike launch ordering, it's deterministically testable, which is the other half of what you asked for.

`AppScanRevision` gains `snapshot()` / `requestsAfter(baseline)`; the repository reads the snapshot synchronously *before* `launch(ioDispatcher)`, so the baseline strictly precedes the scan. A bump after it is delivered however late the watcher subscribes; a bump predating the collection still reads as the baseline and is skipped, since the worker's own initial trigger covers it.

Test coverage is the exact ordering named: baseline taken, bump raised with nobody collecting, watcher subscribes only afterwards. It also asserts that `revision.drop(1)` emits nothing under that same ordering, so the test can't silently stop discriminating if someone reverts the mechanism.

`lintStoreRelease` green, 1623 foss unit tests green.

PR #424, opened 2026-08-21
chore(release): v1.95.1 (versionCode 1951)

trinadhthatakula:
The release commit for **v1.95.1**: `versionCode` 1950 → 1951 plus the three note files and the two fastlane changelogs, in one commit as `release-notes/README.md` requires.

## Why this one is cumulative

No stable shipped in the v1.94.x cycle — the three published pre-releases were `v1.94.1-dev-12`, `v1.94.4-dev-26` and `v1.95.0-dev-40`. So v1.95.1 is baselined on `v1.94.4-dev-26`, folds in everything v1.95.0's notes covered, and adds the work merged after the 1950 bump: #417, #418, #419, #420, #421.

## Corrections applied against the source, not the draft

Four claims inherited from the v1.95.0 notes or the first draft did not survive checking:

- **The Shizuku backup claim was false as written.** #413's own summary: `MeasureAppDataUseCase` and `AppBackupViewModel` measure and display "only supported classes (`EXTERNAL_DATA`, `EXTERNAL_MEDIA`, and APK bundles) when running on non-root shell privileges (e.g. Shizuku), completely hiding internal private data (`CE` and `DE`)". The notes now state the scope per privilege mode rather than promising app data Thor cannot read.
- **The privilege bullet named the wrong mechanism.** There is no resume-time pickup: `HomeViewModel.refreshPrivileges()` is what the Privilege Check dialog's **Refresh** calls, and its KDoc is explicit that root and Dhizuku publish no callback, so a grant made while Thor runs "stays invisible until the process is killed and relaunched". The notes say Refresh.
- **"Finds every backup in your storage" overstated the scanner.** `BackupArchiveScannerImpl` covers `Downloads/Thor`, the `MediaStore` Downloads/Files tables, and one configured SAF export directory. That is what the notes now say.
- **Dropped "45+ pull requests"** — carried over from v1.95.0 and not verifiable from the history.

The freezer watchlist auto-cleanup line, dropped from the first draft, is restored, and #383 (Settings reorg) and #371 (CSV export) are folded in — neither is an ancestor of `v1.94.0`, so neither has ever reached a stable release.

## Verification

| Check | Result |
|---|---|
| `.github/scripts/check-notes-budget.sh 1.95.1` | telegram **1016/1024** units (8 spare), playstore **490/500** chars |
| `:app:testFossDebugUnitTest --rerun-tasks` | 122 suites, **1634 tests, 0 failures** (parsed from `build/test-results`, not the log line) |
| fastlane copies | `en-US/changelogs/1951.txt` and `hi-IN/changelogs/1951.txt` byte-identical to `playstore.txt` (`diff` clean); these are the only two locales that exist |

`sync-shizu-changelog.sh` is deliberately **not** run here. It reads `versionCode` from `origin/production`, so during prep it would sync the last *production* changelog, print `changelog already current`, and exit 0 while the manifest never receives 1951 — see `release-notes/README.md` Step 5. It runs on `master`, after the production promotion.

## Known, not addressed here

Five pre-existing conflicts await the next `dev` → `master` promotion, all under `web/`: `follow-ups-report.html`, `robots.txt`, `BaseLayout.astro`, `lockstep.test.ts`, `site.test.ts`. Out of scope for the release commit.

## Summary by CodeRabbit

* **New Features**
  * Added a Backup & Restore Hub with encrypted offline app-data backups.
  * Added customizable App Info actions, icon gestures, and XAPK/OBB support.
  * Expanded language support and clarified freezer labels.

* **Bug Fixes**
  * Improved app-list loading and refresh behavior, including root and Shizuku access.

* **Documentation**
  * Added comprehensive v1.95.1 release notes and localized Android store changelogs.

### DIFF
--- a/fastlane/metadata/android/en-US/changelogs/1951.txt
+++ b/fastlane/metadata/android/en-US/changelogs/1951.txt
@@ -1,0 +1,9 @@
+• 📦 Backup & Restore Hub: browse, restore, share or delete backups and app bundles from one screen
+
+• 💾 Encrypted offline backups: full app data on root, APK + shared storage on Shizuku, plus .xapk & OBB
+
+• 🔀 Reorder or hide App Info actions; tap an app's icon to open it, long-press for system settings
+
+• 🔧 Fixed app lists that showed only Thor; newly granted root or Dhizuku applies on Refresh
+
+• 🌐 8 languages: EN, AR, ES, FR, PL, PT, PT-BR, ZH, plus clearer freezer action labels
--- a/fastlane/metadata/android/hi-IN/changelogs/1951.txt
+++ b/fastlane/metadata/android/hi-IN/changelogs/1951.txt
@@ -1,0 +1,9 @@
+• 📦 Backup & Restore Hub: browse, restore, share or delete backups and app bundles from one screen
+
+• 💾 Encrypted offline backups: full app data on root, APK + shared storage on Shizuku, plus .xapk & OBB
+
+• 🔀 Reorder or hide App Info actions; tap an app's icon to open it, long-press for system settings
+
+• 🔧 Fixed app lists that showed only Thor; newly granted root or Dhizuku applies on Refresh
+
+• 🌐 8 languages: EN, AR, ES, FR, PL, PT, PT-BR, ZH, plus clearer freezer action labels
--- a/gradle.properties
+++ b/gradle.properties
@@ -51,7 +51,7 @@ org.gradle.welcome=never
 initialVersionCode=1921
 # REQUIRED: The active source of truth for our release manager
 # Logic: 1822 -> 1.82.2 (based on math: code/1000 . code%1000/10 . code%10)
-versionCode=1950
+versionCode=1951
 # versionName intentionally omitted — auto-calculated from versionCode by the
 # resolveVersionName()/calculateVersionName() helpers in app/build.gradle.kts
 # (1908 -> 1.90.8). Per CLAUDE.md, never set versionName directly; bump versionCode only.
--- a/release-notes/v1.95.1/github.md
+++ b/release-notes/v1.95.1/github.md
@@ -1,0 +1,124 @@
+# Thor v1.95.1 Release Notes
+
+The cumulative stable release, uniting everything shipped across the **v1.94.x** development cycle
+and the **v1.95.0** dev build that followed it. If you last updated from a stable release, this is
+the whole of that work in one step.
+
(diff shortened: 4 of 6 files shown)


trinadhthatakula:
Review findings addressed. All three verified against the source before changing anything.

### 1. Privilege refresh statement — **valid, and my error**

`PrivilegeManager.kt:65-71` registers Shizuku's `OnBinderReceivedListener` and `OnRequestPermissionResultListener`, **both of which call `refresh()`**. So Shizuku recovers on its own; root and Dhizuku are the ones that need the Privilege Check dialog's Refresh. Pairing Shizuku with "tap Refresh" was backwards — I had corrected the *mechanism* (it is Refresh, not a resume hook) but then named the one mode the mechanism is not needed for.

Fixed in all four files. The two short channels have no room for the full nuance, so they now name only the modes that require user action:

> `newly granted root or Dhizuku applies on Refresh`

`github.md` carries the whole story in the Highlights bullet and already did so in the #417/#419 section:

> Shizuku granted while Thor is running is picked up on its own; root and Dhizuku are picked up with **Refresh**. Neither needs the process killed.

### 2. "Background Foreground Export" — **valid**

Confirmed the mechanism rather than just reflowing the sentence: `ThorJobNotifications` calls `setForeground(id, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC)`, and the manifest declares `FOREGROUND_SERVICE_DATA_SYNC` with `android:foregroundServiceType="dataSync"` (`AndroidManifest.xml:53,416`). So it is a WorkManager job promoted to a foreground service — which is what the heading now says, with the type named:

> **Background Export via a Foreground Service**: Long exports run in the background as WorkManager jobs promoted to a `dataSync` foreground service, with notification progress channels.

This wording was inherited from v1.95.0's notes, so the confusion predates this PR.

### 3. Stranded staging file sentence — **valid, taken as written**

Readability only; no claim changed.

### Re-validated after the edits

| Check | Result |
|---|---|
| `check-notes-budget.sh 1.95.1` | telegram **1016/1024** units, playstore **485/500** chars |
| fastlane copies | re-synced; `diff` clean against `playstore.txt` for both `en-US` and `hi-IN` |

Telegram briefly sat at 1021/1024 after the correction. Three spare units is too tight against a hardcoded 164-unit wrapper assumption, so I dropped a redundant "live" from the customization bullet to restore the 8-unit margin the branch shipped with.

Amended rather than stacked, so the release commit stays atomic — the next cycle's notes are baselined on the previous `chore/release-NNNN` commit, and a split bump/notes pair would break that.

Issue #57, opened 2025-07-26
Sort by Size Option for Installed Apps


Raviksh96:
Hi @trinadhthatakula 
I don't know if this feature is available, but after installing the app in IQOO Neo 10, I'm not getting the option to sort my apps by size with and without Shizuku.

trinadhthatakula:
Hi @Raviksh96 — sort-by-size is implemented now 🎉 There's a new **Size** option in the sort menu, and each app's total install size shows in its App Info sheet. It needs **Usage Access** (Thor prompts for it, or auto-grants if you have Root/Shizuku/Dhizuku).

Available in pre-release **v1.90.3-dev-92** → grab `foss-release.apk`: https://github.com/trinadhthatakula/Thor/releases/tag/v1.90.3-dev-92

Could you try it on your IQOO Neo 10 and let me know if it works? I'll close this once it ships in the stable release. Thanks for the request!

Raviksh96:
Sorry for the late reply again. I tried the beta version 1.90.3-foss, the sort by size option is there but it's not working as expected. Is there anything I can do to help further, like any kind of logs or something else?

Edit_1 : Tried today, and somehow it's sorting the apps by size because I closed the app on that setting. But the sorting order breaks after I try to sort again by name or any other filter, ascending or descending order. I need to close the app and open again after clearing from recent also to get the sorting or as I want, Sort by size - descending.

trinadhthatakula:
Hi @Raviksh96, please upgrade to the latest stable **v1.92.0** release which has just been published! 

Additionally, to retrieve app size information and enable sorting by size, please make sure that the **Usage Access** permission is enabled. You can check and grant this directly inside **Thor's app settings** (open Thor, go to Settings, and make sure **Usage Access** is enabled). Let us know if that resolves the issue!

trinadhthatakula:
Following up — and being straight with you: **v1.93.1 contains no change to size sorting.** I don't want you installing a new build expecting a fix that isn't in it.

Where this actually stands:

- Sort by **Size** (filter sheet → Sort by → Size) has been in the app since v1.92.0, along with the Size row in App Info.
- The only sorting fix in v1.93.1 is for **Version Code** — codes above roughly 2.1 billion used to sort into the wrong place. Different sort key, unrelated to what you reported.

[...]

PR #435, opened 2026-08-25
feat(components): open, force-open, stop and disable individual components

trinadhthatakula:
Closes the long-standing "let me control components" request: the Components tab in App Info stops being a read-only list and becomes a set of controls, gated by what the active privilege transport can actually do.

Design spec: `docs/superpowers/specs/2026-08-25-component-control-design.md`.

## What you can do now

| Component | Action | Needs |
|---|---|---|
| Activity, exported & unguarded | **Open** | nothing |
| Activity, not exported or permission-guarded | **Force open** | uid 0 |
| Activity, exported | **Force open** (overflow) — the shell route still works when the app is frozen or suspended | uid 0 |
| Service | **Stop now** (transient) | uid 0 |
| Service / Receiver / Activity | **Disable** and **Enable** (persistent) | uid 0 |
| Anywhere | **Restore all** — every component Thor disabled, in every app, back to how it shipped | uid 0 |

Every row also offers **Copy class name**, which needs nothing.

## Which privilege gets what

`ComponentCapability` stores exactly one fact — **does the transport execute at uid 0** — because every privileged verb here collapses onto it:

- **Setting component state** is uid 0 only. `PackageManagerService.setEnabledSetting` carves out `Process.SHELL_UID` for *package*-level changes and explicitly rejects a non-null class name. Reaching `IPackageManager.setComponentEnabledSetting` by reflection hits the same check with the same uid, so there is no fallback.
- **Force-launching** is uid 0 only. `ActivityManager.canAccessUnexportedComponents` waives export and permission checks for `ROOT_UID` and `SYSTEM_UID` alone. `START_ANY_ACTIVITY` is `signature` and the Shell package does not declare it in any release from 9 to 16.
- **Stopping a service** is uid 0 only *here*. `ActiveServices.retrieveServiceLocked` would let any uid stop an **exported** service, so a shell-uid Shizuku could stop some of them — that narrower path is deliberately not offered, because a "Stop" that silently does nothing on half the rows is worse than one that is honestly absent.

So: **root** and **Shizuku started as root** get everything; **shell-uid Shizuku** and **Dhizuku** get the read-only list plus Open on exported activities, under a banner that says why. Dhizuku is not a partial case but an empty one — `DevicePolicyManager` has no component-enabled API at all.

Per the design decision on the issue, Thor **never** touches `Settings.Secure.assistant` and never asks for `WRITE_SECURE_SETTINGS`; force launch stays root-only rather than borrowing the assistant-method trick.

## The ledger is bookkeeping, not policy

`component_overrides` (Room v7, `AutoMigration(6, 7)`, one new table, no existing column touched) records what Thor disabled so it can offer Restore All and flag drift. **PackageManager stays the source of truth.** There is no boot receiver and no re-apply sweep: if another tool re-enables a component, Thor reports it as *"Changed elsewhere"* rather than fighting for it. A restore that fails on one row leaves that row in the ledger and reports `Restored N of M`.

## Verification

Driven end to end on `emulator-5556` (API 37 / SDK 37 userdebug AVD, Shizuku's own server started at uid 0 so `hasUid0 = true` exercises the privileged path through `ShizukuSystemGateway`):

- Open on an exported activity, with and without privilege
- Force open disabled on a non-exported activity when not uid 0
- Stop now on a **running** service and on a **not-running** one — both report honestly
- Disable on a broadcast receiver → `dumpsys package` shows it under `disabledComponents`
- Drift: external `pm default-state` → the row flips from "Restricted by Thor" to "Changed elsewhere" without a process restart
- Restore all, **partial** (`Restored 3 of 4 — the rest are still restricted`, failing row survives in the ledger) and **complete** (`Restored 2 components`, ledger empty, no PM overrides left)
- Both blocker banners: shell-uid Shizuku → *"Changing components needs Shizuku started as root."*; no transport → *"Turn on Root or Shizuku to change components."*

`:app:testStoreDebugUnitTest --rerun-tasks` → **1721 tests, 0 failures, 0 errors**. `:app:lintStoreRelease` → 8 issues, all Hint, identical to the committed baseline.

## Three platform traps worth knowing

1. **`am stopservice` exits 255 in every case, success included.** The only discriminator is stderr: `Service stopped` / `Service not stopped: was not running.` / `Error stopping service`. Hence `ComponentCommandKind.STOP_SERVICE`, which reads markers alone and ignores the exit code.
2. **On SDK 37 `am` writes the outcome to stderr and only an echo to stdout.** `ShizukuHelper.execute` returns `stdout.ifBlank { stderr }`, so the outcome was invisible — `executeCombined` was added alongside it. `execute`'s contract is unchanged for its several dozen existing callers.
3. **`ComponentInfo.enabled` is the manifest `android:enabled` attribute**, never the effective state, so `componentSnapshotOf` calls `getComponentEnabledSetting` per component.

Strings are translated into all eight in-app locales.

**Do not merge** — @trinadhthatakula will merge after a physical rooted-device test.

## Summary by CodeRabbit

* **New Features**
  * Added per-component controls in App Info for enabling, disabling, resetting, launching, stopping, and forgetting components.
  * Added detailed status information for activities, services, receivers, and providers, including state and restriction badges.
  * Added restore-all support for components disabled by Thor.
  * Added capability guidance based on available privilege access, with clear unsupported-action messages.
  * Added a confirmation disclaimer before disabling components.
* **Localization**
  * Added translations for component controls and messages in Arabic, Chinese, English, French, Polish, Portuguese, and Spanish.

### DIFF
--- a/app/schemas/com.valhalla.thor.data.source.local.room.AppDatabase/7.json
+++ b/app/schemas/com.valhalla.thor.data.source.local.room.AppDatabase/7.json
@@ -1,0 +1,352 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 7,
+    "identityHash": "146ad7d9667c480df2c87f09bf066517",
+    "entities": [
+      {
+        "tableName": "apps",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`packageName` TEXT NOT NULL, `appName` TEXT, `versionName` TEXT, `versionCode` INTEGER NOT NULL, `minSdk` INTEGER NOT NULL, `targetSdk` INTEGER NOT NULL, `isSystem` INTEGER NOT NULL, `installerPackageName` TEXT, `publicSourceDir` TEXT, `splitPublicSourceDirs` TEXT NOT NULL, `enabled` INTEGER NOT NULL, `dataDir` TEXT, `nativeLibraryDir` TEXT, `deviceProtectedDataDir` TEXT, `sharedLibraryFiles` TEXT, `obbFilePath` TEXT, `sourceDir` TEXT, `sharedDataDir` TEXT NOT NULL, `lastUpdateTime` INTEGER NOT NULL, `firstInstallTime` INTEGER NOT NULL, `isDebuggable` INTEGER NOT NULL, `isSuspended` INTEGER NOT NULL, `installSize` INTEGER, PRIMARY KEY(`packageName`))",
+        "fields": [
+          {
+            "fieldPath": "packageName",
+            "columnName": "packageName",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "appName",
+            "columnName": "appName",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "versionName",
+            "columnName": "versionName",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "versionCode",
+            "columnName": "versionCode",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "minSdk",
+            "columnName": "minSdk",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "targetSdk",
+            "columnName": "targetSdk",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isSystem",
+            "columnName": "isSystem",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "installerPackageName",
+            "columnName": "installerPackageName",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "publicSourceDir",
+            "columnName": "publicSourceDir",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "splitPublicSourceDirs",
+            "columnName": "splitPublicSourceDirs",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "enabled",
+            "columnName": "enabled",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "dataDir",
+            "columnName": "dataDir",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "nativeLibraryDir",
+            "columnName": "nativeLibraryDir",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "deviceProtectedDataDir",
+            "columnName": "deviceProtectedDataDir",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "sharedLibraryFiles",
+            "columnName": "sharedLibraryFiles",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "obbFilePath",
+            "columnName": "obbFilePath",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "sourceDir",
+            "columnName": "sourceDir",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "sharedDataDir",
+            "columnName": "sharedDataDir",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "lastUpdateTime",
+            "columnName": "lastUpdateTime",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "firstInstallTime",
+            "columnName": "firstInstallTime",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isDebuggable",
+            "columnName": "isDebuggable",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isSuspended",
+            "columnName": "isSuspended",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "installSize",
+            "columnName": "installSize",
+            "affinity": "INTEGER"
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "packageName"
+          ]
+        }
+      },
+      {
+        "tableName": "freezer_apps",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`packageName` TEXT NOT NULL, PRIMARY KEY(`packageName`))",
+        "fields": [
+          {
+            "fieldPath": "packageName",
+            "columnName": "packageName",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "packageName"
+          ]
+        }
+      },
+      {
+        "tableName": "extension_data",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`extensionPackageName` TEXT NOT NULL, `key` TEXT NOT NULL, `value` TEXT NOT NULL, PRIMARY KEY(`extensionPackageName`, `key`))",
+        "fields": [
+          {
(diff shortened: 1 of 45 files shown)


trinadhthatakula:
## Code review

Four reviewers over the diff — data/domain correctness, presentation + i18n, test coverage, and CodeRabbit — plus verification of each finding against the API 37 emulator before acting on it. **Six defects found and fixed; three findings refuted.** Everything below is already pushed.

### Fixed

**1. The numeric error code beat the sentence that explained it.** `ActivityManagerShellCommand` prints `Error type 3` on the line *above* `Error: Activity class {…} does not exist.` Both lines were marked and the scan takes the first matching **line**, not the first matching **marker** — so every refused force-launch reported `Error type 3`, naming neither the component nor the reason. That is precisely the content-free-header failure mode `FAILURE_HEADER_MARKER` was written to prevent, reintroduced through the marker list. `Error type` now sits in a second `WEAK_FAILURE_MARKERS` tier searched only after the descriptive markers have failed on every line, so it still speaks when it is the only evidence there is.

The existing test passed because it fed a **one-line** output that no device emits. Verbatim device capture:

```
Starting: Intent { cmp=moe.shizuku.privileged.api/….ApplicationManagementActivity }
Error type 3
Error: Activity class {…} does not exist.
```

**2. A dead transport was reported as a stopped service.** `componentCommandFailure` ignored the exit code entirely for `STOP_SERVICE` — correct for `am`, which exits 255 whatever happens, but `ShizukuHelper` also returns **-1** for a dead binder, a timeout, or a thrown exception. `"Shizuku binder is null"` carries no marker, so the function returned `null` and the user got "Stopped X" for a command that never ran. `am`/`pm` exit 0..255, so a negative code is unambiguously Thor's own sentinel; negative codes are now read for both kinds. 255-with-a-success-marker still returns before that branch, so device behaviour is unchanged — re-verified on the emulator after the change (screenshot: "Stopped AlarmService").

**3. A Room failure after a successful `pm disable` crashed the process.** Every verb returns a `Result`, but the ledger write that follows a successful one runs inside `Result.onSuccess`, which does not catch. `runExclusively` had `try`/`finally` and **no** `catch`, so a `SQLiteFullException` there went straight to `viewModelScope`'s uncaught handler — after the privileged change had already succeeded. Now caught and reported like any other failure, with `CancellationException` rethrown first.

**4. The Components tab announced a copy that had not happened.** `setClipEntry` suspends; the Toast sat beside the `launch` rather than inside it, so it fired before the clipboard write and would still fire if the scope was cancelled. The same file documents this trap at two other copy sites — this one was the odd one out.

**5. Count-dependent grammatical agreement in four locales.** `component_restore_all_partial` is driven by *two* counts so it cannot be a `<plurals>`, and the first one can be 1: fr "1 composants rétablis", es "Se restauraron 1 de 2" (whose "el resto **siguen**" also disagreed with its own singular subject), pt "Repostos 1 de 2", pt-rBR "Restaurados 1 de 2". All four now lead with a noun, which agrees with nothing. CodeRabbit flagged only French — the others are the same class, found by grepping the construction rather than the finding. English hid it because its past participles do not inflect for number. `ar` (verb-first) and `pl` (impersonal `-no`) were already count-agnostic; `zh-rCN` has no agreement at all.

**6. Dead API.** `ComponentControlUseCase.allOverrides()` had zero call sites. Removed; `restoreAll` uses the repository's `getAll` directly.

### Refuted

**"Force open should not require `component.enabled`."** Tested on device — `am start` on a disabled component fails **even at uid 0**:

```
$ pm disable pkg/….ApplicationManagementActivity
$ am start -n pkg/….ApplicationManagementActivity      # as root
Error type 3
Error: Activity class {…} does not exist.              # EXIT=1
```

PackageManager resolution excludes disabled components regardless of caller uid, so the `&& component.enabled` guard is correct — a disabled component must be enabled first, and the row already offers that.

**"zh-rCN wrongly says Restore All returns components to factory state."** It does return them to the shipped state: `disable` records `restoreToEnabled = component.manifestDefaultEnabled` — the manifest default as it stands, deliberately not `true`, so a component that ships disabled is restored to disabled. 恢复为出厂时的状态 is a faithful translation of both the English string and the behaviour.

**"`launchRequiresRoot` ignores a permission Thor itself holds."** True of AOSP in principle — `ActivityStarter` checks the caller's permission — but this is a conservative false negative, not a wrong outcome: it withholds an action in the rare case of an exported activity guarded by a permission Thor happens to hold. Making it exact means a per-row `checkSelfPermission`, and the design decision on this feature was explicitly root-only force launch. Left as is.

### Not fixed, noted

- **`ComponentControlViewModel` has no unit test**, unlike its two package siblings. `ComponentCapabilityProvider` being a final class is the one real obstacle; `isDrifted`, the `load()` rebind guard, the stale-package guards and the consent gate are all pure and testable today. Worth a follow-up.
- **`info_app_size` is missing from `values-pt-rBR`** — pre-existing on `dev` (introduced by #57), untouched by this branch, so out of scope here.
- **An unreadable Shizuku uid reports `SHIZUKU_NOT_ROOT`**, which asserts shell-mode when the truth is "could not tell". The capability is right (fail closed, deliberate and documented); only the banner wording over-claims, and a fourth blocker string across 8 locales is disproportionate to a path that needs a live Shizuku with a throwing binder.

### State

1727 unit tests, 0 failures. `lintStoreRelease` 8 issues, all Hint, identical to the committed baseline. Every device check re-run after the fixes.

trinadhthatakula:
## Verified on a rooted physical device

@trinadhthatakula has confirmed per-component control works on real rooted hardware. That was the
one gate this PR was held on — the emulator run in the description used Shizuku's own server
started at uid 0 as a stand-in for root, because the AVD's `/system/xbin/su` is
`-rwsr-x--- root shell` and so is unreachable from an app at uid 10xxx without Magisk. The physical
device closes that gap: the `RootSystemGateway` path is now exercised by the real Odin root shell,
not by a uid-0 substitute.

The **Do not merge** line at the end of the description is therefore spent, and this is clear to
merge.

### Two corrections to the description

- **Test count.** It says `1721 tests`, which was true when it was written. The three review fixes
  pushed after it (`cf19a134`, `06781eac`, `e7090e42`) added six tests:
  `:app:testStoreDebugUnitTest --rerun-tasks` → **1727 tests, 0 failures, 0 errors**. Counts read
  from `build/test-results/testStoreDebugUnitTest/**/*.xml`, not from the Gradle log line.
- **Trap 1 is understated.** It says `am stopservice` exits 255 in every case and that the
  discriminator is stderr. Both true, but the first cut of that logic then ignored the exit code
  *entirely* for `STOP_SERVICE` — which made a negative code, Thor's own transport sentinel for a
  dead Shizuku binder or a timeout, read as a stopped service. `e7090e42` restores the distinction:
  a code the *command* chose is ignored, a code the *transport* chose is a failure. `am` and `pm`
  exit `0..255`, so a negative code means the command never ran.

### Post-review state

Three fixes landed after the review, all with tests that fail without them:

| Commit | Defect |
|---|---|
| `cf19a134` | `stopservice`'s verdict read from its exit code rather than from stderr |
| `06781eac` | `Error type 3` outranked the descriptive `Error: Activity class … does not exist.` line, because the scan was first-line-wins. The old test passed only because it fed a one-line output no device emits |
| `e7090e42` | A dead transport (`-1`, `"Shizuku binder is null"`) reported as a stopped service |

Also fixed in review: a missing `catch` in `ComponentControlViewModel.runExclusively` — the ledger
write sits inside `Result.onSuccess`, which does not catch, so a Room failure there took the process
down *after* the privileged change had already succeeded; a copy Toast that fired before the
suspending clipboard write; count-dependent grammatical agreement in `component_restore_all_partial`
across **four** locales (`fr`, `es`, `pt`, `pt-rBR` — the review flagged only French, the rest came
from grepping the construction); and a dead `allOverrides()`.

Three review findings were refuted rather than applied, each on evidence:

- **Force open does not need `component.enabled` dropped from its gate.** `am start` on a disabled
  component fails even at uid 0 (`Error type 3` / `does not exist.`) — PM resolution excludes
  disabled components regardless of caller uid, so gating on `enabled` is correct, not
  over-cautious.
- **The zh-rCN "factory state" wording is faithful.** `disable` records
  `restoreToEnabled = component.manifestDefaultEnabled`, so a restore returns the component to the
  state it shipped in.
- **`launchRequiresRoot` ignoring a permission Thor happens to hold** is a conservative false
  negative, and root-only force launch was an explicit design decision on this PR.

Known follow-up, deliberately not in scope here: `ComponentControlViewModel` has no unit test. The
command-verdict logic and the use case both do.

PR #436, opened 2026-08-25
fix(components): answer the three review findings on the disable path

trinadhthatakula:
Follow-up to #435. CodeRabbit posted three findings on `ComponentControlViewModel` that were not read before that PR merged — this is them, checked against the merged code and answered.

All three were still valid. One of them turned out to be worse than reported.

## 1. A tap could kill the process

`forget` and Restore All each launched bare on `viewModelScope`, which installs **no** exception handler. Both open with a raw Room call, so a `SQLiteFullException` from either reached the thread's default handler and took Thor down — losing the whole screen while leaving whatever had already happened in place.

What makes this worth saying plainly: **this defect was already found and fixed once in this same class**, in `runExclusively`, and was never swept across its siblings. Two launches were left. Every launch in the class now routes through one `launchReporting`, which rethrows `CancellationException` and reports everything else, so there is no longer a second path to sweep.

## 2. A full database made a successful disable report "failed"

`overrides.record` ran inside `Result.onSuccess`, which does not catch. The component was genuinely disabled and the user was told the disable failed — the worst combination, because the obvious response is to try again.

`ComponentActionOutcome` now carries the platform result and the ledger error separately:

- **the platform decides whether the action worked.** A ledger failure is not a failed disable.
- the ledger error is logged in every case, so it survives into a bug report.
- it is shown to the user in exactly one place — a disable whose row was not written, via `component_disabled_unrecorded`, because that is the only case where the failure changes what the user can do next: with no row, **Restore all will never offer that component back**. For the restoring verbs a leftover row is drift, and the list already labels that "Changed elsewhere" and offers Forget.

`restoreAll` counts a row only when both halves land, so it under-reports rather than claiming "3 of 3 restored" while the screen still says 1 component is restricted. The leftover row makes the next press a retry, and `setComponentEnabled` is idempotent.

**CodeRabbit's suggested remedy — compensate or roll back the disable — was not taken.** The ledger is bookkeeping, not policy: PackageManager is the source of truth and Thor never re-applies a row. Re-enabling a component because Thor's own notebook was full would undo what the user asked for in order to fix something they did not, and it adds a failure mode of its own (the rollback can fail too, and then what?). Honest reporting is the smaller promise and the true one.

## 3. The consent gate was fail-open — and the default bought nothing

`consentAccepted` defaulted to `true` while DataStore was still loading, so a disable requested in that window skipped the disclaimer entirely. On inspection it was worse than the finding said: the flag was read in exactly one place and **nothing in the UI rendered it**, so the "avoid a flash of the dialog" reasoning the default rested on described a hazard that could not occur. It was a fail-open default in exchange for nothing.

Rather than pick a safer default for the unknown window, the window is gone: consent is a synchronous in-memory flag with no load to await.

### Consent now lasts one session, and only if asked for

Per the owner's call, and it changes the gate's shape:

- the disclaimer is raised for **every** disable by default,
- the dialog carries a **"Don't ask again this session"** checkbox, and confirming *without* ticking it is consent for that component only,
- ticking it silences the disclaimer until Thor's process dies. Nothing is persisted — no DataStore key, no timestamp, no clock or timezone to be wrong about.

A permanent don't-ask would disarm, on one possibly-mis-tapped checkbox, the only warning standing between a mis-tap and an app that is broken while every visible sign of health stays intact. A session is the longest scope that still expires on its own.

It lives in a `@Single` (`ComponentConsentSession`) rather than in the view model, because `ComponentControlViewModel` is keyed per package — view-model-local state would re-ask for every app opened, which is the annoyance the checkbox exists to remove.

**The extension disclaimer is untouched.** That one is a grant-and-forget liability acknowledgement for a whole feature, not a per-action warning, and it stays persisted.

The `componentControlConsentAccepted` preference is removed end to end: field, interface method, DataStore key, setter and flow mapping. Deleting the `init` collect also removed a fourth latent crash site — `userPreferences` has no `.catch`, so a DataStore read error in that collector was another uncaught throw on `viewModelScope`.

## Tests

`ComponentControlViewModelTest` is new — 8 tests on the consent gate, which is the follow-up #435 recorded as out of scope *because* the gate depended on DataStore. Making consent a synchronous injectable is what made it testable.

Each test asserts both halves, because either alone passes while the gate is broken: `pendingConsent` says the disclaimer was raised, and the calls reaching `FakeSystemRepository` say whether the disable actually went through. A test that only looked for a null `pendingConsent` would be equally happy with a `requestDisable` that quietly did nothing.

Covered: the disclaimer is raised in a fresh session; confirming runs the parked disable; confirming *without* ticking asks again; ticking silences the rest of the session; the silence spans packages (a second view model over a second package); the silence does **not** survive a new session; dismissing clears the request without disabling; confirming with nothing parked is a no-op.

`ComponentControlUseCaseTest` gained 4 tests for the ledger-vs-platform split, driven by a `FakeLedger` that can be told to throw: a ledger failure does not turn a successful disable or enable into a failure; `forget` reports a Room failure rather than throwing; Restore All under-reports a row whose delete failed.

| | |
|---|---|
| `:app:testStoreDebugUnitTest` | **1739 tests, 0 failures, 0 errors** (1727 → +4 use case, +8 view model) |
| `:app:lintStoreRelease` | **8 issues, all Hint** — 3 `AndroidGradlePluginVersion`, 5 `VectorPath`. Unchanged from baseline. |

Counts parsed from `build/test-results/**/*.xml`, not the Gradle log line.

Strings added in all 8 locales. `component_disabled_unrecorded` is phrased impersonally in every one (`Se desactivó`, `Wyłączono`, `تم تعطيل`) — a first draft put gender agreement on the `%1$s` placeholder, which is the same locale defect fixed in the previous review round.

## Not verified on device

The three fixes are reached only by a Room failure (a full or damaged database), which is not a state I can produce on the owner's device without deliberately filling it. The consent gate is unit-tested and the disable path itself was verified on a rooted physical device in #435.

## Summary by CodeRabbit

* **New Features**
  * Added a session-only “Don’t ask again” option for component-disable warnings.
  * Added warnings when disabled components cannot be included in “Restore all.”
  * Added translations for the new messages across supported languages.

* **Bug Fixes**
  * Improved component action feedback by distinguishing system-action success from recording failures.
  * Prevented recording failures from interrupting other restore operations.
  * Ensured restore counts include only successfully restored and recorded components.
  * Improved accessibility semantics for the session-only option.

### DIFF
--- a/app/src/androidTest/java/com/valhalla/thor/presentation/appList/DontAskAgainRowTest.kt
+++ b/app/src/androidTest/java/com/valhalla/thor/presentation/appList/DontAskAgainRowTest.kt
@@ -1,0 +1,137 @@
+// SPDX-FileCopyrightText: 2025-2026 Trinadh Thatakula <github.com/trinadhthatakula/Thor>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package com.valhalla.thor.presentation.appList
+
+import androidx.activity.ComponentActivity
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.SemanticsProperties
+import androidx.compose.ui.test.SemanticsMatcher
+import androidx.compose.ui.test.assert
+import androidx.compose.ui.test.assertIsOff
+import androidx.compose.ui.test.assertIsOn
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.isToggleable
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.percentOffset
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTouchInput
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.valhalla.thor.R
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * What the component disclaimer's "don't ask again this session" option announces.
+ *
+ * The same defect [SettingsSwitchRow][com.valhalla.thor.presentation.settings.SettingsSwitchRow] was
+ * fixed for, on a row written afterwards: `Modifier.clickable` around a `Checkbox` whose own handler
+ * is null contributes an on-click action and no *state*, so between the two nodes nothing carries
+ * ticked-or-not and a screen reader offers "double tap to activate" over an option whose value it
+ * never says.
+ *
+ * Worth asserting on this row in particular rather than trusting the pattern. It is the one control
+ * in Thor that turns *off* a warning — the disclaimer standing between a mis-tap and a component
+ * disabled inside an app that goes on looking perfectly healthy. A user who cannot hear the box's
+ * state can silence that warning for the session without knowing they did.
+ */
+@RunWith(AndroidJUnit4::class)
+class DontAskAgainRowTest {
+
+    @get:Rule
+    val rule = createAndroidComposeRule<ComponentActivity>()
+
+    private val reported = mutableListOf<Boolean>()
+
+    /** Read off the activity rather than hardcoded, so the eight locales cannot break the lookup. */
+    private val label: String
+        get() = rule.activity.getString(R.string.component_disclaimer_dont_ask_session)
+
+    private fun setRow(checked: Boolean) = rule.setContent {
+        DontAskAgainRow(checked = checked, onCheckedChange = { reported += it })
+    }
+
+    private val hasCheckboxRole = SemanticsMatcher.expectValue(
+        SemanticsProperties.Role,
+        Role.Checkbox
+    )
+
+    /** The two facts `clickable` plus a silenced `Checkbox` lost between them: which control, and its state. */
+    @Test
+    fun theRowAnnouncesItselfAsACheckboxAndSaysItIsTicked() {
+        setRow(checked = true)
+
+        rule.onNodeWithText(label).assert(hasCheckboxRole).assertIsOn()
+    }
+
+    /** And unticked, so that "on" above is a reading rather than a constant. */
+    @Test
+    fun anUntickedRowSaysOff() {
+        setRow(checked = false)
+
+        rule.onNodeWithText(label).assert(hasCheckboxRole).assertIsOff()
+    }
+
+    /**
+     * The option is offered once.
+     *
+     * A live `onCheckedChange` on the `Checkbox` would carry its own toggle semantics, making two
+     * toggleable nodes for one option — a screen reader walks past it twice and reads the label on
+     * only one of them. The null handler is what prevents that, and this is the assertion that it
+     * holds.
+     */
+    @Test
+    fun thereIsExactlyOneControlToFind() {
+        setRow(checked = false)
+
+        assertEquals(1, rule.onAllNodes(isToggleable()).fetchSemanticsNodes().size)
+    }
+
+    /** Tapping an unticked row asks to silence the disclaimer. */
+    @Test
+    fun tappingAnUntickedRowAsksForTrue() {
+        setRow(checked = false)
+
+        rule.onNodeWithText(label).performClick()
+
+        assertEquals(listOf(true), reported)
+    }
+
+    /**
+     * And a ticked row asks for **false** — the value the user wants, not the one in force.
+     *
+     * The direction matters because the caller assigns the reported value straight to its state. An
+     * implementation that reported the current value instead would make the box impossible to untick,
+     * which on this row means a consent the user tried to withdraw and could not.
+     */
+    @Test
+    fun tappingATickedRowAsksForFalse() {
+        setRow(checked = true)
+
+        rule.onNodeWithText(label).performClick()
+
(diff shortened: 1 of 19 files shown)


trinadhthatakula:
## CodeRabbit's finding: applied, and it was the third instance

One actionable finding this round — the `Don't ask again this session` row exposed a generic click and **no checked state**. Valid, and read before merging this time.

`Modifier.clickable` contributes an on-click action with no state, and the `Checkbox`'s own handler has to be null or the two nodes disagree on a fast double-tap. Between them nothing carried ticked-or-not. Not "two nodes where there should be one" — **zero** toggleable nodes, so a screen reader announced "double tap to activate" over an option whose value it never said.

Fixed with `toggleable(value, onValueChange, role = Role.Checkbox)`, which is already the idiom in three other places in this repo — `SettingsSwitchRow`, `AppBackupSheet`'s row, and `FixStoreSheet`'s row, whose comment states this exact reasoning. `SettingsSwitchRowTest` exists *because* a settings row announced itself as a button. This row was written with `clickable` anyway, after all of that. Same pattern as finding 1 in the PR body: the fix was in the repo, and the class was not swept.

It matters more on this row than on a selection row. This box turns *off* the disclaimer — a user who cannot hear its state can silence the one warning standing between a mis-tap and a component disabled inside an app that goes on looking perfectly healthy.

### Tested, and the tests were checked for bite

The row is extracted to `DontAskAgainRow` so its semantics are assertable without a bound view model. `DontAskAgainRowTest` mirrors `SettingsSwitchRowTest`: the role and both states, exactly one toggleable node, both toggle directions, and a tap landing on the box itself counting once rather than twice.

Then I put the `clickable` markup back and re-ran, because a green test on a fixed bug proves nothing about whether it would have caught it:

```
FAIL theRowAnnouncesItselfAsACheckboxAndSaysItIsTicked
     AssertionError: Failed to assert the following: (Role = 'Checkbox')
FAIL thereIsExactlyOneControlToFind
     AssertionError: expected:<1> but was:<0>
FAIL anUntickedRowSaysOff
     AssertionError: Failed to assert the following: (Role = 'Checkbox')
```

3 of 6 fail on the old markup, and `expected:<1> but was:<0>` *is* the finding restated — there was no control there to announce. The 3 that still pass are the behavioural ones, which is correct: `clickable` toggled the value fine. Only the announcement was missing, which is exactly why nobody noticed.

| | |
|---|---|
| `connectedStoreDebugAndroidTest` (Stock_A16_API36) | **6 tests, 0 failures** |
| `:app:testStoreDebugUnitTest` | **1739 tests, 0 failures, 0 errors** |
| `:app:lintStoreRelease` | **8 issues, all Hint** — unchanged |

### Not carried into this PR

A sweep for siblings of the three original findings confirmed the same uncaught-`viewModelScope`-throw defect at **9 more sites** across `FreezerViewModel`, `AppListViewModel`, `AppInfoDetailsViewModel` and `MainViewModel` — all raw `FreezerDao` calls in bare launches, several of them bookkeeping that runs *after* an irreversible privileged act has already succeeded. `FreezerViewModel` has the guard in two of its own functions, with a comment naming the hazard, and not in the other two.

Left out deliberately: four unrelated ViewModels is outside a review-response PR on the Components tab, and that sweep deserves its own PR and its own tests. Flagged for the owner.

trinadhthatakula:
## Correction: my own a11y fix introduced a 24dp touch target

Following up on the previous comment — the `toggleable` fix was right about the announcement and wrong about the target size, and I only caught it because a memory note pointed at the exact trade.

`Checkbox` applies `minimumInteractiveComponentSize()` **only while `onCheckedChange != null`**. The null handler that fixes the announcement therefore surrenders the 48dp the box was reserving, and `toggleable` enforces no minimum of its own. So the fix for WCAG 4.1.2 bought a 2.5.8 failure.

I measured rather than assumed, and the number is not marginal:

```
FAIL theTouchTargetIsAtLeast48dp
     AssertionError: Actual height is 24.0.dp, expected at least 48.0.dp (tolerance: 0.5.dp)
```

**Exactly half.** My own inline comment had justified the full-width row with "a 20dp target inside a dialog is below the 48dp minimum" — and then the row was 24dp tall, because widening it only ever fixed the horizontal axis. That comment is corrected in the code.

Fixed with `heightIn(min = 48.dp)`. `AppBackupSheet`'s row already carries it, with a comment stating this exact reasoning — **the fourth time in this change set that the remedy was already in the repo and the class had not been swept.** The only reason this one got caught is that something measured it.

`connectedStoreDebugAndroidTest`: **7 tests, 0 failures** (Stock_A16_API36). 1739 unit tests, 0 failures. Lint unchanged at 8 hints.

### Checked the siblings, one is unresolved

- `AppBackupSheet` — has `heightIn(min = 48.dp)`. Fine.
- `SettingsSwitchRow` — `.padding(16.dp)` around an icon box. Comfortably over.
- `FixStoreSheet`'s selection row — **not measured.** 36dp `AppIcon` + 4dp vertical padding each side is 44dp before its two-line text block, so it depends on whether the text is taller than the icon. It may well be fine. Flagging it rather than fixing it: different sheet, different feature, and I would rather report an unmeasured suspicion honestly than widen this PR on a guess.

Ready for review. Nothing outstanding from CodeRabbit — its one actionable finding was the semantics issue, addressed two commits ago.

